PR #53, opened 2020-05-25
20.07.00 transpikaspen

NashvillePublicLibrary:
Solves some fraction of https://github.com/mdnoble73/aspen-discovery/issues/52

And has a ton of merges...


mdnoble73:
This was against an old version, we merged the updated version against 20.08

PR #59, opened 2020-05-28
Patch 20

jzairo:


mdnoble73:
@jzairo this is against master, is that correct?

mdnoble73:
Talked with JZ and these commits are now part of pull requests on 20.08, closing this one

PR #63, opened 2020-06-17
20.09.00 carl xsip2 holds

NashvillePublicLibrary:
N.B. the code that replaces CarlX PatronAPI holds queue position calculation with SIP2 patron information call/response for 'unavail' could be sidestepped with CarlX SIP ini flag USEHOLDOCCUR first available in CarlX 9.6.5.1. See https://ww2.tlcdelivers.com/helpdesk/Default.asp?TicketID=500458


NashvillePublicLibrary:
https://trello.com/c/ILFhsn7L

PR #65, opened 2020-06-30
20.09.00

librarianbryan:
Fleshes out and cross references Author Pages and Wikipedia sections of help manual.


mdnoble73:
Merging into 20.10.00 Thanks!

PR #66, opened 2020-07-04
Updates nashville.aspenlocal config and .idea files

NashvillePublicLibrary:
Updates files related to Nashville development environment

https://trello.com/c/UT5wqd23

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
+# Editor-based HTTP Client requests
+/httpRequests/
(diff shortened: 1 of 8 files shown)


mdnoble73:
I'm torn about committing these.  On one hand it makes it easier to setup new environments on the other hand they are local files that shouldn't be added to git.  What are your thoughts pro/con?

NashvillePublicLibrary:
My off-the-cuff thought:

Posting the jetbrains config files *potentially* helps me, you, and Bryan set up faster  - and *potentially* causes hiccups, which in turn , *potentially* cause us to learn more about what the config means and how to maximize our phpstorm use.

mdnoble73:
I currently have the web project in code/web that should have debuggers etc setup already.  Do you want to try loading that rather than creating a new project?

mdnoble73:
James will move config things to config.pwd and used .gitignore to remove the .idea files

PR #69, opened 2020-07-04
20.10.00 list api

NashvillePublicLibrary:


mdnoble73:
A few things to revise/decide on before committing:
1. Can you remove the things in .idea and config.ini from this commit?  We should decide separately how much of that should be in.  I added mine to make it easier to setup a new environment, but that may be bad. 
2. Did you intend to remove the IP restrictions from AnodeAPI? If possible it would be good to leave those (although there isn't any patron data being transferred). 
3. I need to test the change to getListRecords to see what issues it will cause in the user interface.  It looks like the Interface is passing a 1 based start record rather than 0 based. 
4. We should decide what to do rather than hardcoding branch names. Maybe a separate configuration for anode?

NashvillePublicLibrary:
You're correct on all counts!

1. jetbrains files are actually a part of https://github.com/mdnoble73/aspen-discovery/pull/66 and I thought I needed 'em in place for me to proceed. Alternatively, I could ignore them :)

2. I did mean to remove the IP restrictions from AnodeAPI. At this point I believe the AnodeAPI-consuming devices don't have stable IP addresses. After reviewing the data and finding no sensitive information revealed, I chose "open to the world". See card at https://trello.com/c/Gwj0o6y3

3. Hrmmm... I can quickly imagine a klugey fix... if x = 0, y = 0; elseif x > 0, y = x-1...

4. AnodeAPI has always hardcoded the branch names. I figure Anode ain't growing library customers right now and we can leave that for another day. 'Course, if you wanna keep tidy for tidy's sake, I'm happy to oblige!

mdnoble73:
Can you resubmit this against 20.11?

PR #78, opened 2020-07-16
20.11.00 rosen register

NashvillePublicLibrary:
@mdnoble73 let's play everyone's favorite game show "How Many Antipatterns Can You Find?"

Rather, let's have code review!

### DIFF
--- a/code/web/services/MyAccount/RegisterRosenLevelUP.php
+++ b/code/web/services/MyAccount/RegisterRosenLevelUP.php
@@ -97,77 +97,78 @@ class MyAccount_RegisterRosenLevelUP extends MyAccount
 					require_once ROOT_DIR . '/sys/Enrichment/RecaptchaSetting.php';
 					$recaptchaValid = RecaptchaSetting::validateRecaptcha();
 					if (!$recaptchaValid) {
-						$interface->assign('captchaMessage', translate('The CAPTCHA response was incorrect, please try again.');
+						$interface->assign('captchaMessage', translate('The CAPTCHA response was incorrect, please try again.'));
 					} else {
-					//Submit the form to Rosen
+						//Submit the form to Rosen
 
-					// check parent username for availability; if it ain't available, check for identical email; if identical email, allow parent to register additional students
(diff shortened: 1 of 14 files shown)


NashvillePublicLibrary:
Items to discuss:

1. have I made $this too much? In order to reduce redundant code, I created functions; to get the functions to alter the same object, I used $this; I decided to throw almost everything in $this, which for me makes the code less legible

2. It works...

3. There's a lot of "Hardcoded for Nashville" in the code. What're better ways to manage these needs?

4. The code seems awfully complicated for what is ultimately a pretty simple workflow. How can I elegantify the logic?

NashvillePublicLibrary:
1. I have made all the parts translatable that I could think to make translatable!
2. I have removed most of the Hardcoded for Nashville blocks (but I just found one I forgot!)
3. I have created a separate class in /web/sys
4. that's the Hardcoded for Nashville I forgot... (ptype eligibility)
5. moved config array to database

code/web/services/MyAccount/RegisterRosenLevelUP.php
```diff
@@ -0,0 +1,378 @@
...
+					case 23:
+						$this->student_grade_level = '1';
+						break;
+					case 24:
+						$this->student_grade_level = '2';
+						break;
+					case 25:
+						$this->student_grade_level = '3'; // 3rd grade will be coded as 2nd grade for LevelUP
+						break;
+				}
```
NashvillePublicLibrary, line 85: @mdnoble73 could we make the expected GUI admin input for ptype be like "21=>'K',22=>'K',23=>'1',24=>'2',25=>'2'" or is that way too much for the library customer admin? (not to mention it doesn't give Nashville the right labels in the dropdown...). A column in ptype seems overkill, but I'll do it with permission...

mdnoble73: Since this is really just for K-2, how about adding regular expressions for the patron types to use for:
K
1
2
3 

Then eventually we could say if those have not been filled out, we prompt the user to enter their grade level.

PR #98, opened 2020-10-09
20.14.00 nashville school reports revisited

NashvillePublicLibrary:


mdnoble73 (approved):
Looks good. Long term we may want to move some of the inline css into the main .less files.  Especially the doNotPrint class.

PR #135, opened 2020-11-15
Create Cover-Images.MD

emily-bywater:
Created page and added video


mdnoble73 (changes requested):
Libraries do not have to use the coce server or OMDB if they do not want to. Content Cafe can also be used as a source of covers.

PR #178, opened 2020-11-18
20.15.00 carl x self reg

NashvillePublicLibrary:
@mdnoble73 we will need to do a code review - of selfReg.tpl. I've changed it! Other folks might not like what I've done!


mdnoble73:
In CARL.X you don't need to translate the labels for each field.  Those are automatically translated when the form is displayed. 

Also, one of the templates is in the nashville subtheme. I'm not sure that is used or needed.

NashvillePublicLibrary:
@mdnoble73 I've made the changes you suggested.

What I want to check in with folks (you) about: I changed the logic of the selfreg tpl file so that non-success + message = display message, do not re-display the form. That's the desired behavior in Nashville - but I don't know about other libraries.

mdnoble73:
Can you resubmit against either 20.15.01 or 20.16.00?

NashvillePublicLibrary:
@mdnoble73 as promised - another update to CarlX SelfReg

mdnoble73:
Changes are all looking good.  Can you resubmit the pull request against 20.16 rather than master?

NashvillePublicLibrary:
D'oh! Resubmitting against 20.16.00

mdnoble73 (approved):

PR #186, opened 2020-12-03
20.17.00

emily-bywater:


mdnoble73 (changes requested):
It looks like the Permissions and Browse Category pages are no longer linked.

The rest of the reorganization is looking good.

mdnoble73 (approved):

PR #190, opened 2020-12-07
Updates to Administration Section

emily-bywater:


mdnoble73 (commented):
On the Aspen-Discovery-Help.MD rather than linking the headings to YouTube and then also adding them to the tutorial, let's link to the actual help manual and release notes within their site.

mdnoble73 (commented):
On ILS-Integration.MD there is a link to Bib Level format in the Video Tutorial section, but that isn't a video tutorial

mdnoble73 (commented):
For the help manual, we need to make sure to expand on the descriptive information shown within the Administration Home Page menu.  It looks like in a lot of places there is just a copy of the description from the menu.  Getting more information embedded should be the next pass through.

mdnoble73 (commented):
For the help manual, we need to make sure to expand on the descriptive information shown within the Administration Home Page menu.  It looks like in a lot of places there is just a copy of the description from the menu.  Getting more information embedded should be the next pass through.

PR #192, opened 2020-12-18
Fix reported memory size

RudyHost:
This change should fix the memory size the API reports.  StringUtils::formatBytes expects bytes, but /proc/meminfo provides kilobytes.


mdnoble73 (approved):

mdnoble73:
Looks great Rudy, thanks!

PR #255, opened 2021-04-21
Various fixes

catsoup11789:


code/web/interface/themes/responsive/DataObjectUtil/objectEditForm.tpl
```diff
@@ -72,6 +72,9 @@
 			{if !empty($initializationJs)}
 				{$initializationJs}
 			{/if}
+			{if !empty($initializationAdditionalJs)}
```
mdnoble73, line 75: Could we do this with just the regular initializationJs?  By making the initialization a function or by calling multiple methods separated with semi-colons?  Let's talk through tomorrow.

PR #263, opened 2021-05-03
Various updates for lists and SearchAPI

catsoup11789:


code/web/interface/themes/responsive/MyAccount/lists.tpl
```diff
@@ -53,6 +53,13 @@
 							</div>
 						</div>
 					{/if}
+					<div class="row">
+
+						<div class="col-xs-12">
+							<p class="text-muted"><small>Created: {$list->created|date_format:"%B %e, %Y %l:%M %p"}<br>
+							Last Updated: {$list->dateUpdated|date_format:"%B %e, %Y %l:%M %p"}</small></p>
```
mdnoble73, line 60: Let's make these translatable.

code/web/interface/themes/responsive/MyAccount/list.tpl
```diff
@@ -16,8 +17,20 @@
 					{if $userList->deleted == 1}
 						<p class="alert alert-danger">{translate text='Sorry, this list has been deleted.'}</p>
 					{else}
+						<div class="row">
+							<div class="col-md-4">
+								<p class="text-muted">
+									<small>Created: {$dateCreated}</small>
```
mdnoble73, line 23: Let's make these translatable. I'm also trying to veer away from colons at the end of text if we put the text and value in different columns.  Not sure if that's the right thing to do here or not.

code/web/services/API/SearchAPI.php
```diff
@@ -245,6 +245,28 @@ function getIndexStatus()
...
+					require_once ROOT_DIR . '/sys/WebBuilder/PortalPage.php';
+					require_once ROOT_DIR . '/sys/WebBuilder/BasicPage.php';
+					require_once ROOT_DIR . '/sys/WebBuilder/WebResource.php';
+					$portalPage = new PortalPage();
+					$basicPage = new BasicPage();
+					$webResource = new WebResource();
+					$portalPage->find();
+					$basicPage->find();
+					$webResource->find();
+					if ($portalPage->fetch()) {
```
mdnoble73, line 259: This could also use ->getNumResults() rather than ->fetch() which could be slightly faster.

code/web/services/Browse/AJAX.php
```diff
@@ -110,9 +110,17 @@ function createBrowseCategory(){
 					);
 				}
 			}else{
+				require_once ROOT_DIR . '/sys/UserLists/UserList.php';
 				$listId = $_REQUEST['listId'];
+				$userList = new UserList();
+				$userList->listId = $listId;
+				$userList->deleted = "0";
+				$userList->find();
```
mdnoble73, line 118: Should this be $userList->find(true) to make sure we load the list?

mdnoble73 (changes requested):

code/web/services/API/SearchAPI.php
```diff
@@ -96,7 +96,7 @@ function getIndexStatus()
 			}
 		}
 
-		//Check free disk space
+/*		//Check free disk space
```
mdnoble73, line 99: This block should not be commented out

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #281, opened 2021-05-27
Enhancements to updating collection spotlights from search tools

catsoup11789:


code/web/services/Admin/AJAX.php
```diff
@@ -166,6 +166,14 @@ function getAddToSpotlightForm()
 		}
 		$collectionSpotlight->orderBy('name');
 		$existingCollectionSpotlights = $collectionSpotlight->fetchAll('id', 'name');
+
+		$spotlightList = new CollectionSpotlightList();
```
mdnoble73, line 170: Do we need to do any filtering of this depending on what spotlights the user has access to?

mdnoble73 (approved):

PR #297, opened 2021-06-04
Does not delete OverDrive titles following OverDrive server errors

NashvillePublicLibrary:
https://ticket.bywatersolutions.com/SelfService/Update.html?id=82382

I have imitated the code that sidesteps error consequences from the CarlX export and submitted a pull request, but I'm not sure it would actually stop flagging the OverDrive product as deleted - and instead it might just stop the error from being tallied in the OverDrive indexing log


mdnoble73:
This doesn't really do anything yet.  Let's not merge this in for now.

PR #301, opened 2021-06-21
Various things, mostly hold updates

catsoup11789:


code/web/services/MyAccount/AJAX.php
```diff
@@ -1990,6 +2371,7 @@ public function getHolds()
 					}
 				}
 				$interface->assign('notification_method', strtolower($notification_method));
+				$interface->assign('profile', $user->id);
```
mdnoble73, line 2374: I think typically profile is the entire user object, maybe use userId instead?

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #307, opened 2021-06-28
Scope the staff directory for the current instance (82639)

catsoup11789:


code/web/services/Record/AJAX.php
```diff
@@ -275,7 +275,7 @@ function getPlaceHoldVolumesForm()
 			}
 
 			$interface->assign('hasItemsWithoutVolumes', $numItemsWithoutVolumes > 0);
-			$interface->assign('majorityOfItemsHaveVolumes', $numItemsWithVolumes > $numItemsWithoutVolumes);
+			$interface->assign('majorityOfItemsHaveVolumes', $numItemsWithVolumes >= 1);
```
mdnoble73, line 278: I don't think this needs to change, which ticket was this related to?

mdnoble73 (changes requested):

code/web/services/Record/AJAX.php
catsoup11789: 80458

PR #311, opened 2021-07-01
Check if pickup location is set when bypass is on

catsoup11789:


code/web/sys/Account/User.php
```diff
@@ -2205,6 +2205,25 @@ function getHomeLocationCode(){
 		return $this->getHomeLocation()->code;
 	}
 
+	function getPickupLocationCode(){
+		if($this->pickupLocationId != $this->homeLocationId){
+			$pickupBranch = $this->pickupLocationId;
+			$locationLookup = new Location();
+			$locationLookup->locationId = $pickupBranch;
+			$locationLookup->find();
```
mdnoble73, line 2213: This can get simplified with:
$lookupLocation->find(true) 
which will return fetch the first result and return false if no matches are found.

code/web/sys/Account/User.php
```diff
@@ -2205,6 +2205,25 @@ function getHomeLocationCode(){
 		return $this->getHomeLocation()->code;
 	}
 
+	function getPickupLocationCode(){
+		if($this->pickupLocationId != $this->homeLocationId){
```
mdnoble73, line 2209: Do we need to check here to see if someone has remember pickup location turned on or is that checked in other places?  Just want to make sure things work if people set remember pickup location on and then turn it off.

mdnoble73 (changes requested):

code/web/sys/Account/User.php
catsoup11789: It was checking it in the hold-popup template, but I also added a check in the function so when it's called elsewhere

mdnoble73 (approved):

PR #317, opened 2021-07-13
Various things...

catsoup11789:


code/web/sys/LibraryLocation/Library.php
```diff
@@ -955,7 +955,7 @@ static function getObjectStructure() : array {
 			'holidays' => array(
 				'property' => 'holidays',
 				'type' => 'oneToMany',
-				'label' => 'Holidays',
+				'label' => '<h2>Holidays</h2>',
```
mdnoble73, line 958: Is there any way that we can do this without embedding HTML in the label?  I think that will break some of the translations.  

We should be able to do something in property.tpl of the property type is section.  It looks like Aspen is already trying to do that on line 44.  So we may just need to add a renderAsHeading attribute to these properties.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #319, opened 2021-07-14
Updates

catsoup11789:


code/web/services/WebBuilder/DownloadPDF.php
```diff
@@ -18,7 +18,7 @@ function launch()
 		}
 
 		global $serverName;
-		$dataPath = '/data/aspen-discovery/' . $serverName . '/uploads/web_builder_pdf/';
+		$dataPath = '/usr/local/aspen-discovery-data/' . $serverName . '/uploads/web_builder_pdf/';
```
mdnoble73, line 21: I think this is just your install.  We should make the data dir path configurable.  It could probably go into System Variables or we could put it in the config file since it doesn't change much.

code/web/services/WebBuilder/PDFs.php
```diff
@@ -55,7 +55,7 @@ function getObjectStructure() : array
 		unset($objectStructure['type']);
 		$fileProperty = $objectStructure['fullPath'];
 		global $serverName;
-		$dataPath = '/data/aspen-discovery/' . $serverName . '/uploads/web_builder_pdf/';
+		$dataPath = '/usr/local/aspen-discovery-data/' . $serverName . '/uploads/web_builder_pdf/';
```
mdnoble73, line 58: Same as above, we can configure this or update to the default.

code/web/sys/File/FileUpload.php
```diff
@@ -15,10 +16,32 @@ static function getObjectStructure() : array
...
+			'thumbFullPath' => array('property'=>'thumbFullPath', 'type' => 'text', 'label'=>'Thumbnail Full Path', 'description'=>'The path of the generated thumbnail on the server', 'serverValidation' => 'makeThumbnail', 'readOnly' => true),
 		];
 	}
 	
 	public function getFileName(){
 		return basename($this->fullPath);
 	}
+
+	/** @noinspection PhpUnused */
+	function makeThumbnail()
```
mdnoble73, line 28: Do we need to check that the file is a PDF?  We allow people to upload other types of files, i.e. Excel files as supplemental materials.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #322, opened 2021-07-20
Various updates

catsoup11789:


code/web/sys/File/FileUpload.php
```diff
@@ -27,21 +27,25 @@ public function getFileName(){
...
+			}
 
-		$this->thumbFullPath = $thumbFullPath;
-		$this->update();
-		return $thumbFullPath;
+			$this->thumbFullPath = $thumbFullPath;
+			$this->update();
+			return $thumbFullPath;
+		} else {
+			die();
```
mdnoble73, line 48: Do we want to die here or just return?

mdnoble73 (approved):

PR #330, opened 2021-07-23
Updated Facet Lock icon to use open-panel foreground color for better visibility

catsoup11789:


code/web/sys/DBMaintenance/version_updates/21.09.00.php
```diff
@@ -311,22 +311,28 @@ function getUpdates21_09_00() : array
...
-			'sql' => [
-				'ALTER TABLE library DROP COLUMN defaultPType',
-				'ALTER TABLE library DROP COLUMN pTypes',
-				'ALTER TABLE location DROP COLUMN defaultPType',
-			]
-		]
-	];
-}
\ No newline at end of file
+		], //rebuildThemes21_09
```
mdnoble73, line 321: This needs to get updated so my change is still in place.  I think the rebuild of themes is already accessible as well?

mdnoble73 (changes requested):

PR #340, opened 2021-08-06
So many commits

catsoup11789:


code/web/services/WebBuilder/BasicPage.php
```diff
@@ -25,6 +37,59 @@ function launch()
...
+	{
+		require_once ROOT_DIR . '/sys/WebBuilder/BasicPageAccess.php';
+		require_once ROOT_DIR . '/sys/Account/PType.php';
+		require_once ROOT_DIR . '/sys/WebBuilder/BasicPage.php';
+
+		$requireLogin = 0;
+		$id = strip_tags($_REQUEST['id']);
+		$page = new BasicPage();
+		$page->id = $id;
+		$page->find();
```
mdnoble73, line 50: This can be simplified to 
if ($page->find(true)) {

mdnoble73 (changes requested):
Changes look great, just a couple of minor things.  Could you also add the p type restrictions to menu links for a library?  That way a library could actually create an intranet for their staff.

mdnoble73 (approved):

PR #391, opened 2021-09-28
Fix to populating IDs on holds/checkouts

catsoup11789:


code/web/app/aspenListHolds.php
```diff
@@ -72,8 +72,15 @@
 # * clean up the title and convert the due date from a timestamp
 # ****************************************************************************************************************************
     if(substr($item['title'], -1) == '/') { $item['title'] = substr($item['title'], 0, -1); }
+
```
mdnoble73, line 75: Looks like this one still has it flipped

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #411, opened 2021-10-29
Kware RTL and Arabic translation contribution PR

kwareict:
RTL Languages support & Changes to make some strings translatable


mdnoble73:
Updating for compatibility with 21.15

PR #415, opened 2021-11-01
Updates and fixes from tickets

catsoup11789:


code/web/interface/themes/responsive/Record/invalidRecord.tpl
```diff
@@ -12,15 +12,15 @@
 </p>
 
 {if $materialRequestType == 1 }
-	<p class="alert alert-info">
+	<p class="alert alert-info materialsRequestLink">
```
mdnoble73, line 15: Does this need the check for whether or not materials request should be shown?

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #425, opened 2021-11-18
Kware RTL and translation contribution PR V1

kwareict:
Kware RTL and translation contribution PR V1


mdnoble73:
Merged this manually with some changes.

PR #478, opened 2022-01-20
Updates, fixes...

catsoup11789:


code/web/services/API/UserAPI.php
```diff
@@ -1201,7 +1201,17 @@ function placeHold() : array
 				if ($source == 'ils' || $source == null) {
 					if (isset($_REQUEST['pickupBranch']) || isset($_REQUEST['campus'])) {
 						if (isset($_REQUEST['pickupBranch'])) {
-							$pickupBranch = trim($_REQUEST['pickupBranch']);
+							if($_REQUEST['pickupBranch'] == "null") {
+								$location = new Location();
+								$userPickupLocations = $location->getPickupBranches($patron);
+								foreach ($userPickupLocations as $tmpLocation) {
+									if ($tmpLocation->code == $patron->getPickupLocationCode()) {
+										$pickupBranch = $tmpLocation->code;
```
mdnoble73, line 1209: We can add a break here so we don't have to loop through the rest of the list if there are multiple branches.

mdnoble73 (approved):
One little comment, but not critical for now.

PR #485, opened 2022-01-24
library branded app config/setup

catsoup11789:


code/web/sys/AspenLiDASetting.php
```diff
@@ -1,19 +1,111 @@
...
+		unset($quickSearches['aspenLidaSettingId']);
+
 		$structure = array(
 			'id' => array('property' => 'id', 'type' => 'label', 'label' => 'Id', 'description' => 'The unique id'),
-			'slugName' => array('property' => 'slugName', 'type' => 'text', 'label' => 'Slug Name', 'description' => 'A name for the app without spaces, i.e. "aspen-lida"', 'maxLength' => 50),
+			'slugName' => array('property' => 'slugName', 'type' => 'text', 'label' => 'Slug Name', 'description' => 'The name for the app without spaces', 'maxLength' => 50, 'note' => 'Matches the slug in the app config', 'required' => true),
+			'logoLogin' => array('property' => 'logoLogin', 'type' => 'image', 'label' => 'Logo for Login Screen', 'description' => 'The logo used on the login screen of the app', 'note' => '1024x1024 or 512x512 is the recommended image size', 'hideInLists' => true, 'required' => false, 'thumbWidth' => 512),
+			'privacyPolicy' => array('property' => 'privacyPolicy', 'type' => 'text', 'label' => 'URL to Privacy Policy', 'description' => 'The web address for users to access the privacy policy for using the app', 'hideInLists' => true, 'required' => false),
+			'quickSearches' => array(
```
mdnoble73, line 23: I think the quick searches need to be attached to the library rather than the entire Aspen instance so each library can have their own quick searches within LiDA.

code/web/sys/AspenLiDASetting.php
```diff
@@ -1,19 +1,111 @@
 <?php
 
+require_once ROOT_DIR . '/sys/DB/DataObject.php';
```
mdnoble73, line 3: Do the settings need a connection to the Library so we can tell which library is having the branding done?  How would this look if there are multiple branded apps within a single instance?

code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
```diff
@@ -15,7 +15,7 @@
 		valid_elements : '*[*]',
 		extended_valid_elements : [
 			'*[*]',
-			'img[class=img-responsive|src|border=0|alt|title]'
+			'img[class=img-responsive|src|border=0|alt|title|width|height]'
```
mdnoble73, line 18: This is to allow images to be resized when added by TinyMCE right?

code/web/interface/themes/responsive/DataObjectUtil/property.tpl
```diff
@@ -415,6 +415,7 @@
 					</label>
 				</div>
 			{/if}
+            {if $property.note}<span id="{$propName}HelpBlock" class="help-block" style="margin-top:0"><small><i class="fas fa-info-circle"></i> {$property.note}</small></span>{/if}
```
mdnoble73, line 418: Where is the note being used?  What is the difference between the note and description?

code/web/sys/DBMaintenance/version_updates/22.02.00.php
```diff
@@ -11,8 +11,8 @@ function getUpdates22_02_00() : array
 			]
 		], //sample*/
 		'aspen_lida_settings' => [
-			'title' => 'Add settings for Aspen LiDA',
-			'description' => 'Add settings for Aspen LiDA',
+			'title' => 'Add settings for white-label Aspen LiDA',
```
mdnoble73, line 14: Let's change white-label to library branded

code/web/interface/themes/responsive/DataObjectUtil/property.tpl
catsoup11789: The note is the helper that appears underneath the field and doesn't require a hover to see it. It was already put on most of the other fields, but I realized it wasn't on the file/image one, yet. Typically it houses info like formatting, sizing, etc.

code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
catsoup11789: yep

code/web/sys/AspenLiDASetting.php
catsoup11789: the library id that establishes the branding is hardcoded into the app, the settings are then tied to the app based on the app slug name (both variables are stored in app.config.js). So another branded app would require a new app.config.js, which likely has a new library id for branding but would have new app settings in the Aspen instance.

code/web/interface/themes/responsive/DataObjectUtil/property.tpl
mdnoble73: Thanks!

code/web/sys/AspenLiDASetting.php
mdnoble73: Will this setup allow a library who just uses Aspen LiDA to customize the Quick Searches?

catsoup11789: I honestly forgot about regular community app libraries and custom quick searches, unless we want to make it something that's only available to those with branded apps and community quick searches would be based on whatever we've got in the aspen-model app settings? i don't have this coded into the app yet, so it's still a good time to make adjustments.

catsoup11789: Quick searches have been moved to library settings

code/web/services/API/SystemAPI.php
```diff
@@ -426,6 +447,81 @@ public function displayAdminAlert() : bool {
...
+						flush();
+					}
+
+					fclose($handle);
+				} else {
+					readfile($fullPath);
+				}
+
+				die();
+			}
```
mdnoble73, line 521: Do we need another return here if there is no image or is just returning blank enough?

code/web/services/API/SystemAPI.php
```diff
@@ -426,6 +447,81 @@ public function displayAdminAlert() : bool {
...
+		if (isset($_REQUEST['themeId']) && isset($_REQUEST['type'])) {
+			global $configArray;
+			require_once ROOT_DIR . '/sys/Theming/Theme.php';
+			$id = strip_tags($_REQUEST['themeId']);
+			$type = strip_tags($_REQUEST['type']);
+
+			$theme = new Theme();
+			$theme->id = $id;
+			if (!$theme->find(true)) {
+				die();
```
mdnoble73, line 461: We should die before calling the theme->find if $id is blank.  If the id is blank, Aspen will actually return everything in the database.

code/web/services/API/SystemAPI.php
```diff
@@ -14,9 +14,13 @@ function launch()
...
 		header('Expires: Mon, 26 Jul 1997 05:00:00 GMT'); // Date in the past
 
+		if($method === "getLogoFile") {
+			return $this->$method();
+		};
+
 		if (isset($_SERVER['PHP_AUTH_USER'])) {
 			if($this->grantTokenAccess()) {
-				if (in_array($method, array('getLibraryInfo', 'getLocationInfo', 'getThemeInfo', 'getAppSettings'))) {
+				if (in_array($method, array('getLibraryInfo', 'getLocationInfo', 'getThemeInfo', 'getAppSettings', 'getLogoFile'))) {
```
mdnoble73, line 23: This change probably not needed since we check for getLogoFile above.

code/web/services/API/SystemAPI.php
```diff
@@ -119,13 +123,30 @@ public function getThemeInfo() : array
...
 			if ($app->find(true)){
+				$settings = [];
+				if($app->logoLogin) {
+					$settings['logo'] = $configArray['Site']['url'] . '/files/original/' . $app->logoLogin;
+				}
+				if($app->privacyPolicy) {
+					$settings['privacyPolicy'] = $app->privacyPolicy;
+				}
+
+				$quickSearches = new AspenLiDAQuickSearch();
```
mdnoble73, line 139: This needs to get updated to return the quick searches based on the selected library rather than the app (by slug).

code/web/sys/AspenLiDAQuickSearch.php
```diff
@@ -0,0 +1,21 @@
...
+	public $id;
+	public $weight;
+	public $libraryId;
+	public $searchTerm;
+	public $label;
+
+	static function getObjectStructure() : array {
+		$structure = array(
+			'id'           => array('property' => 'id', 'type'=> 'label', 'label'=> 'Id', 'description'=> 'The unique id'),
+			'label'        => array('property' => 'label', 'type'=> 'text', 'label'=> 'Label', 'description'=> 'The label for quick search', 'required' => true),
```
mdnoble73, line 16: Do we need properties for libraries or weight here?

code/web/services/API/SystemAPI.php
catsoup11789: Updated to return quick searches back in the getLibraries method from the library api instead

code/web/services/API/SystemAPI.php
catsoup11789: Blank would just load nothing, probably not ideal, but where do we go for an alternative file? The specific app logos will be required fields. I probably won't use this api to call the images from theme, but added them for potential future use.

code/web/sys/AspenLiDAQuickSearch.php
catsoup11789: seems to work ok without them. libraryId gets tied when saving, weight is updated when using the sort.

mdnoble73 (approved):

PR #497, opened 2022-02-08
Tiny MCE Config Changes

sloneccls:
As per email with Mark, expanded config options for TinyMCE to allow for more robust editing.

### DIFF
--- a/code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
+++ b/code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
@@ -8,29 +8,16 @@
 		toolbar1: 'code | cut copy paste pastetext spellchecker | undo redo searchreplace | image table hr codesample insertdatetime | link anchor ',
 		toolbar2: 'bold italic underline strikethrough | formatselect fontselect fontsizeselect forecolor backcolor',
 		toolbar3: 'numlist bullist toc | alignleft aligncenter alignright | preview visualblocks fullscreen help',
-		//menubar: '',
 		toolbar: 'image',
+		menubar:'',
     		image_advtab: true,
-		//
 		images_upload_url: '/WebBuilder/AJAX?method=uploadImageTinyMCE',
 		convert_urls: false,
 		theme: 'modern',
-		valid_elements : "@[id|class|style|title|dir<ltr?rtl|lang|xml::lang],"
-    + "a[rel|rev|charset|hreflang|tabindex|name|href|target|title|class],strong/b,em/i,strike,u,"
-    + "#p[style],-ol[type|compact],-ul[type|compact],-li,br,img[longdesc|usemap|"
-    + "src|border|alt=|title|hspace|vspace|width|height|align],-sub,-sup,"
-    + "-blockquote,-table[border=0|cellspacing|cellpadding|width|frame|rules|"
-    + "height|align|summary|bgcolor|background|bordercolor],-tr[rowspan|width|"
-    + "height|align|valign|bgcolor|background|bordercolor],tbody,thead,tfoot,"
-    + "#td[colspan|rowspan|width|height|align|valign|bgcolor|background|bordercolor"
-    + "|scope],#th[colspan|rowspan|width|height|align|valign|scope],caption,-div,"
-    + "-span,-code,-pre,address,-h1,-h2,-h3,-h4,-h5,-h6,hr[size|noshade],-font[face"
-    + "|size|color],embed[type|width|height|src|*],map[name],"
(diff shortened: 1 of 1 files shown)


code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
```diff
@@ -8,14 +8,29 @@
...
+    + "#td[colspan|rowspan|width|height|align|valign|bgcolor|background|bordercolor"
+    + "|scope],#th[colspan|rowspan|width|height|align|valign|scope],caption,-div,"
+    + "-span,-code,-pre,address,-h1,-h2,-h3,-h4,-h5,-h6,hr[size|noshade],-font[face"
+    + "|size|color],embed[type|width|height|src|*],map[name],"
+    + "button,col[align|char|charoff|span|valign|width],colgroup[align|char|charoff|span|"
+    + "valign|width],fieldset,label[for],legend,noscript,"
+    + "q[cite],small,textarea[cols|rows|disabled|name|readonly]",
 		extended_valid_elements : [
 			'*[*]',
-			'img[class=img-responsive|src|border=0|alt|title|width|height]'
```
mdnoble73, line 18: We believe this can be changed to: 
'img[class=img-responsive|*]'

code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
```diff
@@ -8,14 +8,29 @@
...
-		menubar: '',
+		//menubar: '',
+		toolbar: 'image',
+    		image_advtab: true,
+		//
 		images_upload_url: '/WebBuilder/AJAX?method=uploadImageTinyMCE',
 		convert_urls: false,
 		theme: 'modern',
-		valid_elements : '*[*]',
+		valid_elements : "@[id|class|style|title|dir<ltr?rtl|lang|xml::lang],"
```
mdnoble73, line 18: We think it's better to leave this as all valid elements (we restrict tags that can be used in system variables).

code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
```diff
@@ -8,14 +8,29 @@
 		toolbar1: 'code | cut copy paste pastetext spellchecker | undo redo searchreplace | image table hr codesample insertdatetime | link anchor ',
 		toolbar2: 'bold italic underline strikethrough | formatselect fontselect fontsizeselect forecolor backcolor',
 		toolbar3: 'numlist bullist toc | alignleft aligncenter alignright | preview visualblocks fullscreen help',
-		menubar: '',
+		//menubar: '',
```
mdnoble73, line 11: Do you need the menubar on?  We think it could be confusing for some of our less experienced users.

mdnoble73 (changes requested):
Just a couple updates.  Let us know what you think.

code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
sloneccls: Oops, didn't really mean to still have that on, but I'm double-checking with my co-worker to make sure that wasn't something she was really needing either.

code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
sloneccls: If I understood the above counter changes correctly, it seems like keeping the valid elements the same as in the base file, and then changing the extended elements here from `'img[class=img-responsive|src|border=0|alt|title|width|height]'` to `'img[class=img-responsive|*]'`  will probably do for is what we need (in a much cleaner way).

sloneccls (commented):
Let me know if I'm understanding the proposed changes correctly. If so, we're testing it out like this now.

code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
mdnoble73: That's correct

code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
mdnoble73: Correct, thanks

code/web/interface/themes/responsive/DataObjectUtil/textarea.tpl
mdnoble73: Thanks

sloneccls (commented):
Changes made

PR #500, opened 2022-02-10
Folio integration

sekjal:
Low-hanging fruit in the FOLIO driver:

1. Freeze/thaw holds not supported in FOLIO, so add those error messages to appropriate functions
2. placeHold just refers to placeItemHold with an empty item, since that's how FOLIO's API works
3. Adds the ability for a user to change their pickup Location (assuming that Pickup Locations are mapped to FOLIO's Service Points)


code/web/Drivers/FOLIO.php
```diff
@@ -396,16 +399,44 @@ public function placeVolumeHold(User $patron, $recordId, $volumeId, $pickupBranc
...
+			'message' => 'Thawing holds not implemented for this ILS');
 	}
 
 	function changeHoldPickupLocation($patron, $recordId, $itemToUpdateId, $newPickupLocation)
 	{
+		global $logger;
+		$hold = $this->makeRequest('GET', '/circulation/requests/' . $recordId);
+		$hold = json_decode($hold, true);
+
+		$hold['pikcupServicePointId'] = $newPickupLocation;
```
mdnoble73, line 420: Is there a spelling error here?  Should it be pickupServicePointId?

sekjal: Oh, yes, good catch, thank you.  it should be 'pickupServicePointId'

mdnoble73 (approved):

PR #506, opened 2022-02-15
api updates

catsoup11789:


code/aspen_app/App.js
```diff
@@ -5,13 +5,19 @@ import * as Sentry from 'sentry-expo';
...
-	sessionTrackingIntervalMillis: 10000,
-	debug: true, // If `true`, Sentry will try to print out useful debugging information if something goes wrong with sending the event. Set it to `false` in production
-});
+import { LogBox } from 'react-native';
+//LogBox.ignoreLogs(['Warning: ...']); // Ignore log notification by message
+//LogBox.ignoreAllLogs();//Ignore all log notifications
+
+if (!__DEV__) {
+	Sentry.init({
+		dsn: 'https://7fe05c932e2b4fd7b93fa3a0acae0649@o994439.ingest.sentry.io/5952947',
```
mdnoble73, line 14: Is it ok for this to be in code?

code/web/services/API/UserAPI.php
```diff
@@ -375,18 +375,42 @@ function getPatronProfile() : array
 			}
 
 			$numCheckedOut = 0;
+			$numOverdue = 0;
 			$numHolds = 0;
 			$numHoldsAvailable = 0;
 			$accountSummary = $user->getAccountSummary();
 			$userData->numCheckedOutIls = (int)$accountSummary->numCheckedOut;
 			$userData->numHoldsIls =(int) $accountSummary->getNumHolds();
-			$userData->numHoldsAvailableIls = (int) ($accountSummary->numAvailableHolds == null ? 0 : $accountSummary->numAvailableHolds);
```
mdnoble73, line 383: This was probably done for Arlington? Is casting it to an int causing an issue

code/web/services/API/UserAPI.php
```diff
@@ -375,18 +375,42 @@ function getPatronProfile() : array
...
 			$numCheckedOut += $userData->numCheckedOutIls;
 			$numHolds += $userData->numHoldsIls;
 			$numHoldsAvailable += $userData->numHoldsAvailableIls;
+			$numOverdue += $userData->numOverdue;
+
+			$userData->expires = $accountSummary->expiresOn();
+			$userData->expireClose = $accountSummary->isExpirationClose();
+			$userData->expired = $accountSummary->isExpired();
+
+			if ($user->getLinkedUsers() != null) {
```
mdnoble73, line 397: We probably need a switch for whether or not to include Linked Users.  Arlington handles linked users in a different way in the app so we would want the default off.

code/web/services/API/UserAPI.php
```diff
@@ -405,9 +429,23 @@ function getPatronProfile() : array
 				$userData->numCheckedOutOverDrive = (int)$overDriveSummary->numCheckedOut;
 				$userData->numHoldsOverDrive = (int)$overDriveSummary->getNumHolds();
 				$userData->numHoldsAvailableOverDrive = (int)$overDriveSummary->numAvailableHolds;
-				$numCheckedOut += $userData->numCheckedOutOverDrive;
-				$numHolds += $userData->numHoldsOverDrive;
-				$numHoldsAvailable += $userData->numHoldsAvailableOverDrive;
+				$numCheckedOut += (int)$overDriveSummary->numCheckedOut;
```
mdnoble73, line 432: Similar to above, let's add an option to either include linked user data or not

mdnoble73 (changes requested):
Let's drop the files in .tmp and a few other comments below

code/aspen_app/App.js
catsoup11789: Sentry says DSNs are safe to be in public code, but I'll be reworking this eventually to come from the config file to be able to send errors to different projects for branded apps.

code/web/services/API/UserAPI.php
catsoup11789: (int) was greyed out in phpstorm, like it would never be accessed, but I can add it back in just in case.

mdnoble73 (approved):

PR #526, opened 2022-03-18
22.04.00 lib cal

tnlibrarian:


mdnoble73 (commented):
Can you remove deleting runConfigurations.xml from the commit?

mdnoble73 (changes requested):
Actually if you can remove all the file changes in the .idea directory, that would be great.  I try to avoid committing those as much as possible. 

They are useful for getting setup initially, but can cause lots of conflicts.

code/events_indexer/src/com/turning_leaf_technologies/events/LibraryMarketLibraryCalendarIndexer.java
```diff
@@ -109,7 +109,7 @@ void indexEvents() {
 		if (rssFeed != null){
 			if (doFullReload) {
 				try {
-					solrUpdateServer.deleteByQuery("type:library_calendar AND source:" + this.settingsId);
```
mdnoble73, line 112: This could cause problems for existing libraries with library calendar.  We probably need to delete both source event and source library_calendar

code/web/sys/DBMaintenance/version_updates/22.03.00.php
```diff
@@ -54,127 +54,48 @@ function getUpdates22_03_00() : array
 				'ALTER TABLE grouped_work_test_search ADD COLUMN notes VARCHAR(500)',
 			]
 		], //search_test_notes
-		'search_test_search_index_multiple_terms' => [
```
mdnoble73, line 57: Let's review this, I think all the changes in 22.03 can be removed and the changes can just be in 22.04

mdnoble73 (commented):
There are a few changes that might just be due to changes in whitespace.  We should review to ensure that tabs are used rather than spaces at the start of lines.

mdnoble73 (changes requested):
Looks great!! Just a couple little changes needed

PR #587, opened 2022-06-02
Rt 93843

gabmunoz:


code/web/interface/themes/responsive/RecordDrivers/GroupedWork/browse_result.tpl
```diff
@@ -1,17 +1,21 @@
...
+	{assign var="vSummAuthor" value=""}
+	{if $summAuthor != ''} 
+		{assign var="vSummAuthor" value="by $summAuthor"}
+	{/if}
 	{if $browseMode == '1'}
 		<div class="browse-list grid-item">
 			<a onclick="return AspenDiscovery.GroupedWork.showGroupedWorkInfo('{$summId}', '{$browseCategoryId}')" href="{$summUrl}">
-				<img class="img-responsive" src="{$bookCoverUrl}" alt="{$summTitle} by {$summAuthor}" title="{$summTitle} by {$summAuthor}">
-				<div><strong>{$summTitle}</strong><br> by {$summAuthor}</div>
+				<img class="img-responsive" src="{$bookCoverUrl}" alt="{$summTitle|escape} {$vSummAuthor|escape}" title="{$summTitle|escape} {$vSummAuthor|escape}">
```
mdnoble73, line 9: For this, I think we will need to check to see if the Author is empty and only include the word by and the author if the author is not empty.  
In smarty you can do that with code similar to: 
{if !empty($summAuthor)} by {$summAuthor}{/if}
We should also translate that text. Similar to this: 
{if !empty($summAuthor)}{translate text=" by %1%" 1=$summAuthor inAttribute=true}{/if}

mdnoble73 (changes requested):
See comment, we'll need to update so the author still shows if provided.

mdnoble73 (approved):

PR #631, opened 2022-07-21
Issue 630

AndrewIsh:
As detailed here: https://github.com/mdnoble73/aspen-discovery/issues/630


AndrewIsh:
Something very odd happened with this PR. It was created on top of 22.07.00 but somehow has thousands of commits in it, I'll try again...

PR #665, opened 2022-08-16
SSO

AndrewIsh:


code/web/Drivers/Koha.php
```diff
@@ -2570,6 +2570,8 @@ function selfRegister()
 			}
 
 			$postFields = [];
+			$postFields = $this->setPostField($postFields, 'userid', $library->useAllCapsWhenSubmittingSelfRegistration);
```
mdnoble73, line 2573: Do we need to worry about any form of URL hacking so someone could submit a userid / cardnumber to cause issues?

code/web/sys/DBMaintenance/version_updates/22.08.00.php
```diff
@@ -10,6 +10,31 @@ function getUpdates22_08_00() : array
 				''
 			]
 		], //sample*/
+		'add_library_sso_config_options' => [
```
mdnoble73: Can we move this into 22.09.00.php?

code/web/sys/LibraryLocation/Library.php
```diff
@@ -259,6 +261,27 @@ class Library extends DataObject
 	public $edsSettingsId;
 	public $ebscohostSearchSettingId;
 
+	//SSO
```
mdnoble73, line 264: We may want to pull this out into it's own settings class to make it easier to manage within a consortium and also keep the size of the library settings down a little.

code/web/sys/LibraryLocation/Library.php
```diff
@@ -292,6 +315,7 @@ class Library extends DataObject
 
 	private $_cloudLibraryScopes;
 	private $_libraryLinks;
+	private $curlWrapper;
```
mdnoble73: Let's make this a local variable if we can.

sites/template.linux/conf/crontab_settings.txt
```diff
@@ -62,4 +62,9 @@
 # Fix Cover Permissions    #
 ############################
 */15 * * * * root chown aspen:aspen_apache /data/aspen-discovery/{sitename}/covers/original/*
-*/15 * * * * root chmod g+rw /data/aspen-discovery/{sitename}/covers/original/*
\ No newline at end of file
+*/15 * * * * root chmod g+rw /data/aspen-discovery/{sitename}/covers/original/*
```
mdnoble73, line 65: We'll need to add these into the active crontab as appropriate during upgrade.  There is an example of doing that in 22.08

mdnoble73: Can they run as the aspen user?

install/upgrade_debian.sh
```diff
@@ -21,6 +21,9 @@ if [ -f "/usr/local/aspen-discovery/install/upgrade_debian_$2.sh" ]; then
   /usr/local/aspen-discovery/install/upgrade_debian_$2.sh
 fi
 
+git checkout -b $2
```
mdnoble73: Is this a duplicate of lines 16 & 17? Do we just need to be able to run code before & after the pull? 

We might want to switch to the git checkout -b $2 rather than the git pull everywhere as well.

install/installer_debian.sh
```diff
@@ -52,6 +52,26 @@ systemctl start rngd
 
 apt install -y python-certbot-apache
 
+read -p "Configure SAML Single Sign On? (y/N) " SAMLSSO
```
mdnoble73: We'll want to do this as part of the centos installer as well and we will want to figure out how to install it for servers that already have Aspen setup. Maybe as a separate script that the main installer calls?

mdnoble73 (changes requested):
A few things we want to take a look at

install/upgrade_debian.sh
velvetanchovy: Sorry, this one is my fault - a temporary change I made for a specific problem on the dev server which I didn't revert and can come out

code/web/sys/DBMaintenance/version_updates/22.08.00.php
AndrewIsh: Mark G has moved this to `22.09.00.php`

sites/template.linux/conf/crontab_settings.txt
AndrewIsh: > We'll need to add these into the active crontab as appropriate during upgrade. There is an example of doing that in 22.08

Mark G has created an `updateCron_22_09.php` which appends to `crontab_settings.txt`

install/upgrade_debian.sh
AndrewIsh: Mark G has removed this

code/web/Drivers/Koha.php
AndrewIsh: I don't think so. We call `selfRegister` from `saml2auth.php` which is the script that the IdP returns the client to post-authentication. `saml2auth.php` checks whether the user attributes have been passed by the IdP prior to proceeding. This way, the `$postFields` associative array in the line you mentioned will only ever be dealing with data supplied by the IdP. A malicious request to `saml2auth.php` would fail due to the mutual trust relationship between Aspen and the IdP.

code/web/sys/LibraryLocation/Library.php
AndrewIsh: OK, done

code/web/sys/LibraryLocation/Library.php
AndrewIsh: I'm happy to do this but there doesn't seem to be a precedent for it. There are parts that rely on large data structures in that class, such as `$structure`, there's no precedent for portions of that being derived from separate classes. As I say, I'm happy to do it, and set a precedent, I just wanted to check with you that you're happy with that.

install/installer_debian.sh
AndrewIsh: Mark has moved the portion of `installer_debian.sh` that set up SSO into its own script `samlsso_installer_debian.sh`, so it can be run on a pre-existing Aspen and will set up the SSO.

As far as the CentOS installer is concerned, I'm just looking into this now. I'm not sure we'll be able to install SimpleSAML as a package, as we do with Debian, so it might be a manual process. I'll update here once I've sorted it.

code/web/sys/LibraryLocation/Library.php
mdnoble73: We do it with things like Grouped Work Display Settings where the bulk of the settings are in a separate table and then we just refer to that table by ID.  It's not critical for initial implementation though.

install/installer_debian.sh
mdnoble73: Thanks, let me know how the investigation goes.

mdnoble73 (approved):

install/installer_debian.sh
AndrewIsh: Thanks Mark. I've now created a CentOS install script for this. I've not been able to test it fully so if you have the ability to test this, it would be very much appreciated. I see you've approved and merged this PR, so I'll create a new one for the outstanding changes.

code/web/sys/LibraryLocation/Library.php
AndrewIsh: Ah, I completely missed that, sorry for saying there was no precedent there! Is it OK if I leave this for now, we've had some complications at this end that make it hard to test this. I'll definitely come back to this!

PR #684, opened 2022-08-23
Log rotate

kendralittle:


install/installer.sh
```diff
@@ -63,6 +63,8 @@ timedatectl set-timezone $timezone
 
 #Setup LogRotate
 #cp install/logrotate.conf /etc/logrotate.d/aspen_discovery
```
mdnoble73: These will need to be un-commented to run properly.  We should also look at the debian installer as well as install it in an update script for completeness.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #686, opened 2022-08-23
Indexing Updates

K-Alette:
Format info for Koha is now in overridden loadItemFormat() in KohaRecordProcessor.java. Removed iType boosing/counting. Moved checkRecorForLargePrint() to IlsRecordProcessor.java in loadPrintFormatInformation(). Changed default format source to 'item' for Koha libraries and updated javascript to show correct options for selected format source in ILS indexing menu. Updated release notes

### DIFF
--- a/code/evolve_export/src/org/aspendiscovery/evolve_export/EvolveExportMain.java
+++ b/code/evolve_export/src/org/aspendiscovery/evolve_export/EvolveExportMain.java
@@ -376,6 +376,8 @@ public class EvolveExportMain {
 						logEntry.incErrors("Error parsing marc record", e);
 					}
 				}
+			}else{
+				logEntry.incErrors("Error searching catalog for recently changed titles " + getBibsResponse.getResponseCode() + " " + loginResponse.getMessage());
 			}
 
 			//Also ask for holdings modified from a specific date
--- a/code/web/release_notes/22.09.00.MD
+++ b/code/web/release_notes/22.09.00.MD
@@ -1,9 +1,22 @@
 ###Evergreen Updates
 - Add additional logging for Evergreen API calls
 
+###Evolve Updates
+- Add additional error checking when loading newly changed items.
+
 ###Indexing Updates
 - Allow titles with an iType of ebk to be treated as eContent within the ILS. 
 - Add additional indexes to improve indexing speed
+
+###Interlibrary Loan Updates
+- Add an integration with VDX from OCLC
+  - Allow configuration of which VDX system to connect to.
+  - Allow configuration of which branches within a consortium will lend directly to each other.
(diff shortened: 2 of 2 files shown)


code/reindexer/src/com/turning_leaf_technologies/reindexer/IlsRecordProcessor.java
```diff
@@ -1559,10 +1563,27 @@ public void loadPrintFormatInformation(RecordInfo recordInfo, Record record){
 					recordInfo.addFormatCategory(recordInfo.getFirstItemFormatCategory());
 					return;
 				}
+				try {
```
mdnoble73: We need to move this up with the if (uniqueItemFormats.size() == 1) statemte.  Since if that statement is true, we will never get to this statement.

code/reindexer/src/com/turning_leaf_technologies/reindexer/KohaRecordProcessor.java
```diff
@@ -171,75 +171,46 @@ protected boolean isItemAvailable(ItemInfo itemInfo, String displayStatus, Strin
...
 	@Override
-	public void loadPrintFormatInformation(RecordInfo recordInfo, Record record) {
-		HashMap<String, Integer> itemCountsByItype = new HashMap<>();
-		HashMap<String, String> itemTypeToFormat = new HashMap<>();
-		int mostUsedCount = 0;
-		String mostPopularIType = ""; //Get a list of all the formats based on the items
-		for(ItemInfo item : recordInfo.getRelatedItems()){
-			if (item.isEContent()) {continue;}
+	protected void loadItemFormat(RecordInfo recordInfo, DataField itemField, ItemInfo itemInfo) {
+		HashMap<String, String> itemInfoToFormat = new HashMap<>();
```
mdnoble73: I'm getting a warning that this variable (itemInfoToFormat) is being updated but not read from in IntelliJ.  We'll want to set the format for the item based on the information we find.

code/reindexer/src/com/turning_leaf_technologies/reindexer/KohaRecordProcessor.java
```diff
@@ -171,75 +171,46 @@ protected boolean isItemAvailable(ItemInfo itemInfo, String displayStatus, Strin
...
+		for (ItemInfo item : recordInfo.getRelatedItems()) {
```
mdnoble73: We are already looping through the items in the calling code so we should be able to remove this loop and use itemInfo everywhere we were using item in this method.

mdnoble73 (changes requested):
Just a couple things to look into

PR #712, opened 2022-09-03
22.08.00 hide subject facet

tnlibrarian:


mdnoble73 (approved):

mdnoble73:
I made the same change as part of update, I think we are good.

PR #715, opened 2022-09-09
Fix for installer overwriting php7.3 with php8.1

PerplexedTheta:
* On Debian 10 and Debian 11, php8.1 is selected by the sury list erroneously.
* This proposition will resolve this problem, by specifying to apt which version to use.


kwenoch:
I think I targeted the wrong branch - sorry! Hopefully the commit should be clear to compare.

mdnoble73 (approved):

PR #722, opened 2022-09-13
Evergreen Driver: fix crash when displaying certain fines

gmcharlt:
When a set of billings is associated with a checkout, the response from open-ils.actor.user.transactions.have_charge.fleshed for the underlying transaction will have additional keys with more details about the loan and the title associated with the loan. The billing summary record is the value of the 'transaction' key in all cases, however.

To put it another way, the response from that method is an array of hashes, each of which is guaranteed to have a 'transaction' key, while most but not all will have additional keys.

This patch fixes the handling of the response and avoids a crash that causes no fines to be displayed if one of them is associated with a checkout.

Further enhancement is possible to include title, author, due date, etc. in the Aspen fines display.

### DIFF
--- a/code/web/Drivers/Evergreen.php
+++ b/code/web/Drivers/Evergreen.php
@@ -932,23 +932,21 @@ class Evergreen extends AbstractIlsDriver
 			if ($this->apiCurlWrapper->getResponseCode() == 200) {
 				$apiResponse = json_decode($apiResponse);
 				if (isset($apiResponse->payload)){
-					foreach ($apiResponse->payload[0] as $transactionList){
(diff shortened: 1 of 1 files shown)


mdnoble73:
We're deploying 22.09 tonight and I'd like to do a bit more testing with this before deploying.  Could you resubmit against version 22.10?

gmcharlt:
> We're deploying 22.09 tonight and I'd like to do a bit more testing with this before deploying. Could you resubmit against version 22.10?

Sure, I've rebased it against 22.10.00.

mdnoble73 (approved):

PR #759, opened 2022-10-05
Upgrade OpenSSh to 9.0

kendralittle:


mdnoble73:
Let's tackle this again in 22.11 once we know more the benefits of going to open ssh 9 and we can figure out how to modify the existing httpd files.

PR #767, opened 2022-10-07
SSO: Add definable entityId & XML metadata upload

AndrewIsh:
This commit adds the following:
- Add the ability for the SSO admin to define the entityID for an IdP
- Add the ability to upload metadata as an XML file, this is in addition to the pre-existing ability to specify a URL that the metadata can be retrieved from

### DIFF
--- a/code/web/app/viewIdpMetadata.php
+++ b/code/web/app/viewIdpMetadata.php
@@ -8,14 +8,14 @@
 	$auth = new SAML2Authentication('/app/viewIdpMetadata.php');
 
 	// If we need to forward the user to an IdP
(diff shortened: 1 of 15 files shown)


AndrewIsh:
Hi Mark. This is the final piece in the puzzle for the SSO metadata. I've created the pull request against 22.10.00 but we also need it in 22.09.02, could you also please try merging it into there please? I've not much confidence it'll go without conflicts since it touches files that have definitely changed between 22.09.02 & 22.10.00, if it doesn't go smoothly, let me know and I'll try and come up with another solution. Thanks!

mdnoble73 (approved):

PR #774, opened 2022-10-19
Updates to title_display/displayTitle

K-Alette:
Fix for LiDA title displays - API calls check display_title. Now use short title if no subtitle and concat short and sub titles if subtitle exists. Were previously using sortableTitle since it was set to tmpTitle before displayTitle. removed some redundancy (removal of common formats and punctuation is done for short and sub title).


code/reindexer/src/com/turning_leaf_technologies/reindexer/AbstractGroupedWorkSolr.java
```diff
@@ -702,7 +693,7 @@ void clearSeries(){
 	}
 
 	void addSeriesWithVolume(String seriesName, String volume) {
-		if (series != null) {
```
mdnoble73, line 705: I think this should be checking seriesName rather than series.  I think series is always not null so this check probably never did anything.

code/reindexer/src/com/turning_leaf_technologies/reindexer/AbstractGroupedWorkSolr.java
```diff
@@ -466,27 +466,18 @@ void setTitle(String shortTitle, String subTitle, String displayTitle, String so
 				//remove punctuation from the sortable title
 				sortableTitle = punctuationPattern.matcher(sortableTitle).replaceAll("");
 				this.titleSort = sortableTitle.trim();
-				displayTitle = AspenStringUtils.trimTrailingPunctuation(displayTitle);
```
mdnoble73, line 469: We still want this section I think otherwise we wind up showing information that isn't compatible with all of the editions.

mdnoble73 (changes requested):

code/reindexer/src/com/turning_leaf_technologies/reindexer/AbstractGroupedWorkSolr.java
K-Alette: we're just concating the short and sub titles with new suggestion so as long as both of those have these checks to remove punctuation and such we shouldn't need to do it again?

PR #796, opened 2022-11-02
Adds db update

NashvillePublicLibrary:
+ includes sideload log in numInvalidRecords, fixing sideload reindexer


mdnoble73:
Added this right before I saw the pull request

PR #806, opened 2022-11-05
22.11.00 export excel

tnlibrarian:


code/web/services/MaterialsRequest/ManageRequests.php
```diff
@@ -383,15 +383,15 @@ function exportToExcel($selectedRequestIds, $allRequests){
                     $activeSheet->setCellValueByColumnAndRow($curCol++, $curRow, $request->getCreatedByUser()->email);
 
                     // Place hold?
-                    if ($request->placeHoldWhenAvailable == 1){
+                    if ($request->placeHoldWhenAvailable == 1 || !empty(request->holdPickupLocation)){
```
mdnoble73, line 386: For this, let's give the actual value set.  I can see a library setting it to false after they place the hold.

mdnoble73 (changes requested):
Can you resubmit against 22.11? Plus the one change for place hold when available.

tnlibrarian:
I had goofed and based on 22.12.00 instead of 22.11.00...

PR #824, opened 2022-11-18
Fixes, updates to various things

catsoup11789:


code/web/Drivers/Koha.php
```diff
@@ -60,8 +58,9 @@ function updateHomeLibrary(User $patron, string $homeLibraryCode)
 		];
 		$oauthToken = $this->getOAuthToken();
 		if ($oauthToken == false) {
-			$result['messages'][] = translate(['text' => 'Unable to authenticate with the ILS.  Please try again later or contact the library.', 'isPublicFacing'=>true]);
-		} else {
+			$result['messages'][] = translate(['text' => 'Unable to authenticate with the ILS.  Please try again later or contact the library.', 'isPublicFacing' => true]);
+		}
```
mdnoble73: Can we update the code style so the else has the else on the same line as the closing brace of the previous if?
} else {
rather than 
} 
else {

code/web/services/API/ListAPI.php
```diff
@@ -23,17 +21,20 @@ function launch()
 					header("Cache-Control: max-age=300");
 					require_once ROOT_DIR . '/sys/SystemLogging/APIUsage.php';
 					APIUsage::incrementStat('ListAPI', $method);
-				} else {
+				}
```
mdnoble73: Let's fix the formatting of the else here as well.

mdnoble73 (changes requested):
Why did we delete the GothicA1-Bold.ttf font?
Also, there are a few formatting things that are making reviewing hard with some of the big files. 
Can you update the release notes as well?

mdnoble73 (approved):

PR #828, opened 2022-11-23
BasicPage should return better HTTP response codes

lucasmontoya13:
This patch makes accessing a basic page without the right permissions return a 401 instead of a plain 404 as it did before.

- response codes now match messages.
- new files added: Handle401.php and 401.tpl


mdnoble73 (changes requested):
Could you do the same change for PortalPage.php as well so that both Basic and Custom pages will get the 401 response appropriately? 

Thanks!

- Mark

PR #830, opened 2022-11-23
Prompt User to Change PIN

K-Alette:
Prompt user to change PIN after removing managing account - user will be told to contact library if pin reset is not allowed or they do not have a home library.
Added ability to reload page after closing modal with buttons
Updated release notes


mdnoble73:
Looks like Pullbot ate this pull request, can you resubmit? Sorry :(

PR #831, opened 2022-11-23
Update VDX Cost field

K-Alette:
Updated VDX form to separate ReqMaxCostCurr and ReqMaxCost
Updated release notes


mdnoble73 (changes requested):
Can you update code style from the PHPStorm export in aspendiscovery/code and the reformat so the indenting is done with tabs rather than spaces?

PR #832, opened 2022-11-25
LiDA updates

catsoup11789:


code/web/services/API/UserAPI.php
```diff
@@ -1023,97 +1110,138 @@ function getOverDriveLendingOptions() : array
...
 				}
-			} else if ($source == 'hoopla') {
-				return $this->openHooplaItem();
-			} else if ($source == 'cloud_library') {
-				return $this->openCloudLibraryItem();
-			} else if ($source == 'axis360') {
-				return $this->openAxis360Item();
 			} else {
-				return array('success' => false, 'message' => 'Invalid source');
+				if ($source == 'hoopla') {
```
mdnoble73: Let's see if we can keep the if and else on the same lines.  I'll check settings

mdnoble73 (changes requested):
Try turning the "Special 'else if' treatment on in code style and reformat UserAPI and AJAX.php that should cut down on the number of changes.

PR #838, opened 2022-11-30
PortalPage should return better HTTP response codes

lucasmontoya13:
This patch makes accessing a custom page without the right permissions return a 401 instead of a plain 404 as it did before.

- response codes now match messages.

### DIFF
--- a/code/web/services/WebBuilder/PortalPage.php
+++ b/code/web/services/WebBuilder/PortalPage.php
@@ -3,45 +3,21 @@
 class WebBuilder_PortalPage extends Action
 {
 	/** @var PortalPage */
-	private $page;
+	private $portalPage;
 
 	function __construct()
 	{
 		parent::__construct();
-
-		//Make sure the user has permission to access the page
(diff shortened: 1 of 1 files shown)


code/web/services/WebBuilder/PortalPage.php
```diff
@@ -3,80 +3,77 @@
...
 		}
+		$this->display('portalPage.tpl', $title, '', false);
 	}
 
 	function canView() : bool
 	{
-		require_once ROOT_DIR . '/sys/WebBuilder/PortalPageAccess.php';
-		require_once ROOT_DIR . '/sys/Account/PType.php';
-		require_once ROOT_DIR . '/sys/WebBuilder/PortalPage.php';
+		$this->portalPage->find();
```
mdnoble73, line 69: We shouldn't need to repeat the ->find() and ->fetch(true) statements here since that is already done in the launch.

mdnoble73 (approved):

PR #851, opened 2022-12-02
Small Linking Fixes

K-Alette:
- Go to /MyAccount/ResetPinPage instead of /MyAccount/RequestPinReset 
- Submit processAddLinkedUser() when hitting enter on the username/password input modal for adding a link
- Add prompt to change pin for users disabling account linking (if library does not allow, tell user to contact library) 
- If user removes managing account, notify that managing account and link to /MyAccount/LinkedAccounts


code/web/interface/themes/responsive/js/aspen/account.js
```diff
@@ -652,7 +653,7 @@ AspenDiscovery.Account = (function () {
 		},
 
 		redirectPinReset: function() {
-			window.location.href = Globals.path + "/MyAccount/RequestPinReset";
```
mdnoble73, line 655: This might need to change depending on ILS

PR #870, opened 2022-12-12
Debian Installer Updates take 2

HitScan:
Rather than copying whole files just make the necessary edits or
use partial files for *.d/ style configs. (php and mysql config;
solr security limits)

Simplify installer_debian.sh a little and make it behave better if
run a second time. Also use more portable constructions for some
bash-specific builtins used even though the shebang is #!/bin/sh.
(user / group creation, 'read -p' vs 'printf ; read', path creation
and perms)

Rather than calling a script with a #!/bin/sh shebang as an argument
to bash, just change the shebang to #!/bin/bash. (samlsso_installer_debian.sh)

You can't always assume #!/bin/sh is bash (it isn't on recent Debian,
certainly) so a pass with shellcheck at some point would be a good
idea for most shell scripts.

### DIFF
--- a/install/installer_debian.sh
+++ b/install/installer_debian.sh
@@ -1,73 +1,72 @@
 #!/bin/sh
 
-#This will need to be copied to the server manually to do the setup.
-#Expects to be installed on Debian 10 Buster
-#Run as sudo ./installer.sh
-apt update
-apt install -y wget
-apt install -y apache2
-apt -y install apt-transport-https lsb-release ca-certificates curl
-wget -O /etc/apt/trusted.gpg.d/php.gpg https://packages.sury.org/php/apt.gpg
-sh -c 'echo "deb https://packages.sury.org/php/ $(lsb_release -sc) main" > /etc/apt/sources.list.d/php.list'
-apt update
-apt install -y php7.3 php7.3-mcrypt php7.3-gd php7.3-curl php7.3-mysql php7.3-zip
-apt install -y php7.3-xml
-apt install -y bind9 bind9utils
-apt install -y php7.3-intl
-apt install -y php7.3-mbstring
-apt install -y php7.3-pgsql
-apt install -y php7.3-ssh2
-service apache2 start
-systemctl enable apache2
-# New PHP ini file
-# - Change max_memory to 256M (from 128M)
-# - Increase max file size to 50M
-# - Increase max post size to 50M
-mv /etc/php/7.3/apache2/php.ini /etc/php/7.3/apache2/php.ini.old
-cp php.ini /etc/php/7.3/apache2/php.ini
+#Expects to be installed on Debian 10 Buster or later
+#Run as sudo ./installer_debian.sh
+apt-get update
+apt-get -y install gpg openjdk-11-jre-headless openjdk-11-jdk-headless apache2 certbot python3-certbot-apache mariadb-server apt-transport-https lsb-release ca-certificates curl zip
+
+# Install "plain" php package here to determine OS default. Override below if desired.
+if ! dpkg -l | grep ii | grep -qE ' php[0-9]+\.[0-9]+ ' ; then
+  apt-get install -y php
+fi
+
+php_vers="$(dpkg -l | grep ii | grep -E ' php[0-9]+\.[0-9]+ ' | grep -Eo '[0-9]+\.[0-9]+' | head -1)"
+
+if test -z "$php_vers" ; then
+  echo "Unable to determine default php version!"
+  exit 1
+fi
+
+# Install Ondrej Sury's php repo for additional php modules
+# Specifically, the abandonware php-mcrypt...
+keyrings="/etc/apt/keyrings"
+test -d "$keyrings" || (mkdir -p "$keyrings" ; chmod 0755 "$keyrings")
+if ! test -f "$keyrings/sury.gpg" || ! test -f /etc/apt/sources.list.d/sury.list ; then
+  wget -q -O - https://packages.sury.org/php/apt.gpg | gpg -o "$keyrings/sury.gpg" --dearmor
+  echo "deb [signed-by=$keyrings/sury.gpg] https://packages.sury.org/php/ $(lsb_release -sc) main" > /etc/apt/sources.list.d/sury.list
+  apt-get update
+fi
+
+# Have to use versions for these or the highest version available from sury.org is used rather than the system verison.
+apt-get install -y "php${php_vers}-mcrypt" "php${php_vers}-gd" "php${php_vers}-curl" "php${php_vers}-mysql" "php${php_vers}-zip" "php${php_vers}-xml" "php${php_vers}-intl" "php${php_vers}-mbstring" "php${php_vers}-soap"
+
+# - Change max_memory to 256M
+# - Increase max file size to 75M
+# - Increase max post size to 75M
+php_ini="/etc/php/${php_vers}/apache2/php.ini"
+# Necessary or surprise baggage from dragging around an old php.ini?
+#grep -q '^max_input_vars = 2000' "$php_ini" || sed -Ei 's/^;max_input_vars = [0-9]+/max_input_vars = 2000/' "$php_ini"
+grep -q '^memory_limit = 256M' "$php_ini" || sed -Ei 's/^memory_limit = [0-9]+M/memory_limit = 256M/' "$php_ini"
+grep -q '^post_max_size = 75M' "$php_ini" || sed -Ei 's/^post_max_size = [0-9]+M/post_max_size = 75M/' "$php_ini"
+grep -q '^upload_max_filesize = 75M' "$php_ini" || sed -Ei 's/^upload_max_filesize = [0-9]+M/upload_max_filesize = 75M/' "$php_ini"
+
+./samlsso_installer_debian.sh
+
+# MariaDB overrides
+cp 60-aspen.cnf /etc/mysql/mariadb.conf.d/
+
 a2enmod rewrite
-apt install -y mariadb-server
-mv /etc/mysql/mariadb.cnf /etc/mysql/mariadb.cnf.old
-cp mariadb.cnf /etc/mysql/mariadb.cnf
-systemctl start mariadb
-systemctl enable mariadb
-apt install -y software-properties-common
-apt install -y default-jdk
-apt install -y openjdk-11-jdk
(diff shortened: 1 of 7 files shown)


data_dir_setup/update_solr_files.sh
```diff
@@ -9,4 +9,4 @@ echo "Updating $1"
 cp -r solr7 /data/aspen-discovery/$1
 sudo chown -R solr:aspen /data/aspen-discovery/$1/solr7
 
-sudo -u solr /usr/local/aspen-discovery/sites/$1/$1.sh restart
+sudo su solr -c  "/usr/local/aspen-discovery/sites/$1/$1.sh restart"
```
mdnoble73: I updated 23.01 so we can drop the need for sudo here. We already expect the command to be run as root.

install/installer_debian.sh
```diff
@@ -1,73 +1,72 @@
...
-systemctl daemon-reload
-systemctl start rngd
+if test -z "$php_vers" ; then
+  echo "Unable to determine default php version!"
+  exit 1
+fi
 
-apt install -y python-certbot-apache
+# Install Ondrej Sury's php repo for additional php modules
+# Specifically, the abandonware php-mcrypt...
```
mdnoble73: We can probably drop php-mcrypt.  Are there any other php modules that require using the non-standard location?

install/installer_debian.sh
```diff
@@ -1,73 +1,72 @@
...
-echo "Generate new root password for mariadb at: https://passwordsgenerator.net/ and store in passbolt"
-mysql_secure_installation
-echo "Enter the timezone of the server"
-read timezone
-timedatectl set-timezone $timezone
+# - Change max_memory to 256M
+# - Increase max file size to 75M
+# - Increase max post size to 75M
+php_ini="/etc/php/${php_vers}/apache2/php.ini"
+# Necessary or surprise baggage from dragging around an old php.ini?
```
mdnoble73: I'm ok with removing copying the php.ini file, let's remove the comment though.  If there are additional concerns, let me know.

install/installer_debian.sh
```diff
@@ -1,73 +1,72 @@
...
-apt -y install apt-transport-https lsb-release ca-certificates curl
-wget -O /etc/apt/trusted.gpg.d/php.gpg https://packages.sury.org/php/apt.gpg
-sh -c 'echo "deb https://packages.sury.org/php/ $(lsb_release -sc) main" > /etc/apt/sources.list.d/php.list'
-apt update
-apt install -y php7.3 php7.3-mcrypt php7.3-gd php7.3-curl php7.3-mysql php7.3-zip
-apt install -y php7.3-xml
-apt install -y bind9 bind9utils
-apt install -y php7.3-intl
-apt install -y php7.3-mbstring
-apt install -y php7.3-pgsql
```
mdnoble73, line 18: It looks like this is no longer installed

install/installer_debian.sh
```diff
@@ -1,73 +1,72 @@
...
-apt install -y php7.3-ssh2
```
mdnoble73, line 19: It looks like this is no longer installed

install/installer_debian.sh
```diff
@@ -1,73 +1,72 @@
...
+  apt-get update
+fi
 
-bash ./samlsso_installer_debian.sh
+# Have to use versions for these or the highest version available from sury.org is used rather than the system verison.
+apt-get install -y "php${php_vers}-mcrypt" "php${php_vers}-gd" "php${php_vers}-curl" "php${php_vers}-mysql" "php${php_vers}-zip" "php${php_vers}-xml" "php${php_vers}-intl" "php${php_vers}-mbstring" "php${php_vers}-soap"
 
-echo "Generate new root password for mariadb at: https://passwordsgenerator.net/ and store in passbolt"
-mysql_secure_installation
-echo "Enter the timezone of the server"
```
mdnoble73, line 59: Is there a reason this portion (setting time zone was removed)?

install/setup_aspen_user_debian.sh
```diff
@@ -1,16 +1,18 @@
...
+
 #create a new user and user group to run Aspen Discovery
-adduser aspen
-#Add all existing users to the group
+grep -q aspen /etc/passwd || useradd -m -s /bin/bash -G aspen_apache aspen
+#Add all existing users to the aspen group
 for ID in $(cat /etc/passwd | grep /home | cut -d ':' -f1); do (usermod -a -G aspen $ID);done
 
-#create an aspen_apache group as well for files that need to be readable (and writable) by apache
-groupadd aspen_apache
```
mdnoble73, line 8: It looks like adding www-data to the aspen_apache group was removed

mdnoble73 (changes requested):
Some questions and changes in the review comments. Let me know if you have questions. 

- Mark

mdnoble73 (approved):

install/installer_debian.sh
mrenvoize: Do we actually need both mysql and pqsql drivers?  I was surprised to see both installed when reviewing the script a while back, but I've not delved deep enough into the codebase to understand why yet, so took it for what it was.

install/installer_debian.sh
```diff
@@ -1,73 +1,72 @@
...
+  echo "deb [signed-by=$keyrings/sury.gpg] https://packages.sury.org/php/ $(lsb_release -sc) main" > /etc/apt/sources.list.d/sury.list
+  apt-get update
+fi
 
-bash ./samlsso_installer_debian.sh
+# Have to use versions for these or the highest version available from sury.org is used rather than the system verison.
+apt-get install -y "php${php_vers}-mcrypt" "php${php_vers}-gd" "php${php_vers}-curl" "php${php_vers}-mysql" "php${php_vers}-zip" "php${php_vers}-xml" "php${php_vers}-intl" "php${php_vers}-mbstring" "php${php_vers}-soap"
 
-echo "Generate new root password for mariadb at: https://passwordsgenerator.net/ and store in passbolt"
-mysql_secure_installation
```
mrenvoize, line 58: This was probably a bit ptfs-e specific, happy for it's removal, I'll make sure it's just documented at our end to ensure it's run as part of server setups.

install/installer_debian.sh
mdnoble73: PGSQL is for connection to Sierra so not needed everywhere, but needed in some cases.

install/installer_debian.sh
HitScan: It doesn't look like Aspen is using php-mcrypt directly at all, but it's used by the PHPExcel (for PDFs) and recaptcha (for mailhide) modules. When I initially tried leaving out Sury's repo I think that was the only package that couldn't be installed.

install/installer_debian.sh
HitScan: Sorry, yeah, a few of these comments are me just leaving notes to myself and I forgot to pull them out.

install/installer_debian.sh
HitScan: Pretty much all of the php modules should still be installed, I just do it all at once instead of one here and one there. If we can get rid of the Sury.org repo I can also drop all of the versioning variable stuff and just default to the packaged versions.

I did intentionally leave out bind9 and bind9utils because that didn't seem to be used anywhere.

install/installer_debian.sh
HitScan: It does look like I missed php7.3-ssh2, I'll do another pass on the module installation to make sure nothing else was missed.

install/installer_debian.sh
HitScan: That was me intending to change it to dpkg-reconfigure but forgetting to actually put it in place.

install/setup_aspen_user_debian.sh
HitScan: The usermod -a -G ... line above does that. Param -a "appends" secondary groups and -G is a secondary group to add the user to. I moved things around a little so the aspen_apache group could be added to the aspen user at create time.

data_dir_setup/update_solr_files.sh
HitScan: I must have just missed that; I'll drop this change

HitScan:
I've added another commit to address the comments.

install/installer_debian.sh
mdnoble73: We're going to drop PHPExcel and export to CSV directly so that can be removed. It looks like recaptcha degrades with a warning if mcrypt is not deployed.  We're testing to verify the mailhide is not in use (we don't think it should be)

install/installer_debian.sh
mdnoble73: It looks like bind9 and bind9utils are only installed on debian so should be ok to remove them.

install/installer_debian.sh
mdnoble73: This needs to be put back in place before I can accept the PR

mdnoble73 (changes requested):

HitScan:
Specifically a call to timedatectl or something else? (the mysql_secure_installation call is also still present, just moved down) The "dpkg-reconfigure tzdata" line handles both asking and setting the TZ on Debian. That said, I was mistaken in thinking that using timedatectl doesn't correctly set the TZ permanently so I don't have strong feelings about which way that's handled.

mdnoble73:
Can you resubmit with installing PHP 8 rather than PHP 7? 
Thanks!

HitScan:
By removing all of the version checking stuff and still using Sury's repo we automatically install on php 8.2 on Debian. If Aspen needs a specific php 8 point release I can put some of that back. (I haven't experimented much beyond making sure that the installer works)

mdnoble73:
Our servers are running 8.0 in general.  8.2 should work, but we are not testing it specifically at this point. I'll work on getting this merged. Thanks

mdnoble73:
Our servers are running 8.0 in general. 8.2 should work, but we are not testing it specifically at this point. I'll work on getting this merged. Thanks

mdnoble73:
Do you want to resubmit without the version checking now or for 23.02?

HitScan:
Yeah, I think just aiming for 23.02.00 is best at this point. I'll default the version to 8.0 so there's still control over what point release is used. That will also allow me to squash the various commits so it's less messy.

HitScan:
Closing this PR to resume on 23.02.00 in PR 964.

PR #894, opened 2022-12-28
Account Linking Fixes

K-Alette:
Fix spacing issue in display of linked accounts on /Account/Home 
Fix issue where linking was not honoring pType setting


code/web/sys/Account/User.php
```diff
@@ -2312,7 +2312,7 @@ public function getPType() {
 		require_once ROOT_DIR . '/sys/Account/PType.php';
 		$patronType = new PType();
 		$patronType->pType = $this->patronType;
-		return $patronType;
```
mdnoble73, line 2315: You're going to need to do a $patronType->find(true) before returning the value here.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #935, opened 2023-01-11
add Aspen LiDA Build Tracker, hold sorting

catsoup11789:


code/web/services/API/UserAPI.php
```diff
@@ -772,9 +772,11 @@ function getPatronHolds(): array {
 		} else {
 			$user = $this->getUserForApiCall();
 			if ($user && !($user instanceof AspenError)) {
+				$unavailableSort = $_REQUEST['unavailableSort'] ?? 'title';
```
mdnoble73: It looks like this should default to 'sortTitle' here

mdnoble73 (changes requested):
One little tiny change

mdnoble73 (approved):

PR #942, opened 2023-01-13
Updates for PHP 8/Smarty

K-Alette:
- Added summId to variables - used in title-rating.tpl
- Fix for error when looking at formats for overdrive titles (undefined property: OverDriveAPIProductFormats::$size in view-formats.tpl)


code/web/interface/themes/responsive/OverDrive/view-formats.tpl
```diff
@@ -19,7 +19,7 @@
 			</div>
 		</div>
 
-		{if $overDriveFormat->size != 0 && strcasecmp($overDriveFormat->size, 'unknown') != 0}
+		{if !empty($overDriveFormat)}
```
mdnoble73: I think we need to preserve the other 2 checks here.  So !empty($overDriveFormat) && ....

PR #964, opened 2023-01-23
Debian Installer Updates the third

HitScan:
Rather than copying whole files just make the necessary edits or use partial files for *.d/ style configs. (php and mysql config; solr security limits)

Simplify installer_debian.sh a little and make it behave better if run a second time. Also use more portable constructions for some bash-specific builtins used even though the shebang is #!/bin/sh. (user / group creation, 'read -p' vs 'printf ; read', path creation and perms)

Rather than calling a script with a #!/bin/sh shebang as an argument to bash, just change the shebang to #!/bin/bash. (samlsso_installer_debian.sh)

You can't always assume #!/bin/sh is bash (it isn't on recent Debian, certainly) so a pass with shellcheck at some point would be a good idea for most shell scripts.

Signed-off-by: Jason Boyer <JBoyer@equinoxOLI.org>

### DIFF
--- a/code/cron/nightly_mysql_dump.sh
+++ b/code/cron/nightly_mysql_dump.sh
@@ -35,11 +35,7 @@ if [[ $# -eq 0 ]]; then
 else
 ASPENSERVER=$1
 
-if [[ $# -eq 2 ]]; then
-	DBNAME=$2
-else
-	DBNAME="aspen"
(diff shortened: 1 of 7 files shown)


HitScan:
Putting this PR on hold so I can rework a couple things.

PR #977, opened 2023-01-26
Debian installer updates

HitScan:
Rather than copying whole files just make the necessary edits or use partial files for *.d/ style configs. (php and mysql config; solr security limits)

Simplify installer_debian.sh a little and make it behave better if run a second time. Also use more portable constructions for some bash-specific builtins used even though the shebang is #!/bin/sh. (user / group creation, 'read -p' vs 'printf ; read', path creation and perms)

You can't always assume #!/bin/sh is bash (it isn't on recent Debian, certainly) so a pass with shellcheck at some point would be a good idea for most shell scripts.

Signed-off-by: Jason Boyer <jboyer@equinoxOLI.org>

### DIFF
--- a/code/cron/nightly_mysql_dump.sh
+++ b/code/cron/nightly_mysql_dump.sh
@@ -35,11 +35,7 @@ if [[ $# -eq 0 ]]; then
 else
 ASPENSERVER=$1
 
-if [[ $# -eq 2 ]]; then
-	DBNAME=$2
-else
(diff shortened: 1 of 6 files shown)


HitScan:
Sorry about all of the repeat PRs; I was using the default branches in my gh repo which made things a lot harder to do properly.

mdnoble73 (approved):

PR #990, opened 2023-01-31
Holds Messaging Updates

K-Alette:
- Add popup for when title is already on hold for user to confirm placing a duplicate hold.
- Disable hold button for when user submits hold form through the modal submit form and reloadHolds when action is taken

### DIFF
--- a/code/web/Drivers/SirsiDynixROA.php
+++ b/code/web/Drivers/SirsiDynixROA.php
@@ -2077,6 +2077,9 @@ class SirsiDynixROA extends HorizonAPI {
 
 							if (isset($_REQUEST['preferredName'])) {
 								$updatePatronInfoParameters['fields']['preferredName'] = $_REQUEST['preferredName'];
+								if ($homeLibrary->setUsePreferredNameInIlsOnUpdate) {
+									$updatePatronInfoParameters['fields']['usePreferredName'] = !empty($_REQUEST['preferredName']);
+								}
 								$patron->_preferredName = $_REQUEST['preferredName'];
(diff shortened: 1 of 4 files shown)


code/web/interface/themes/responsive/Record/hold-popup.tpl
```diff
@@ -161,7 +167,7 @@
 						{/foreach}
 					{/if}
 					<br>
-					{if $showLogMeOut == 1}
+					{if $showLogMeOut == 1 && !$isOnHold}
```
mdnoble73: I think it's probably ok to leave this here if they do decide to place the hold they can still be logged out after.

code/web/services/Record/AJAX.php
```diff
@@ -205,6 +205,10 @@ function getPlaceHoldForm() {
 			require_once ROOT_DIR . '/RecordDrivers/MarcRecordDriver.php';
 			$marcRecord = new MarcRecordDriver($id);
 
+			require_once ROOT_DIR . '/sys/Account/User.php';
+			$isOnHold = (new User)->isRecordOnHold($recordSource, $id);
```
mdnoble73: Rather than creating a new user object here, we can use $user that is defined earlier since that represents the active user.

mdnoble73 (changes requested):
A couple little changes.  We should talk with Jordan about what to do with volume and item level holds.  I think it's probably ok to ignore those since they are more rare and more likely the user would actually want multiple holds in normal cases.  We just want to make sure that the new logic doesn't interrupt any of those workflows if we don't make changes.

PR #1033, opened 2023-02-24
Linked accounts: disable disable

NashvillePublicLibrary:
For patrons with patron type that do not have permission "Allow users to remove managing account links", do not show "Disable Account Linking" button


NashvillePublicLibrary:
bad code! I'll fix in a different pull request

PR #1053, opened 2023-03-02
Tooltip edits + increase char limit for ILL URL + adjust some default values in library sys settings

thelastcatbender:
I don't know why there are two other weird commits in here. The one from Mar 1 is the only one with changes that I'm actually trying to submit.


mdnoble73 (approved):

mdnoble73:
Includes extra changes, merged manually and closing this

PR #1076, opened 2023-03-09
Mimic centos 23.01.00 php8 upgrade for debian

Jacobomara901:
This patch aims to mimic the centos 23.01.00 script by diabling any php version currently in use, installing the same packages as in the centos script and enabling php8 in apache.


mdnoble73 (changes requested):
Could you put this in upgrade_debian_23.03.00.sh instead? That way it will run automatically during the upgrade to 23.03 or alternatively we could do it as part of both 23.01 and 23.03, but I think you're currently on 23.02.

Jacobomara901:
No problem, moved the changes in to upgrade_debian_23.03.00.sh as requested. I dont think there is any need to have them in both, it was only contained in 23.01.00 originally to match the centos upgrade version

mdnoble73 (approved):

PR #1129, opened 2023-04-06
Evergreen: password recovery and reset

gmcharlt:
This set of patches adds support for patron password reset and recovery when using Evergreen. It also proposes a minor renaming of some page templates used by ILS-driver-based password recovery.


mdnoble73:
Let's create separate template files for koha and evergreen.  We have separate ones for symphony now so long term maintenance will be easier with them being separate now.

gmcharlt:
OK, I've updated the pull request to use separate templates.

mdnoble73 (approved):

PR #1140, opened 2023-04-11
fixes, app privacy policy

catsoup11789:


code/web/services/GroupedWork/AJAX.php
mdnoble73, line 1: We'll need to undo this so permissions for uploaded files don't conflict with ones downloaded from Koha

mdnoble73 (approved):

PR #1227, opened 2023-06-02
Your Events Display Order

K-Alette:
Fixes order that events are displayed in Your Events (I think the previous fix for this was lost somehow?)


mdnoble73:
We merged this back in earlier

PR #1239, opened 2023-06-12
Fixes to tooltips

thelastcatbender:
Fixing some of my errors with quotation marks and HTML tags


thelastcatbender:
Forgot to mention that I've also attempted to put the System Messages in library system settings in an accordion section so it's less immediately visible, and moved the "Generate Sitemap" setting above it.

mdnoble73 (approved):

PR #1242, opened 2023-06-13
Custom Default Covers

K-Alette:
Adds ability to use a custom image for default covers

- Added column defaultCover to themes table
- Added ability to upload a default cover image in Theme settings 
- Updated DefaultCoverImageBuilder.php to use uploaded image if one exists 
- Updated release notes

### DIFF
--- a/code/web/release_notes/23.07.00.MD
+++ b/code/web/release_notes/23.07.00.MD
@@ -35,8 +35,8 @@
 
 <div markdown="1" class="settings">
 
-#### Revised Settings
-- New setting in Themes to add an image to be used for default covers 
+#### New Settings
+- New setting in Themes > Default Cover to add an image to be used for default covers 
 </div>
 
 // other
--- a/code/web/release_notes/23.07.00.MD
+++ b/code/web/release_notes/23.07.00.MD
@@ -30,6 +30,14 @@
 // kirstien
 
 // kodi
+###Default Cover Updates
+- Add ability to choose a default cover image for default covers generated for library catalog items (Tickets 113975, 100969)
+
+<div markdown="1" class="settings">
+
+#### Revised Settings
+- New setting in Themes to add an image to be used for default covers 
+</div>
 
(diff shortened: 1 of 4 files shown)


code/web/sys/Covers/DefaultCoverImageBuilder.php
```diff
@@ -113,11 +117,39 @@ public function getCover($title, $author, $filename) {
 		imagefilledrectangle($imageCanvas, 0, 0, $this->imageWidth, $this->topMargin, $backgroundColor);
 
 		$artworkHeight = $this->drawArtwork($imageCanvas, $backgroundColor, $foregroundColor, $title);
```
mdnoble73, line 119: Do we want this in the else (if defaultCoverImage is empty) or do we need to always draw the artwork?

code/web/release_notes/23.07.00.MD
```diff
@@ -30,6 +30,14 @@
 // kirstien
 
 // kodi
+###Default Cover Updates
+- Add ability to choose a default cover image for default covers generated for library catalog items (Tickets 113975, 100969)
+
+<div markdown="1" class="settings">
+
+#### Revised Settings
+- New setting in Themes to add an image to be used for default covers
```
mdnoble73: Let's specify the path to the new setting here, i.e. Themes > Default Cover. Also update to be New Settings rather than Revised Settings

mdnoble73 (changes requested):
Have we verified that only jpg and png files can be uploaded? we might need to add that restriction into DataObjectUtils.  We can separate that out into a separate story too.

code/web/sys/Covers/DefaultCoverImageBuilder.php
K-Alette: we do need to draw the artwork to get the artworkHeight but I added a check for the default cover in that function so nothing actually gets drawn

PR #1244, opened 2023-06-16
fixes for SAML SLS

catsoup11789:


code/web/services/MyAccount/Logout.php
```diff
@@ -7,12 +7,20 @@ class MyAccount_Logout extends Action {
...
 		if(UserAccount::isLoggedInViaSSO()) {
 			global $library;
+			global $configArray;
 			require_once ROOT_DIR . '/sys/Authentication/SSOSetting.php';
 			$ssoSettings = new SSOSetting();
 			$ssoSettings->id = $library->ssoSettingId;
 			if ($ssoSettings->find(true)) {
 				if ($ssoSettings->ssoSPLogoutUrl) {
 					$_REQUEST['return'] = $ssoSettings->ssoSPLogoutUrl;
+				} else {
```
mdnoble73: I don't think this else is going to hit since $ssoSPLogoutUrl will be set in the first if block.  Does this if just need to go inside the initial if?

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #1248, opened 2023-06-27
Update setup_server.bat

NashvillePublicLibrary:


setup_server.bat
```diff
@@ -1,41 +1,45 @@
...
+mkdir %HOST%
+cd %HOST%
+
+echo Installing Solr Files for HOST
+cd %WD%/data_dir_setup/
+update_solr_files.bat %HOST%
+
+echo In lieu of "Creating symbolic link in /etc/httpd/conf.d to apache config file"
+echo Windows apache config file (httpd.conf) needs an Include pointing to %HOST% file, e.g.,
+echo Include "c:/web/aspen-discovery/sites/nashville.aspenlocal/httpd-nashville.aspenlocal.conf"
```
mdnoble73: This needs to be parameterized

setup_server.bat
```diff
@@ -1,41 +1,45 @@
 @echo off
-rem Setup data and logs for a new server by copying the appropriate files from default. 
-set SERVERNAME=%1
+rem Setup data and logs for a new aspen-discovery server by copying the appropriate files from default.
+set HOST=%1
```
mdnoble73, line 3: Let's leave this SERVERNAME for consistency with other setup files

mdnoble73 (changes requested):

setup_server.bat
NashvillePublicLibrary: Because createSite.php already instructs user to insert the Include statement, I just made the whole mess a comment in the Windows setup server batch file. Zehr gut?

mdnoble73:
We reviewed and decided these scripts are no longer needed

PR #1289, opened 2023-07-18
Cookie consent

Jacobomara901:
This patch adds a Cookie consent banner for GDPR purposes. It can be toggled on under system variables and additionally allows you to add your cookie policy as html to be displayed when the user requests more information.

Test plan:
 - Log in as aspen_admin user
 - Navigate to system variables in the administration menu
 - Click edit and scroll to bottom
 - Notice new data protection section and toggle cookie consent on. Save changes.
 - *New cookie consent banner should now appear*
 - Set new cookie policy HTML in same section of system variables.
 - Notice these changes are shown when more information is requested on the cookie consent banner
 - Accept cookies and check cookie is assigned correctly in browser dev tools.
 - Check that IP storage checkbox still works as intended in new data protection section, then navigating to "Usage by IP" in the administration section.
When checked: IP addresses should be listed.
When not checked: This page should not list any IP addresses


code/web/sys/DBMaintenance/version_updates/23.07.00.php
catsoup11789, line 1: Can you move these db changes to the 23.08.00 file?

catsoup11789 (changes requested):

Jacobomara901:
Hey, sorry for the delay on this! 
Changes for 23.08.00.php are at the top of the file as I was tracking 23.07.00 and thought this would be the easiest fix in terms of merge conflicts. If you'd prefer something different next time please let me know and I will pass it on to the rest of the team on this end

code/web/index.php
```diff
@@ -233,6 +233,18 @@
...
+$interface->assign('cookieStorageConsent', false);
+try {
+	require_once ROOT_DIR . '/sys/SystemVariables.php';
+	$systemVariables = new SystemVariables();
+	if ($systemVariables->find(true) && !empty($systemVariables->cookieStorageConsent)) {
+		$interface->assign('cookieStorageConsent', true);
+		$interface->assign('cookieStorageConsentHTML', $systemVariables->cookiePolicyHTML);
+	}
+} catch (Exception $e) {
+	AspenError::raiseError("Error loading cookie consent banner");
```
mdnoble73: Rather than raising an error here, let's just leave the default.  This shouldn't get triggered, but could if the system variables table hasn't been created yet.

code/web/interface/themes/responsive/cookie-consent.tpl
```diff
@@ -0,0 +1,17 @@
+{if !$smarty.cookies.cookieConsent}
+    <div class="stripPopup">
+        <div class="container">
+            <div class="contentWrap">
+                <span>We use cookies on this site to enhance your user experience.</span>
```
mdnoble73: This needs to be translatable

code/web/interface/themes/responsive/cookie-consent.tpl
```diff
@@ -0,0 +1,17 @@
...
+                <abbr>By clicking any link on this website, you are giving us consent to set and store cookies.</abbr>
```
mdnoble73: Is abbr the correct tag for this? I don't think the semantics are quite right here? Also needs translation.

code/web/interface/themes/responsive/cookie-consent.tpl
```diff
@@ -0,0 +1,17 @@
...
+            </div>
+            <div class="btnWrap">
+                <a href="#" id="consentAgree" class="button">Yes, I agree</a>
```
mdnoble73: Needs translation

code/web/interface/themes/responsive/cookie-consent.tpl
```diff
@@ -0,0 +1,17 @@
...
+                <a href="#" id="consentDisagree" class="button">No, I want to find out more</a>
```
mdnoble73: Needs translation, how are we providing more info? I don't see it yet, but also haven't read all the way through the code

code/web/interface/themes/responsive/cookie-consent.tpl
```diff
@@ -0,0 +1,17 @@
...
+            </div>
+        </div>
+    </div>
+    <link rel="stylesheet" href="/interface/themes/responsive/css/cookie-consent.css" />
```
mdnoble73: The CSS should all get merged into the main.css file which we build using a File Watcher in PHP Storm so you shouldn't need to include the file separately.

code/web/interface/themes/responsive/cookie-consent.tpl
```diff
@@ -0,0 +1,17 @@
...
+    <script>cookiePolicyHTML='{$cookieStorageConsentHTML|escape:javascript|regex_replace:"/[\r\n]/" : " "}'</script>
```
mdnoble73: This will want to be defined in globals.js and then set in cssAndJsIncludes

code/web/interface/themes/responsive/cookie-consent.tpl
```diff
@@ -0,0 +1,17 @@
...
+    <script src="/interface/themes/responsive/js/aspen/cookieConsent.js"></script>
```
mdnoble73: This will get merged into aspen.js. The merge is done as part of merge_javascript.php so it won't need to get inserted here

code/web/interface/themes/responsive/css/cookie-consent.less
```diff
@@ -0,0 +1,96 @@
...
+    background-color: rgb(29, 127, 240);
+    padding: 15px 0;
+    position: fixed;
+    bottom: 0;
+    left: 0;
+    width: 100%;
+    z-index: 1;
+    font-family: 'Poppins', sans-serif;
+
+    .btnWrap{
```
mdnoble73, line 13: I don't think we want to override .btnWrap here. It's probably better to give it a new class with the customizations needed and even with that we should think about why we want to override them from the defaults

code/web/interface/themes/responsive/css/cookie-consent.less
```diff
@@ -0,0 +1,96 @@
+@import url('https://fonts.googleapis.com/css2?family=Poppins:wght@200;400;700&display=swap');
+
+*{
```
mdnoble73: Overriding this for everything will probably cause lots of layout errors

code/web/interface/themes/responsive/cookie-consent.tpl
mdnoble73: These buttons will need translations. Also stylistically, we tend to have these provide onclick handlers in the .tpl file for clarity.

code/web/interface/themes/responsive/cookie-consent.tpl
mdnoble73: This will want to be defined in globals.js and then can be updated in cssAndJsIncludes

code/web/interface/themes/responsive/cookie-consent.tpl
mdnoble73: This should be included as part of the main.css file that we merge with a PHPStorm file watcher

code/web/interface/themes/responsive/js/aspen/cookieConsent.js
```diff
@@ -0,0 +1,14 @@
+$(document).ready(function () {
```
mdnoble73: These should get merged into the aspen.js file using merge_javascript.php so they can be called with onclick handlers within the tpl.

code/web/sys/DBMaintenance/version_updates/23.08.00.php
```diff
@@ -0,0 +1,18 @@
...
+			],
+		], //enable/disable cookie consent banner
+		
+		'cookie_policy_html' => [
+			'title' => 'Cookie Policy HTML',
+			'description' => 'Allow admin to set HTML for cookie policy',
+			'continueOnError' => true,
+			'sql' => [
+				"ALTER TABLE system_variables ADD COLUMN cookiePolicyHTML TEXT",
+				"UPDATE system_variables set cookiePolicyHTML = 'This body has not yet set a cookie storage policy, please check back later.'",
```
mdnoble73: Let's run the default text by Jordan to see what she thinks

mdnoble73 (changes requested):

code/web/index.php
Jacobomara901: Removed the raising of AspenError

code/web/interface/themes/responsive/cookie-consent.tpl
Jacobomara901: You are correct i believe, I've changed this to a small instead.

code/web/interface/themes/responsive/cookie-consent.tpl
Jacobomara901: Added

code/web/interface/themes/responsive/cookie-consent.tpl
Jacobomara901: Added

code/web/interface/themes/responsive/cookie-consent.tpl
Jacobomara901: Translation added. Was originally using jquery for an onclick check which presented with an aspen modal populated with the html set in sysvars

Jacobomara901: should now be more stylistically cohesive with other areas of aspen

code/web/interface/themes/responsive/cookie-consent.tpl
Jacobomara901: both have now been added

code/web/interface/themes/responsive/css/cookie-consent.less
Jacobomara901: You are correct, overriding is in fact not necessary at all so the css has been removed.

code/web/interface/themes/responsive/cookie-consent.tpl
Jacobomara901: This has now been done (though just using lessc manually as we do not currently use phpstorm) I couldn't see any issues with the command but if there are any please let me know and I will amend.

code/web/interface/themes/responsive/cookie-consent.tpl
Jacobomara901: The above has now been merged in to aspen.js and the amalgamated file has been added to the refactor commit.

code/web/interface/themes/responsive/cookie-consent.tpl
Jacobomara901: onclick handlers have been added instead and translations added also.

code/web/interface/themes/responsive/cookie-consent.tpl
Jacobomara901: Added to main.css and removed from the .tpl file

code/web/interface/themes/responsive/cookie-consent.tpl
Jacobomara901: both have now been added.

code/web/interface/themes/responsive/js/aspen/cookieConsent.js
Jacobomara901: as above, this has now been amended.

code/web/sys/DBMaintenance/version_updates/23.08.00.php
Jacobomara901: Agreed, can be added to the list for Monday. as such, I have not changed this with the rebase commit.

code/web/release_notes/23.08.00.MD
mdnoble73, line 1: It looks like this replaced the previous file. I can manually merge, but would be better if your PR had the file with changes from everyone else.

code/web/sys/DBMaintenance/version_updates/23.08.00.php
mdnoble73, line 1: It looks like this replaced the previous file. I can manually merge, but would be better if your PR had the file with changes from everyone else.

code/web/interface/themes/responsive/css/cookie-consent.less
```diff
@@ -0,0 +1,90 @@
+@import url('https://fonts.googleapis.com/css2?family=Poppins:wght@200;400;700&display=swap');
```
mdnoble73: This import should not be needed

mdnoble73 (changes requested):
A couple files are replacing the previous version.  Can you fix those and also remove the extra import?

code/web/interface/themes/responsive/cookie-consent.tpl
```diff
@@ -0,0 +1,14 @@
+{if !$smarty.cookies.cookieConsent}
```
mdnoble73: This is generating a notice.  You should be able to check 
if empty($smarty.cookies.cookieConsent)

code/web/interface/themes/responsive/css/cookie-consent.less
mdnoble73, line 1: The default styling on this is generating white buttons on a white bar within a larger blue bar so the text and buttons aren't readable.  We're going to need to see if we can use existing colors from the theme for the buttons and we'll need to make the foreground and background of the consent bar able to be controlled by the theme.

mdnoble73 (changes requested):
We need to do a bit more with the theming for this, let's target getting it into 23.09 once we get that branch going rather than pushing it into 23.08 since that is due for demo tomorrow.

code/web/interface/themes/responsive/cookie-consent.tpl
Jacobomara901: I've replaced the above

code/web/interface/themes/responsive/css/cookie-consent.less
Jacobomara901: This should now be fixed, my .container class was being overridden by the theming. Now replaced with a new classname.

Jacobomara901:
Hi Mark, thanks for looking back over this. I've taken a quick look this morning and will separate the theming integration in to a separate PR to follow shortly. How would you suggest I go about adding the integration to themes? From my understanding: I could add it to the theme.css.tpl file, use jquery/selectors or use the getDefaultColor method from admin.js. Let me know what youd prefer and i'll get it over to you asap

mdnoble73 (approved):

PR #1294, opened 2023-07-18
A new 'DBPort' entry has been added to the Aspen section in createSite configuration

lucasmontoya13:


sites/template.linux/conf/config.pwd.ini.template
```diff
@@ -11,7 +11,7 @@ minTimeToLog    = 0.05
 database_dsn          = "mysql:dbname={aspenDBName};host={aspenDBHost}"
```
mdnoble73: I think we're going to need to update the database_dsn as well to include the port here.

install/createSiteTemplate.ini
```diff
@@ -21,6 +21,8 @@ timezone =
 [Aspen]
 ; Database host for Aspen
 DBHost =
+; Database port for Aspen
+DBPort =
```
mdnoble73: Let's just default this to 3306 so it's one less thing to customize by default.

mdnoble73 (changes requested):
Looks like we will also need to update importAspenBackup.php, backupAspen.php, and createDefaultDatabaseScript to use the defined host as well.  

Then we'll also want to set a default port within config.ini in the Database section so Aspen can default back to that.

PR #1296, opened 2023-07-19
Little improvements

lucasmontoya13:
- A new variable that refers to a new connection to an aspen database.
- Port variable has been added to database dsn in config template

### DIFF
--- a/install/createSite.php
+++ b/install/createSite.php
@@ -280,17 +280,18 @@ if (!$siteOnWindows) {
 }
 
 //Import the database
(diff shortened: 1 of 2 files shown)


sites/template.linux/conf/config.pwd.ini.template
mdnoble73, line 1: we'll need to make the same change to config.pwd.ini.template within template.windows

install/createSite.php
```diff
@@ -280,17 +280,18 @@
 }
 
 //Import the database
+$aspen_db_cmd = "mysql -u{$variables['aspenDBUser']} -p\"{$variables['aspenDBPwd']}\" -h\"{$variables['aspenDBHost']}\" --port \"{$variables['aspenDBPort']}\"";
```
mdnoble73, line 283: Can we change this to $mysqlConnectionCommand?

mdnoble73 (changes requested):
Looking good, just a couple of little changes

PR #1335, opened 2023-08-12
self check changes, api changes

catsoup11789:


code/web/interface/themes/responsive/Record/hold-popup.tpl
```diff
@@ -153,11 +153,29 @@
 					{if $showHoldCancelDate == 1}
 						<div id="cancelHoldDate" class="form-group">
 							<label class="control-label" for="cancelDate">{translate text="Automatically cancel this hold if not filled by" isPublicFacing=true}</label>
-							<input type="date" name="cancelDate" id="cancelDate" placeholder="mm/dd/yyyy" class="form-control" size="10" min="{$smarty.now|date_format:"%Y-%m-%d"}">
+							<input type="date" name="cancelDate" id="cancelDate" placeholder="mm/dd/yyyy" value="{$smarty.now|date_format:"%Y-%m-%d"}" class="form-control" size="10" min="{$smarty.now|date_format:"%Y-%m-%d"}">
```
mdnoble73: Do we want to provide a default date here?  I think that might break things since we handle the blank date in other places.

mdnoble73: Can we just set a min value here instead?

mdnoble73 (changes requested):
Can you take a look at setting the minimum date for the cancellation date again?

code/web/interface/themes/responsive/Record/hold-popup.tpl
catsoup11789: When I didn't set a default value, the date validation didn't like the values provided because it comes across as an invalid value if something isn't provided for every part (ie they leave everything as-is but only modify the "day" value) and then wouldn't properly reset; still allowing them to enter the invalid date. Min value was already set and did nothing for manually entered values (Google tells me this is a known issue with the date input fields)

catsoup11789: I definitely see the issue with having a date by default if they don't want to have a hold cancelation date. Maybe we add a checkbox to enable/disable the date field? And we can set the default day to like +30 days or something?

catsoup11789: We can also disable keyboard interactions with the field to force use of the date picker --- but I'm not sure about accessibility implications this may have.

mdnoble73: I'll check with Jordan to see what she thinks. We might need to validate things server side and just come back with an error message. I'm guessing this doesn't really happen in practice so we don't want to make it harder for normal usage.

catsoup11789: @mdnoble73 I went ahead and just commented out these changes so that I can get the other changes in to continue work on other things tomorrow.

mdnoble73: Talked with Jordan today and she was thinking that it's best to leave the field with a blank value and then on the server side we can validate that the date is in the future.  If it's in the past we can just return an error message and they can resubmit.  That is probably the least amount of disruption to the normal workflow.

catsoup11789: Can you go ahead and merge the commented code for now so I can finish working on self-check stuff? I will revisit the validation later.

catsoup11789: Re-worked this so it checks the date when processing submitHoldForm.

mdnoble73 (approved):

PR #1365, opened 2023-08-24
23.09.00 e commerce

tnlibrarian:


code/web/interface/themes/responsive/Admin/propertiesListCSV.tpl
```diff
@@ -0,0 +1,177 @@
...
+					,
+				{/if}
+			{/foreach}
+			{* CSV record delimiter *}
+			{"\n"}
+		{/foreach}
+	{/if}
+{/strip}
+
+{*
```
mdnoble73: Let's remove this commented out bit

code/web/services/Admin/DonationsReport.php
```diff
@@ -18,14 +18,25 @@ function getPageTitle(): string {
...
-			$objectList[$object->id] = clone $object;
-		}
+        $objectList = [];
+        if (UserAccount::userHasPermission('View Donations Reports for All Libraries')){
+            $object->find();
+            while ($object->fetch()) {
+                $objectList[$object->id] = clone $object;
+            }
+        } elseif (UserAccount::userHasPermission('View Donations Reports for Home Library')) {
+            $locationList = Location::getLocationListAsObjects(true);
```
mdnoble73: We should be able to optimize this with a ->whereAddIn($listOfLocations) where $listOfLocations is the locations for the library

code/web/services/Admin/ObjectEditor.php
```diff
@@ -224,26 +227,38 @@ function viewExistingObjects($structure) {
...
+                    'perPage' => $recordsPerPage,
+                    'canChangeRecordsPerPage' => true,
+                    'canJumpToPage' => true,
+                ];
+                $pager = new Pager($options);
+                $interface->assign('pageLinks', $pager->getLinks());
+            }
+            if (isset($_REQUEST['objectAction']) && $_REQUEST['objectAction'] == 'exportSelectedToCSV') {
+                $exportObjects = [];
+                foreach ($_REQUEST['selectedObject'] as $k=>$v) {
```
mdnoble73: This needs to check to see if anything has been selected using isset or empty to see if selectedObjects have been passed.

code/web/services/Admin/eCommerceReport.php
```diff
@@ -18,14 +18,21 @@ function getPageTitle(): string {
...
-		$objectList = [];
-		while ($object->fetch()) {
-			$objectList[$object->id] = clone $object;
-		}
+        $object->orderBy($this->getSort());
+        $this->applyFilters($object);
+        $object->limit(($page - 1) * $recordsPerPage, $recordsPerPage);
+        $objectList = [];
+        if (UserAccount::userHasPermission('View eCommerce Reports for Home Library')) {
+            $homeLibrary = Library::getPatronHomeLibrary()->libraryId;
```
mdnoble73, line 26: This might be able to be simplified?

mdnoble73 (changes requested):
Looking good, just a little cleanup!

mdnoble73:
Initial review comments: 
- All reports are downloaded with the name DonationsReport.csv, we should pass that from the ObjectEditor maybe based on the page title for the active editor? 
- If no rows are selected, there are 2 notices that get displayed.  We should either catch this with JavaScript (best) or process everything if nothing is selected, might need to check with Jordan. 
- Should we always always quote values? Right now some are quoted and some aren't
- Can you reformat with code styles so tabs etc at start of line are correct

code/web/interface/themes/responsive/Admin/propertiesListCSV.tpl
tnlibrarian: d'oh!

code/web/services/Admin/ObjectEditor.php
tnlibrarian: https://github.com/tnlibrarian/aspen-discovery/commit/5a6b2747b1b4b2e782fb6da7cb74a474ac471e78

code/web/services/Admin/eCommerceReport.php
mdnoble73: Nevermind

mdnoble73 (approved):

PR #1369, opened 2023-08-24
Image Upload by URL

K-Alette:
Fix for cover images not uploaded at the grouped work level when using URL upload


mdnoble73:
Let's put this in 23.08.01 if we need a patch for it

PR #1401, opened 2023-09-08
check for companionSystems when updating

catsoup11789:


code/web/cron/runScheduledUpdate.php
```diff
@@ -152,7 +205,7 @@ function updateGitAndRunDatabaseUpdates($operatingSystem, $versionToUpdateTo, Sc
 	}
 }
 
-function runDatabaseMaintenance($versionToUpdateTo, $scheduledUpdate) {
+function runDatabaseMaintenance($versionToUpdateTo, $scheduledUpdate, CompanionSystem $companionSystem = null) {
 	// run db maintenance
 	$scheduledUpdate->notes .= "Running database maintenance $versionToUpdateTo\n";
 	require_once ROOT_DIR . '/services/API/SystemAPI.php';
```
mdnoble73: This block only needs to run if it's not a companion system.  I think it can go in an else check for the new code to update the companion system

mdnoble73 (changes requested):

PR #1427, opened 2023-09-22
Masquerade Mode Check

K-Alette:
Add check for guidingUserId vs activeUserId in ContactInformation.php to prevent masqueraders from updating patron info in Aspen 
Updated release notes

### DIFF
--- a/code/web/Drivers/Koha.php
+++ b/code/web/Drivers/Koha.php
@@ -2824,14 +2824,14 @@ class Koha extends AbstractIlsDriver {
 				$error = $renewResponse->error;
 				$success = false;
 				$message = 'The item could not be renewed: ';
-				$message = $this->getRenewErrorMessage($error, $message);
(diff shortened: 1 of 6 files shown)


code/web/services/MyAccount/ContactInformation.php
```diff
@@ -20,7 +20,13 @@ function launch() {
 			global $librarySingleton;
 			// Get Library Settings from the home library of the current user-account being displayed
 			$patronHomeLibrary = $user->getHomeLibrary(true);
-			if ($patronHomeLibrary == null) {
+			$currentUser = $_SESSION['activeUserId'];
```
mdnoble73: We're going to need to add the active userId as a hidden form variable to the contactInfo.tpl and other settings forms that don't currently store the user id when the form was displayed.  Then when we submit we will need to check to see if the userid set when the form displayed is the same as the active userid.

mdnoble73 (changes requested):

PR #1428, opened 2023-09-22
Disable Cookie Consent

mrenvoize:
Comment out ability to enabled cookies consent control whilst we await further follow-ups from @Jacobomara901


mdnoble73:
Rather than commenting out the setting, let's update so that cookieStorageConsent is not set to true within index.php and add a note to the setting to explain that this functionality is not available.  That way if anyone has turned the functionality on, the cookie consent form will not show and they can understand why.

mrenvoize:
@Jacobomara901 is back now and is going to have a look at this for us.

mdnoble73:
Looks great, thanks!  Welcome back Jacob!

mdnoble73 (approved):

PR #1457, opened 2023-10-20
Refactor: Moved language and display options into burger menu

AlexanderBlanchardAC:
The burger menu now contains a section for languages, which displays all languages currently active on the user's account and a section for displays to reflect the themes available. This supports accessibility and creates continuity between the display whether logged in or not.
To Test:
1)Prior to implementation, if not logged in you should see a button in the header that will read "Languages and Display" or "Languages". When clicked, a modal opens up so that you can make your selection.
2)Log in and note that to change the language, you need to use the burger menu. Click on an available language and note that you are taken to a preferences page.
3)Implement new code and note that you are not shown the "Languages and Display" button, regardless of whether you are logged in or not.
4)Use the burger menu to select an available language and note that the language can now be selected directly from the menu, the page should not redirect you, but alter the display so that it is now in your chosen language.

### DIFF
--- a/code/web/interface/themes/responsive/WebBuilder/quickPoll.tpl
+++ b/code/web/interface/themes/responsive/WebBuilder/quickPoll.tpl
@@ -30,7 +30,7 @@
 				{$submissionError}
 			</div>
 	    {/if}
-		<form id="quickPoll{$id}" class="form-horizontal" role="form" action="/WebBuilder/SubmitQuickPoll"  onsubmit="setFormSubmitting();">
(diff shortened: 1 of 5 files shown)


mdnoble73:
I want to talk this through with Jordan to see what she thinks and see if we need a toggle to allow it to work either way.

mdnoble73:
Talked today during our joint meeting and discussed that this will likely need a toggle.  Martin will follow up as well.

mdnoble73:
@AlexanderBlanchardAC we will talk about this at the gathering, but if you want to proactively create a toggle for whether or not the button in the header is enabled or not on the library settings we can work on merging this in.

PR #1458, opened 2023-10-20
Koha apirest class

lucasmontoya13:
-Added a new class that facilitates interaction with Koha's API
- For testing purposes, the KohaAPIRESTClient class has been used in the patternLogin method within the Koha class which uses a POST operation.


code/web/services/API/KohaRESTAPIClient.php
mdnoble73, line 1: I'm trying to think of the best place to put this.  /services/API is where we have the APIs that Aspen exposes.  I think that this is probably best within 
/Drivers since that is where we have all of the code to connect to the various ILSs we support

code/web/Drivers/Koha.php
```diff
@@ -821,32 +822,14 @@ public function patronLogin($username, $password, $validatedViaSSO) {
...
-					], true);
-				}
-				$responseBody = $this->apiCurlWrapper->curlSendPage($apiURL, 'POST', json_encode($postParams));
-				$responseCode = $this->apiCurlWrapper->getResponseCode();
-				$jsonResponse = json_decode($responseBody);
-				$cardNumber = $jsonResponse->cardnumber;
-				$patronId = $jsonResponse->patron_id;
-				ExternalRequestLogEntry::logRequest('koha.patronLogin', 'POST', $apiURL, $this->curlWrapper->getHeaders(), json_encode($postParams), $responseCode, $responseBody, ['password' => $password]);
-				if ($responseCode == 201) {
+				$postResponse = $this->kohaRestAPIClient->post("/api/v1/auth/password/validation",$params,"koha.patronLogin",['password' => $password]);
```
mdnoble73, line 834: I think we'll need the API wrapper to be able to return both the Response Code and the body of the response since several cases require checking the response code before we check the body.  Maybe we create another class that has properties for the response code and the body that we can pass back from API wrapper?

mdnoble73 (changes requested):
Let me know if you have questions on any of the comments

mdnoble73:
Can you update based on the latest code in 23.11?  We've found a few cases where there were PHP Notices and Warnings generated if the patron could not be logged in and I want to make sure we preserve those updates.

mdnoble73:
Let's talk through this with Tomas next week to make sure that we are covering all of the cases of patron authentication and also see if we can use this class for all API calls.

PR #1460, opened 2023-10-23
- The "request through ILL" button redirects the complete search for the WorldCat service.

lucasmontoya13:


mdnoble73:
I think we're going to need to check if the active ILL system is a a WorldCat system or not.  Here is a link to a non WorldCat ILL System that we can use for comparison.  https://nashville.illiad.oclc.org/illiad

We may need to have the user specify the ILL System within the Library System settings. 
For now we can probably use the following values: 
- INN-Reach
- WorldCat
- Other

code/web/services/Search/Results.php
```diff
@@ -548,8 +548,17 @@ function launch() {
 		$interface->assign('exploreMoreSection', 'catalog');
 		$interface->assign('showExploreMoreBar', $showExploreMoreBar);
 		$interface->assign('exploreMoreSearchTerm', $exploreMoreSearchTerm);
-
 		$interface->assign('sectionLabel', 'Library Catalog');
+
+		$activeLibrary = Library::getActiveLibrary();
```
mdnoble73: The active library is stored in the global $library which can be referenced with: 
global $library
(if it is not already loaded in the file)

code/web/services/Search/Results.php
```diff
@@ -548,8 +548,17 @@ function launch() {
...
+		$selectedILL = $activeLibrary->getInterLibraryLoanInfo();
```
mdnoble73: Rather than calling getInterLibraryLoanInfo and bundling the properties into a new array, let's just call the properties directly.

code/web/services/Search/Results.php
```diff
@@ -548,8 +548,17 @@ function launch() {
...
+		if (isset($selectedILL['ILL System']) && $selectedILL['ILL System'] == 1 ) {
+			$searchSource = new SearchSources();
+			$interLibraryLoanURL = $searchSource->getExternalLink('worldcat',$_REQUEST['searchIndex'],$_REQUEST['lookfor']);
```
mdnoble73: I researched and it looks like for INN-Reach we can call 
$searchSource->getExternalLink('innReach', $_REQUEST['searchIndex'], $_REQUEST['lookfor']) just like you're doing for worldcat.

code/web/sys/LibraryLocation/Library.php
```diff
@@ -3748,6 +3762,14 @@ public function getMasqueradeStatus(): int {
 		return $this->allowMasqueradeMode;
 	}
 
+	public function getInterLibraryLoanInfo(): array {
```
mdnoble73: Let's just call these properties directly instead of creating a new method to bundle them.

mdnoble73 (changes requested):
Looking great, just a few style updates.  Can you also update the 23.11.00.MD file with release notes describing the changes you made? There is a section for you to add to.

code/web/sys/DBMaintenance/version_updates/23.11.00.php
```diff
@@ -40,5 +40,13 @@ function getUpdates23_11_00(): array {
 		],
 		//Jacob - PTFS
 		//other
+		'select_ILL_system' => [
+			'title' => 'Dropbox ILL systems',
+			'description' => 'Add a setting to allow users to specify ILL system used.',
+			'continueOnError' => true,
+			'sql' => [
+				'ALTER TABLE  library ADD COLUMN ILLSystem TINYINT(1) DEFAULT 0',
```
mdnoble73: Let's default this to 2 - Other 
That way we won't apply the search parameters incorrectly.

mdnoble73 (changes requested):
One last change just for which default we have

PR #1461, opened 2023-10-23
Cookie consent user storage

Jacobomara901:
This PR changes the cookie consent cookie to a session based cookie, allows storage on a per user basis and allows the preferences to be amended in myPreferences. Also redesigns the cookie consent banner slightly to be more user friendly 

This WILL need a small patch to go alongside to disable google analytics when analytics cookies have not been allowed but would like your input on where you would like it disabled  (I was thinking index.php - around line 138) let me know if you have a preference. Should just be an if statement so literally 30 seconds worth of work.

Test plan:
Enable cookie consent in library settings
Allow cookies,
Check that cookies are set,
delete cookie and check cookie is reapplied from DB on refresh,
delete cookie from both browser and set the cookies to 0 in the user table of the DB to re-enable the banner on refresh
check that the option for "only allow essential cookies" keeps analytics set to 0 in the DB
set cookies preferences to 0 and test that if you set your preferences and log in after, your preferences are saved in the db.
Check that you can amend your preferences in the myPreferences page (you should not be able to modify essential cookies preferences as these are essential)

I also still have the DB changes in 23.10.00 at the moment as there does not seem to be a file for 23.11.00 yet. If you need it moved, please let me know.

### DIFF
--- a/code/web/index.php
+++ b/code/web/index.php
@@ -135,31 +135,36 @@ $timer->logTime('Loaded display options within interface');
 
 global $active_ip;
 
+
 try {
-	require_once ROOT_DIR . '/sys/Enrichment/GoogleApiSetting.php';
-	$googleSettings = new GoogleApiSetting();
-	if ($googleSettings->find(true)) {
-		$googleAnalyticsId = $googleSettings->googleAnalyticsTrackingId;
-		$googleAnalyticsLinkingId = $googleSettings->googleAnalyticsTrackingId;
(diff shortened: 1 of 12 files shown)


mdnoble73:
Let's look at how this works when we have an old cookieConsent value stored (mine is just a string so it doesn't decode as json).  Let's also look at how it works if the library is not opting in.  From my initial testing it looks like when the user is not opted in and the library has cookie consent off that tracking will not be included.

mdnoble73 (approved):

PR #1464, opened 2023-10-25
Cron Updates & Barcode Prefixes

K-Alette:
Updated crontab_settings.txt in sites directory for those that needed it
Added option to use barcode prefixes for symphony custom self registration forms
Updated release notes

### DIFF
--- a/code/web/Drivers/SirsiDynixROA.php
+++ b/code/web/Drivers/SirsiDynixROA.php
@@ -753,7 +753,42 @@ class SirsiDynixROA extends HorizonAPI {
 				$createPatronInfoParameters['fields']['phoneList'][] = $cellPhoneInfo;
 			}
 
-			//TODO: We should be able to create either a random barcode or a barcode starting with a specific prefix and choose the length.
+			$barcodePrefix = new SelfRegistrationForm;
+			$barcodePrefix->id = $library->selfRegistrationFormId;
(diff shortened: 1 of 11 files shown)


code/web/Drivers/SirsiDynixROA.php
```diff
@@ -753,7 +753,42 @@ function selfRegister(): array {
...
+						$selfRegResult = [
+							'success' => true,
+							'barcode' => $barcodePrefix . $barcodeAppend,
+							'requirePinReset' => true,
+						];
+					}
+				}
+			}
+
 			$barcode = new Variable();
```
mdnoble73: Does this need to go in an else or do we need to check to make sure that the user has not already been created?  We don't want two created for the patron.

sites/model2.production/conf/crontab_settings.txt
mdnoble73, line 1: We can leave this one not running clam av since it's on the same system that model is on.  We probably need to think about how to do this for systems that include companion systems.  But for now, let's just leave it off running for model2

mdnoble73 (changes requested):

code/web/Drivers/SirsiDynixROA.php
K-Alette: youu are correct :) fixing now

mdnoble73 (approved):

PR #1485, opened 2023-11-03
Update UserAPI.php 

K-Alette:
Make UserAPI.php honor pType restrictions for account linking
Updated release notes

### DIFF
--- a/code/web/services/API/UserAPI.php
+++ b/code/web/services/API/UserAPI.php
@@ -4404,37 +4404,85 @@ class UserAPI extends Action {
 		$accountToLink = UserAccount::validateAccount($accountToLinkUsername, $accountToLinkPassword);
 		$patron = UserAccount::validateAccount($username, $password);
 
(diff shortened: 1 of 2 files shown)


code/web/services/API/UserAPI.php
```diff
@@ -4404,37 +4404,85 @@ function addAccountLink() {
 		$accountToLink = UserAccount::validateAccount($accountToLinkUsername, $accountToLinkPassword);
 		$patron = UserAccount::validateAccount($username, $password);
 
+		require_once ROOT_DIR . '/sys/Account/PType.php';
+		$user = UserAccount::getLoggedInUser();
```
catsoup11789: Can we just re-use the $patron user here instead of trying to call a new user? Sometimes getLoggedInUser acts a little funny over API requests.

K-Alette: done!

catsoup11789 (approved):

PR #1488, opened 2023-11-07
Symphony Self Registration Updates

K-Alette:
-displayName field is now required
-Updated city/state field logic
-Added prompting for parent name to custom forms
-Updated release notes

### DIFF
--- a/code/web/Drivers/SirsiDynixROA.php
+++ b/code/web/Drivers/SirsiDynixROA.php
@@ -3358,22 +3358,22 @@ class SirsiDynixROA extends HorizonAPI {
 		} else {
 			//Use self registration fields
 			/** @var SelfRegistrationFormValues $customField */
+			if ($library->promptForParentInSelfReg){
+				$fields[] = [
+					'property' => 'cardType',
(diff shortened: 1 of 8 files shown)


code/web/Drivers/SirsiDynixROA.php
```diff
@@ -3365,6 +3361,25 @@ public function getSelfRegistrationFields() {
 			foreach ($customFields as $customField) {
 				if ($customField->symphonyName == 'library') {
 					$fields[$customField->symphonyName] = $pickupLocationField;
+				} elseif(($customField->symphonyName == 'parentname' || $customField->symphonyName == 'care_of' || $customField->symphonyName == 'careof') && $library->promptForParentInSelfReg) {
+					$fields[] = [
```
mdnoble73: I think this block to show or hide the Type of Card needs to only be shown based on the promptForParentInSelfReg.  But we need to always show the field.

mdnoble73 (changes requested):
See inline comments

mdnoble73 (approved):

PR #1524, opened 2023-11-29
Gmcharlt/use unapi for evergreen bib holdings fetch

gmcharlt:


gmcharlt:
Noting that this has been tested in production.

mdnoble73:
Is there a minimum version of Evergreen needed for this to work?

gmcharlt:
> Is there a minimum version of Evergreen needed for this to work?

2.10, but that version is seven years old and very unlikely to be encountered nowadays.

PR #1526, opened 2023-11-30
OverDrive to Custom Reader Name

K-Alette:
Changes to make "OverDrive" a custom set name
Updated release notes
Note: java file changes not included since I am working on changes to make the value scoped

### DIFF
--- a/code/web/RecordDrivers/Axis360RecordDriver.php
+++ b/code/web/RecordDrivers/Axis360RecordDriver.php
@@ -69,6 +69,10 @@ class Axis360RecordDriver extends GroupedWorkSubDriver {
 		$interface->assign('bookcoverInfo', $this->getBookcoverInfo());
 
 		$interface->assign('axis360Extract', $this->axis360RawMetadata);
+		$readerName = new OverDriveDriver();
+		$readerName = $readerName->getReaderName();
+		$interface->assign('readerName', $readerName);
+
 		return 'RecordDrivers/Axis360/staff-view.tpl';
 	}
 
--- a/code/web/RecordDrivers/MarcRecordDriver.php
+++ b/code/web/RecordDrivers/MarcRecordDriver.php
@@ -259,6 +259,10 @@ class MarcRecordDriver extends GroupedWorkSubDriver {
 
 		$interface->assign('uploadedSupplementalFiles', $this->getUploadedSupplementalFiles());
 
+		$readerName = new OverDriveDriver();
+		$readerName = $readerName->getReaderName();
+		$interface->assign('readerName', $readerName);
+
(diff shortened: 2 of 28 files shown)


code/web/RecordDrivers/Axis360RecordDriver.php
```diff
@@ -69,6 +69,10 @@ public function getStaffView() {
 		$interface->assign('bookcoverInfo', $this->getBookcoverInfo());
 
 		$interface->assign('axis360Extract', $this->axis360RawMetadata);
+		$readerName = new OverDriveDriver();
```
mdnoble73, line 72: Do we need to load the reader here?

code/web/RecordDrivers/MarcRecordDriver.php
```diff
@@ -259,6 +259,10 @@ public function getStaffView() {
 
 		$interface->assign('uploadedSupplementalFiles', $this->getUploadedSupplementalFiles());
 
+		$readerName = new OverDriveDriver();
```
mdnoble73, line 262: Do we need the readerName loaded here?

code/web/RecordDrivers/Axis360RecordDriver.php
mdnoble73: Or does this need to be set to Boundless rather than the OverDrive reader?

code/web/services/OverDrive/AspenData.php
```diff
@@ -40,13 +40,20 @@ function launch() {
 			$interface->assign('overDriveId', '');
 		}
 
+		$readerName = new OverDriveDriver();
+		$readerName = $readerName->getReaderName();
+
+		$interface->assign('readerName', $readerName);
+
 		$this->display('overdriveAspenData.tpl', 'OverDrive Aspen Data');
```
mdnoble73, line 48: Looks like this needs to be modified so the title changes as well

code/web/services/OverDrive/Dashboard.php
```diff
@@ -78,7 +81,9 @@ function launch() {
 		$interface->assign('loansAllTime', $loansAllTime);
 		$interface->assign('holdsAllTime', $holdsAllTime);
 
-		$this->display('dashboard.tpl', 'OverDrive Dashboard');
+		$interface->assign('title', $readerName . ' Dashboard');
+
+		$this->display('dashboard.tpl', $readerName . ' Dashboard');
```
mdnoble73, line 86: Trying to decide if we should translate this and use a placeholder for the title.   I.e. %1% Dashboard.  Then when we call display we make sure not to translate the title.  That will allow another language that translates it as Dashboard Libby to work.

mdnoble73 (approved):

PR #1527, opened 2023-12-01
Update Loading.js

b-lawlor:
added dark theme colors in loading text and progress bar


b-lawlor:
I wasn't able to test this locally, but tested in the nativebase docs https://docs.nativebase.io/progress

mdnoble73:
This creates accessibility issues because the text is the same color as the background.  I'll close for now, but plese discuss with us in slack if we're miss-understanding the change.

PR #1528, opened 2023-12-01
Update loadError.js

b-lawlor:
fixed typo in color name


b-lawlor:
although coolCray does sound like a fun color

PR #1534, opened 2023-12-05
OverDrive -> Custom Reader Name

K-Alette:
Updates java files for indexing custom reader names


code/reindexer/src/com/turning_leaf_technologies/reindexer/OverDriveProcessor.java
```diff
@@ -377,12 +377,7 @@ void processRecord(AbstractGroupedWorkSolr groupedWork, String identifier, BaseI
 						int numHolds = 0;
 						//Just create one item for each with a list of sub formats.
 						ItemInfo itemInfo = new ItemInfo();
-						itemInfo.seteContentSource("OverDrive");
```
mdnoble73: I think this change might not work if we have a consortium where one library has the name set to Libby and the other set to Sora.  I think whichever is set last will be set?

mdnoble73 (changes requested):

PR #1542, opened 2023-12-07
ensure that Square is passed a string for idempotency_key

gmcharlt:
Depending on PHP and MySQL driver versions, PHP is increasingly more likely to return numeric columns as PHP integers or floats.

For example, PHP 8.1+ will now do this for emulated prepared statements.

Since Square requires that the idempotency_key be a JSON string, but Aspen uses the numeric user_payments ID for the purpose, this patch ensures that it gets cast to a string.

Concretely, this patch fixes a bug where payments can fail to be created in Square with the following error (as found in Square's API logs):

```
{
  "errors": [
    {
      "code": "EXPECTED_STRING",
      "detail": "Expected a string value (line 1, character 20)",
      "field": "idempotency_key",
      "category": "INVALID_REQUEST_ERROR"
    }
  ]
}
```

### DIFF
--- a/code/web/services/MyAccount/AJAX.php
+++ b/code/web/services/MyAccount/AJAX.php
@@ -4859,7 +4859,7 @@ class MyAccount_AJAX extends JSON_Action {
 				$paymentId = $payment->id;
 				$paymentAmount = $payment->totalPaid;
 				$body = [
-					'idempotency_key' => $paymentId,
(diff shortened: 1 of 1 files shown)


gmcharlt:
This has been tested in production.

mdnoble73 (approved):

PR #1564, opened 2023-12-20
Fallback to User Phone When Placing Holds in Evergreen

HitScan:
If a user has phone notifications enabled but no default phone notification number set, fallback to the day / evening / other phone on their account. This is how the Evergreen OPAC works; the default notification phone number will be used if present, but is not required if another phone number is available on the account.


mdnoble73:
Changes look good.  Can you update the readme file. 

Thanks

mdnoble73:
Can you move release notes to the 24.01.00.MD file since that is the release these changes are targeted to?

HitScan:
Oh, good point. Coming soon

HitScan:
Moved. I didn't add a // jboyer - Equinox or anything like that since I'm not entirely clear how that's used or if needed in general. Can adjust if need be.

mdnoble73:
If you could add the //jboyer - Equinox section that would be great.  It helps to avoid merge conflicts as everyone contributes.  Then I remove them all before release.

mdnoble73 (approved):

PR #1589, opened 2024-01-04
- A new group of Field Types (with prefilling properties) have been added when users creates custom forms.

lucasmontoya13:


mdnoble73:
Could you update the release notes file and then I can merge this in.  I wonder if we also want a textarea field type with all of the values formatted together?

mdnoble73:
Sorry one more little change.  Can you put the release notes in 24.02.00.MD to match the version this code is going into? 

Thanks!

mdnoble73 (approved):

PR #1608, opened 2024-01-15
User's address information can be added to the donations

lucasmontoya13:
- Added a checkbox to request users to add their address (street, city, state and zip code) when making a donation.


mdnoble73:
Could you update so these fields will prefill if the user is logged in similar to what you did for the last pull request?  I think we also need the address2 field (used for apartment number etc). Thanks!

mdnoble73 (approved):

mdnoble73 (approved):

PR #1618, opened 2024-01-18
Get Barcode and Call Numbers of checkouts in Evergreen

Dyrcona:
Add a getCallNumberForCopy() method to the Evergreen driver to retrieve the call number label for a mapped copy.

Refactor the Evergreen driver's getCheckouts() method to use the open-ils.circ.actor.user.checked_out API, to build the checkout object itself from the returned values, and to use the getCallNumberForCopy() method to get the copy's call number for display.

The loadCheckoutData() method remains untouched because it is currently used elsewhere.

This is intended to fix the "issue" of copy barcodes and call numbers not showing up in Aspen when the ILS is Evergreen.

### DIFF
--- a/code/web/Drivers/Evergreen.php
+++ b/code/web/Drivers/Evergreen.php
@@ -37,13 +37,13 @@ class Evergreen extends AbstractIlsDriver {
 
 		$authToken = $this->getAPIAuthToken($patron, true);
 		if ($authToken != null) {
-			//Get a list of holds
+			//Get a list of circulations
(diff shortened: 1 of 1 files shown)


mdnoble73 (approved):

mdnoble73:
I made a few minor updates during testing.  It looks great now and I merged.  We'll figure out when to deploy to your server or if we should backport to an earlier release to deploy to your server.

Dyrcona:
Yes, thanks for the fixes. I missed those. I can't exactly copy and paste from my test bed, and I don't have a place to test the code myself.

PR #1627, opened 2024-01-23
API changes

catsoup11789:


code/web/sys/LibraryLocation/Location.php
```diff
@@ -120,7 +120,7 @@ class Location extends DataObject {
 		$defaultToCombinedResults;
 	public $useLibraryCombinedResultsSettings;
 
-	private $_hours;
+	public $_hours;
```
mdnoble73: We can make this protected, but not public.  If it's public it won't clear automatically.

mdnoble73 (changes requested):

code/web/sys/LibraryLocation/Location.php
catsoup11789: when it was private, it returned an error about it being private when trying to forcefully clear it...

mdnoble73 (approved):

PR #1665, opened 2024-02-08
Support a Thaw Date When Freezing Evergreen Holds

HitScan:
Evergeen supports freezing holds indefinitely or until a specified thaw date; the driver just needs a small adjustment to support this.


mdnoble73:
Can you please add this to 24.03 rather than 24.02? 

Thanks,

- Mark

HitScan:
Well that force-pushed update did not work out as planned. I'll resubmit in a new PR.

PR #1671, opened 2024-02-09
Respect Proxy Protocol

HitScan:
If Aspen is behind a TLS-terminating proxy we don't want to use http:// when building links, even though $_SERVER['HTTPS'] will be false. The X-Forwarded-Proto header is enabled by default for AWS and Google Cloud load balancers, and is easily enabled if using nginx directly.


mdnoble73:
This will need release notes

HitScan:
Updated

mdnoble73 (approved):

PR #1695, opened 2024-02-28
Copyable Self Reg Forms (the remix) and Updating Profile info in All …

K-Alette:
…Caps for Symphony

Copyable Self Reg Forms (the remix)
Updating Profile info in All Caps for Symphony should now work (Ticket 126162) 
Updated release notes

### DIFF
--- a/code/web/sys/SelfRegistrationForms/SelfRegistrationForm.php
+++ b/code/web/sys/SelfRegistrationForms/SelfRegistrationForm.php
@@ -241,4 +241,14 @@ class SelfRegistrationForm extends DataObject {
 			unset($this->_libraries);
 		}
 	}
+
+	public function loadCopyableSubObjects() {
+		$this->getFields();
+		$index = -1;
+		foreach ($this->_fields as $subObject) {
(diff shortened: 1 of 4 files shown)


code/web/sys/SelfRegistrationForms/SelfRegistrationForm.php
```diff
@@ -241,4 +241,14 @@ public function saveLibraries() {
 			unset($this->_libraries);
 		}
 	}
+
+	public function loadCopyableSubObjects() {
+		$this->getFields();
+		$index = -1;
+		foreach ($this->_fields as $subObject) {
+			$subObject->id = $index;
+			$subObject->browseCategoryGroupId = $this->id;
```
mdnoble73: In this case the subObject will be a SelfRegistrationFormValues so we will be setting the selfRegistrationFormId rather than the browseCategoryGroupId

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #1702, opened 2024-03-05
avoid putting authors into list of translatable terms

gmcharlt:
/Author/Home puts the (normalized) author name in the breadcrumbs, reasonably enough. However, the author name was getting added to the list of translatable terms, pooching it out with a bunch of terms that normally should be displayed as is. This patch marks normalized author breadcrumbs as not to be translated.

### DIFF
--- a/code/web/ResultsAction.php
+++ b/code/web/ResultsAction.php
@@ -2,10 +2,10 @@
 
 
 abstract class ResultsAction extends Action {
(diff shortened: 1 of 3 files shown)


mdnoble73:
Can you please add release notes for this.

Thanks,

- Mark

gmcharlt:
I've added a commit to update the release notes.

mdnoble73 (approved):

PR #1733, opened 2024-03-26
Fs to db 24 04

lucasmontoya13:


code/web/sys/DBMaintenance/version_updates/24.04.00.php
```diff
@@ -28,4 +64,23 @@ function getUpdates24_04_00(): array {
...
+	require_once ROOT_DIR . 'sys/File/FileUpload.php';
+	$uploadedFile = new FileUpload();
+	$uploadedFile->find();
+	$numUpdates = 0;
+	while ($uploadedFile->fetch()){
+		if (isset($uploadedFile->fullPath) && !isset($uploadedFile->uploadedFileData)){
+			$dataFromPath = file_get_contents($uploadedFile->fullPath);
+			$uploadedFile->uploadedFileData = $dataFromPath;
+			unset($uploadedFile->fullPath);
+			$numUpdates+= $uploadedFile->insert();
```
mdnoble73: Assuming the file is uploaded, we can probably delete the original file?

code/web/sys/DBMaintenance/version_updates/24.04.00.php
```diff
@@ -28,4 +64,23 @@ function getUpdates24_04_00(): array {
...
+function migrateUploadedFiles(&$update) {
+	require_once ROOT_DIR . 'sys/File/FileUpload.php';
+	$uploadedFile = new FileUpload();
+	$uploadedFile->find();
+	$numUpdates = 0;
+	while ($uploadedFile->fetch()){
+		if (isset($uploadedFile->fullPath) && !isset($uploadedFile->uploadedFileData)){
+			$dataFromPath = file_get_contents($uploadedFile->fullPath);
+			$uploadedFile->uploadedFileData = $dataFromPath;
+			unset($uploadedFile->fullPath);
```
mdnoble73: When I test this, I get an error if fullPath is unset.  Let's try setting to a blank string during migration

code/web/sys/DBMaintenance/version_updates/24.04.00.php
```diff
@@ -19,6 +19,42 @@ function getUpdates24_04_00(): array {
 		//kodi - ByWater
 
 		//lucas - Theke
+			'store_files_into_database' => [
+				'title' => 'Store files into database',
+				'description' => 'New column for uploaded files as blobs',
+				'continueOnError' => false,
+				'sql' => [
+					"ALTER TABLE file_uploads ADD COLUMN IF NOT EXISTS uploadedFileData LONGBLOB DEFAULT NULL",
```
mdnoble73: My version of MySQL isn't liking the IF NOT EXISTS statement here, let's leave that out and Database Maintenance will run ok even with errors.

code/web/sys/DBMaintenance/version_updates/24.04.00.php
```diff
@@ -19,6 +19,42 @@ function getUpdates24_04_00(): array {
...
+				]
+			],
+
+			'store_image_into_database' => [
+				'title' => 'Store images into database',
+				'description' => 'New column for uploaded images as blobs',
+				'continueOnError' => false,
+				'sql' => [
+					"ALTER TABLE image_uploads ADD COLUMN IF NOT EXISTS fullSizeImageData LONGBLOB DEFAULT NULL",
```
mdnoble73: My version of MySQL isn't liking the IF NOT EXISTS statement here, let's leave that out and Database Maintenance will run ok even with errors.

code/web/sys/DBMaintenance/version_updates/24.04.00.php
```diff
@@ -19,6 +19,42 @@ function getUpdates24_04_00(): array {
...
+				]
+			],
+
+			'store_custom_fonts_into_database' => [
+				'title' => 'Store custom fonts into database',
+				'description' => 'New columns for uploaded custom fonts as blobs',
+				'continueOnError' => false,
+				'sql' => [
+					"ALTER TABLE themes ADD COLUMN IF NOT EXISTS customHeadingFontData LONGBLOB DEFAULT NULL",
```
mdnoble73: My version of MySQL isn't liking the IF NOT EXISTS statement here, let's leave that out and Database Maintenance will run ok even with errors.

code/web/sys/DBMaintenance/version_updates/24.04.00.php
mdnoble73: This will need to do an update rather than an insert

code/web/sys/DBMaintenance/version_updates/24.04.00.php
```diff
@@ -28,4 +64,23 @@ function getUpdates24_04_00(): array {
 
 
 	];
+
+
+}
+
+function migrateUploadedFiles(&$update) {
```
mdnoble73: We'll want to add similar functions to migrate uploaded images as well as the fonts

code/web/sys/File/FileUpload.php
```diff
@@ -37,7 +39,7 @@ static function getObjectStructure($context = ''): array {
 			],
 			'fullPath' => [
```
mdnoble73, line 41: Do we want to change the name of this property from fullPath to uploadedFileData and then store the data back to the database?  That way we will have the data even after moving the database between servers.  I think we could do it either way (store thumbnail on disk or store thumbnail in DB).  We just would to make sure we regenerate the thumbnail if it doesn't exist when it is requested.

code/web/sys/File/FileUpload.php
```diff
@@ -5,6 +5,7 @@ class FileUpload extends DataObject {
 	public $id;
 	public $title;
 	public $fullPath;
+	public $uploadedFileData;
 	public $thumbFullPath;
```
mdnoble73, line 9: Similar to the uploadedFileData we will want to create another property thumbnailFileData

mdnoble73 (changes requested):
This is looking great! 

Let's get the Uploaded Files done all the way through and then we can tackle the other areas.  

To finish Uploaded Files, we will need to handle the areas where the files are shown.  It looks like those locations are: 
- Files/Contents
- Files/ViewPDF
- Files/WatchVideo
- GroupedWork/DownloadPDF
- GroupedWork/DownloadSupplementalFile
- Record/DownloadPDF
- Record/DownloadSupplementalFile
- WebBuilder/DownloadPDF
- WebBuilder/ViewThumbnail
-
We'll also want to handle the uploads in 
- Record/AJAX

PR #1746, opened 2024-04-04
Fix custom form redirection

lucasmontoya13:


mdnoble73 (changes requested):
Seeing some notices when viewing the form logged out: 
image
Other than those notices, everything seems to be working great!

mdnoble73 (dismissed):

mdnoble73 (changes requested):
Doing more testing, it looks like you will need to add checking for the form within index.php, there are examples of loading the urlAlias from lines 530-555 for Basic Pages and Custom Pages.

mdnoble73 (approved):

PR #1756, opened 2024-04-10
The group of buttons More info/Access Online has extended its width

lucasmontoya13:


mdnoble73 (changes requested):
When showing "Show Covers for Editions" is turned off in grouped work display options, the buttons get pushed to the next line as shown below.  We should also check "Show Related Record Labels" in grouped work display options to be sure everything looks good with that on and off. 
image

PR #1770, opened 2024-04-21
24.04.00 circulation reports

NashvillePublicLibrary:


mdnoble73 (changes requested):
Please add release notes

code/web/sys/Account/User.php
```diff
@@ -3552,6 +3552,10 @@ public function getAdminActions() {
 			'View Location Collection Reports',
 			'View All Collection Reports',
 		]);
+        $sections['circulation_reports']->addAction(new AdminAction('Weeding Report', 'View a collection weeding report for all items for a branch.', '/Report/WeedingReport'), [
```
mdnoble73: Let's add some sort of note that this will only work for Nashville and/or check to see if Nashville is active before adding these permissions that will just help support questions.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #1781, opened 2024-04-26
Reconcile GPL version number between license.txt and readme.txt

lizrea:
license.txt cites the GPL v3 license, this file references GPL v2, they should match up.


mdnoble73:
This should actually be GPL v2 since that is what VuFind was released under.  Thanks for catching this!

mdnoble73 (changes requested):
Should be GPL v2

mdnoble73:
Updated license.txt to match instead.

PR #1795, opened 2024-05-03
Proposed changes to httpd conf template for bad bot blocking via u/a

lizrea:
Proposal only - no need to merge (unless you like it...)

This patch is a proposal for a potential syntax to block known bad bots. 

The first line includes a file (that we'll need aspen to install) that can contain site specific, emergent bots to detect and send to 403.

The second include is a file that we would maintain to block known bad bots that bother every Aspen for central distribution to All aspens, no matter if they are hosted by us or not.


mdnoble73:
Set this up in later pull request

PR #1813, opened 2024-05-21
Summon new features 24.06.00

AlexanderBlanchardAC:


code/web/interface/themes/responsive/images/summon.png
AlexanderBlanchardAC, line 1: This image is used for the Summon options of the Explore More Bar.

code/web/interface/themes/responsive/Search/explore-more-bar.tpl
AlexanderBlanchardAC, line 1: The partner asked for the ability to remove the 'explore more' bar from the results pages. I could customise it more by giving the option to display it or hide it for different types of searches? e.g. Summon, library catalogue etc?

AlexanderBlanchardAC:
I removed the zone:identifier file. Windows generated it with Windows specific metadata, but it seems to work without it.

code/web/interface/themes/responsive/Search/explore-more-bar.tpl
mdnoble73: Yeah I think we'd want to do it based on the type of search and also be able to control what options show within Explore more as well as the order of the options showing in explore more.  Do we want to split that out into another PR?

PR #1814, opened 2024-05-22
GDPR  cleanup script

AlexanderBlanchardAC:
This commit alters the deletion of unsaved searches to be carried out each time the cron script runs. It also adds the option for libraries to choose to have stored information about the last list a user has accessed deleted after 14 days.

Test Plan:

1.Apply the patch.
2.Navigate to Primary Configuration > Library Systems.
3.Select your library and click to edit.
4.Use the search bar to find 'Delete last used list after 14 days' or locate this new option under the 'Searching' section.
5.The box should be unchecked by default as this is an opt in option.
When checked, when the cron runs, information from the users table relating to their last list used will be deleted if it has not been updated for at least 14 days.


code/cron/src/com/turning_leaf_technologies/cron/DatabaseCleanup.java
```diff
@@ -447,4 +448,56 @@ private void removeOldSearches(Connection dbConn, Logger logger, CronProcessLogE
 		}
 	}
 
+	private void removeOldLastListUsed(Connection dbConn, Logger logger, CronProcessLogEntry processLog) {
+		//Remove old last list used
+		try {
+			//Get list of libraries that want last used list cleared
+			PreparedStatement librariesListStmt = dbConn.prepareStatement("SELECT libraryId, deleteOldLastListUsedEntries from library where deleteOldLastListUsedEntries > 0");
+			PreparedStatement libraryLocationsStmt = dbConn.prepareStatement("SELECT locationId from location where libraryId = ?");
+			PreparedStatement deleteLastListUsedStmt = dbConn.prepareStatement("DELETE from user where lastListUsed = ?");
```
mdnoble73: This completely deletes the user rather than clearing lastListUsed, we'll need to preserve the user.

mdnoble73 (changes requested):

code/cron/src/com/turning_leaf_technologies/cron/DatabaseCleanup.java
```diff
@@ -447,4 +448,57 @@ private void removeOldSearches(Connection dbConn, Logger logger, CronProcessLogE
 		}
 	}
 
+	private void removeOldLastListUsed(Connection dbConn, Logger logger, CronProcessLogEntry processLog) {
+		//Remove old last list used
+		try {
+			//Get list of libraries that want last used list cleared
+			PreparedStatement librariesListStmt = dbConn.prepareStatement("SELECT libraryId FROM library WHERE deleteLastListUsedEntries = 1");
+			PreparedStatement libraryLocationsStmt = dbConn.prepareStatement("SELECT locationId FROM location where libraryId = ?");
+			PreparedStatement deleteLastListUsedStmt = dbConn.prepareStatement("UPDATE user SET lastListUsed = NULL WHERE lastListUsed = ?");
```
mdnoble73: We should reset this to the default which is -1

mdnoble73 (changes requested):

code/cron/src/com/turning_leaf_technologies/cron/DatabaseCleanup.java
```diff
@@ -447,4 +448,57 @@ private void removeOldSearches(Connection dbConn, Logger logger, CronProcessLogE
...
+						libraryLocations.append(", ");
+					}
+					libraryLocations.append(libraryLocationsRS.getString("locationId"));
+				}
+				if (libraryLocations.length() > 0) {
+					long now = new Date().getTime() /1000;
+					long daysToPreserve = 14;
+					long earliestDateToPreserve = now - (daysToPreserve * 24 * 60 * 60);
+
+					PreparedStatement lastListUsedEntriesToDeleteStmt = dbConn.prepareStatement("SELECT lastListUsed FROM user WHERE user.homeLocationId IN (" + libraryLocations + ") AND lastListused < ?");
```
mdnoble73, line 481: the lastListused isn't a date so this query won't work.  We don't actually track the date the last list was added to in the user table so we'd either need to query the user's lists or store extra data just to clear this.  I don't really think this is a privacy issue for patrons and there are other ways to find the same data by querying the list entries.

mdnoble73 (changes requested):

PR #1817, opened 2024-05-23
Target Audience & Manga vs Manga Graphic Novels 

K-Alette:
Fix issue where target audiences were not being mapped using the correct translation maps
Add match for graphic novels where 655a equals "manga" we will add "Manga" as a format
Updated release notes


code/java_shared_libraries/src/com/turning_leaf_technologies/indexing/BaseIndexingSettings.java
```diff
@@ -210,9 +210,6 @@ public String translateValue(String mapName, String value) {
 	}
 
 	public boolean hasTranslation(String mapName, String value) {
-		if (value == null) {
```
mdnoble73: I think we need to keep this in here.  I was getting null pointer errors on a few servers.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #1822, opened 2024-05-30
New e content

AlexanderBlanchardAC:
This commit adds eResource to the e-content list and enables the display of an access online button for eResources.


mdnoble73:
I'm not actually seeing the AccessOnline link being generated anywhere.  Where is this triggered from?

PR #1837, opened 2024-06-07
If catalog is offline, don't show buttons that prompt a login

Abhyastamita:
In offline mode, eContent logins are not enabled, buttons won't show up if they would prompt the user to login, since logins won't work when the catalog is offline.  The buttons that won't show are:
- Add to List
- Add a Review
- Add to your Events
- Save Search
- Email this Search
- Materials Request
- Check out buttons for Palace Project, Hoopla, Boundless, Cloudlibrary, Libby/Overdrive

Additional behavior:
- For Libby/Overdrive only, instead of a checkout button, there should be an Access Online button instead that goes to the Overdrive page (I'm not sure if we want it to work this way or not?)
- Events Registration button is not hidden - instead it no longer asks for a login since there wasn't any reason to require a login for this.

### DIFF
--- a/code/web/RecordDrivers/Axis360RecordDriver.php
+++ b/code/web/RecordDrivers/Axis360RecordDriver.php
@@ -201,7 +201,10 @@ class Axis360RecordDriver extends GroupedWorkSubDriver {
 				$loadDefaultActions = count($this->_actions) == 0;
 			}
 
-			if ($loadDefaultActions) {
+            //Check if catalog is offline and login for eResources should be allowed for offline
+            global $offlineMode;
+            global $loginAllowedWhileOffline;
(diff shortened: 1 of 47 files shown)


Abhyastamita:
There are some changes to indentation that I didn't intend that look right in PHPStorm and are wrong in Github and I'm not sure how to fix them.  Particularly [code/web/RecordDrivers/EbscohostRecordDriver.php](https://github.com/Aspen-Discovery/aspen-discovery/pull/1837/commits/0dab6ecaa8a97b60d340adddb09e0823e0a3f3ac#diff-ea2abdb20c7aa56d98e338c9fcf14f906feceac6e800066e0c6e961fc0beca08) was not actually supposed to have any changes at all.

mdnoble73 (approved):

PR #1850, opened 2024-06-12
SMTP Setting

ammopt:

PR #1851, opened 2024-06-12
Preferred Mail Sender

ammopt:
The relevant 4 commits are part of this PR:
https://github.com/PTFS-Europe/aspen-discovery/pull/75/commits

This depends on (built on top of) https://github.com/Aspen-Discovery/aspen-discovery/pull/1850


ammopt:
This will require a rebase on https://github.com/Aspen-Discovery/aspen-discovery/pull/1850 but putting this on hold until that moves further.

mdnoble73:
Can you rebase this when you get a chance?  

Also, I think rather than providing an order to use all of the possible settings we'll want to assign specific settings to libraries.  That will be more consistent with the existing patterns in Aspen and will allow 2 libraries within a consortium to use 2 different settings.  So library A could use an SMTP server, library B could use Amazon SES, and library C could use Send Grid, and Library D can use a different Send Grid account.  

If we do want to allow fallback servers in the event of an outage, we have patterns for attaching those in order as well.  Let me know if you want an example.

ammopt:
Hi Mark, thank you for your time looking here. I'm only able to reply now as I was on leave.
Previous to [this commit](https://github.com/Aspen-Discovery/aspen-discovery/commit/1db0b2e1fdde39c7b50170a7d9b3f06449dd684f) in the SMTP dev piece, the code was hardcoded to globally consider Amazon first and then SendGrid, which means that if both were configured, Amazon would always take precedence over SendGrid, regardless of library.

When I implemented the SMTP mail sender, I had to make a decision and opted to make SMTP first in priority in that hardcoded logic, but it all seemed very arbitrary to me, so I consulted some of my colleagues and we agreed it'd make sense to have this priority order be configurable. The goal of this patchset was/is simply to turn what is global hardcoded logic into a global customizable configuration (through a system variable).

Your suggestion seems to be a bit of wider scope than what I originally intended with this patchset. My understanding is that you'd prefer we have a "Preferred mail sender" option at each library level individually (in 'Library Systems') so that each can have the option to choose their own preferred sender method.

Please let me know if what I say above makes sense. I'm happy to have this patchset dropped if it doesn't make sense. I'm happy to follow-up with your suggestion as a replacement of this patchset, and I'm also happy to follow-up with your suggestion as a continuation to this patchset.

mdnoble73:
No worries, I was also out.  

If the goal is not to support multiple email services being setup for a single instance I think that all instances will only have a single email service setup so the order shouldn't really matter.  If we do have multiple email services setup for a single instance, they would need to be assigned to different libraries. 

Are you intending to use multiple providers for a single instance?  If not, the current approved code should be fine and we could drop this PR.

ammopt:
Hi Mark, understood. I agree that the PR can be dropped. Thank you.

PR #1855, opened 2024-06-14
Renewals correct branch

tomascohen:
Koha's API for renewability will pick the logged in user home branch for calculating renewability. So depending on the home branch for the logged in user, the results it will return.

In order to properly replicate OPAC behavior, Aspen should override the 'current' branch, by passing the `x-koha-library` header, which was designed for this specific purpose. And it should override using the patron's home branch.

That's what this patch does:

* Add a new parameter to the `checkAllowRenewals` method so it can be passed the library id.
* Use the new parameter to build the `x-koha-library` header.
* Retrieve the home library ID from the DB, the same way Aspen already does for other bits, and pass it to `checkAllowRenewals`.


tomascohen:
Not sure why the PR got generated against 24.06.00!

PR #1862, opened 2024-06-18
Fix tab order for Browse Category keyboard navigation

Abhyastamita:
This relates to ticket 133620.  Tabbing through accessible browse categories on the homepage went to the next book off the screen instead of to the Next button, which caused Next to move.  This change adds tabindex="-1" to the hidden books that are off the screen so that they won't get focus.  It also moves the Previous button so that the tab order is now Previous -> Book Cover 1-5 -> Next instead of Book Cover 1-5 -> Previous -> Next.  I have tested in Firefox, Edge, and Chrome.


Abhyastamita:
Just realized this was supposed to be 24.06, not 24.07

PR #1869, opened 2024-06-18
Add new apache's rules

lucasmontoya13:
- New rules have been added to the httpd configuration file to allow Aspen to switch to other languages even when users fill out forms.


sites/default/httpd-default.conf
```diff
@@ -2,59 +2,61 @@
...
+		# Note: The following RewriteRule directives include the [B] flag to escape
+		# backreferences.  This prevents encoding problems caused by special characters
+		# like & if they show up in ids.  Note that the flag doesn't work in some
+		# versions of Apache prior to 2.2.12; if you run into trouble, try upgrading.
+		RewriteRule   ^(MyAccount)/([^/]+)/(.+)$   index.php?module=$1&action=$2&id=$3      [B,L,QSA]
+		RewriteRule   ^(Record)/([^/]+)/(.+)$       index.php?module=$1&id=$2&action=$3     [B,L,QSA]
+		RewriteRule   ^(Record)/(.+)$               index.php?module=$1&id=$2               [B,L,QSA]
+		RewriteRule   ^$                            index.php?module=Search&action=Home     [B,L,QSA]
+		RewriteRule   ^([^/]+)$                     index.php?module=$1&action=$2           [B,L,QSA]
+		RewriteRule   ^([^/]+)/(.+)$                index.php?module=$1&action=Home         [B,L,QSA]
```
mdnoble73: I believe these rules got reversed:
RewriteRule   ^([^/]+)$                     index.php?module=$1&action=$2           [B,L,QSA]
does not have a second capturing group so it should default to Home. 
RewriteRule   ^([^/]+)/(.+)$                index.php?module=$1&action=Home         [B,L,QSA]
does have to capturing groups so the second should be passed to index.php

mdnoble73: This will need to get changed in all the files

mdnoble73 (changes requested):
Looks like a few of the RewriteRules need to be adjusted still.  Is it possible to achieve the same thing by adjusting the logic in loadModuleActionId within index.php?

lucasmontoya13:
Sorry, i haven't seen the first comment! I'll try to fix it in index.php

mdnoble73:
Do you want to keep this PR open or do you think you can do it in index.php? I'm fine either way.

PR #1882, opened 2024-06-28
Issue 117950: Enter key on input fields activates Translation Mode

lucasmontoya13:
In the Translation section, when the user tries to use any filter ,using the enter key, activates Translation Mode instead of activate 'Update Filters' button. This patch fixes that error.


mdnoble73 (changes requested):
Can you add release notes for this?

code/web/sys/Translation/Translator.php
```diff
@@ -338,11 +338,11 @@ private function loadTranslationsFromIniFile() {
 
 	public function translationModeActive() {
 		if ($this->translationModeActive === null) {
-			if (isset($_REQUEST['startTranslationMode'])) {
+			if (isset($_REQUEST['startTranslationMode']) && $_REQUEST['startTranslationMode'] == 'true' ) {
 				@session_start();
 				$_SESSION['translationMode'] = 'on';
 				$translationModeActive = true;
-			} elseif (isset($_REQUEST['stopTranslationMode'])) {
+			} elseif (isset($_REQUEST['stopTranslationMode']) && $_REQUEST['stopTranslationMode'] = 'true' ) {
```
mdnoble73: This needs to be == rather than =

mdnoble73 (changes requested):
One little change.  Can your resubmit for 24.08?

mdnoble73:
This was manually merged

PR #1885, opened 2024-07-01
Message "old PIN provided is incorrect" should be translatable

Abhyastamita:
- Made message translatable.
- Updated release notes.
- This is for ticket 123151, which has been closed already.  I couldn't actually get this error message to show up and it looks to me like it may only show for Polaris or possibly not at all, since the ticket says the problem is not happening anymore.

### DIFF
--- a/code/web/interface/themes/responsive/js/aspen/overdrive.js
+++ b/code/web/interface/themes/responsive/js/aspen/overdrive.js
@@ -294,7 +294,9 @@ AspenDiscovery.OverDrive = (function(){
 					url: ajaxUrl,
 					cache: false,
 					success: function(data){
-						AspenDiscovery.showMessage("Title Returned", data.message, data.success);
+						AspenDiscovery.showMessage(
+							data.success ? 'Title Returned' : 'Unable to Return Title', data.message, data.success
+						);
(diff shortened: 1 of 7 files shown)


Abhyastamita:
This PR now also includes a QA fix for the Kindle error message problem.  The header said "Title Returned" even when there was an error and the title was not actually returned.  Now it displays the appropriate header if the request doesn't succeed.

Abhyastamita:
And a QA fix to put the return to list button for forms back again, since somehow that change got reversed again.

code/web/interface/themes/responsive/js/aspen/overdrive.js
```diff
@@ -294,7 +294,9 @@ AspenDiscovery.OverDrive = (function(){
 					url: ajaxUrl,
 					cache: false,
 					success: function(data){
-						AspenDiscovery.showMessage("Title Returned", data.message, data.success);
+						AspenDiscovery.showMessage(
```
mdnoble73, line 297: Long term, we should probably return the title from the AJAX call so it can be translated.

mdnoble73 (approved):

PR #1901, opened 2024-07-12
Show summon search tools

AlexanderBlanchardAC:
Test Plan:

1. Before applying patch, set up Summon and carry out a search.
2. Notice that there is no checkbox to hide covers or option to switch between list and covers view at the top of the page.
3. Repeat the above with an Ebsco EDS search.
4. Apply the patch.
5. Notice that when you carry out a Summon Search now, the search tools are at the top of the screen as they are in Ebsco EDS and checking the hide covers box refreshes the page and returns results without the covers.


mdnoble73:
The hide covers action works great, but the toggle between covers mode and list mode does not work. You could just remove the covers and list mode toggles rather than implementing them since covers mode is unlikely to be used for this type of resource.  I did the same for EBSCO and EBSCOhost in 24.08 if you want an example of how I did that.

mdnoble73 (changes requested):
This will need release notes, also see main discussion thread

AlexanderBlanchardAC:
> The hide covers action works great, but the toggle between covers mode and list mode does not work. You could just remove the covers and list mode toggles rather than implementing them since covers mode is unlikely to be used for this type of resource. I did the same for EBSCO and EBSCOhost in 24.08 if you want an example of how I did that.

I have removed the covers / list option

AlexanderBlanchardAC:
> This will need release notes, also see main discussion thread

Sorry, MD file updated.

code/web/interface/themes/responsive/Search/search-toolbar.tpl
mdnoble73, line 1: Changing this file removes those options from all searches not just summon.

code/web/interface/themes/responsive/Search/results-displayMode-toggle.tpl
mdnoble73, line 1: Changing this file removes those options from all searches not just summon.

mdnoble73 (changes requested):
This change removes those tools from all searches which isn't what we want.

PR #1902, opened 2024-07-12
Fix summon facet results

AlexanderBlanchardAC:
Adjust code so that if not on page one of results when a filter is applied, the new results set begins on page one.
Fix bug: if current result numerical value is greater than the number of results available in a facet filter, the facet filter did not apply when clicked.

TEST PLAN:
Prior to applying patch:

Carry out a search in Summon.
Move to a results page that is greater than page 1.
Apply a facet.
Notice that when results load, they start form the page and result number you were on prior to selecting a filter.
APPLY PATCH

Repeat the above.
Notice that this time when the page reloads after selecting a facet filter, you are on page one of your new results set.


mdnoble73:
Can you resubmit against 24.08

mdnoble73:
Can you resubmit against 24.08 please?

PR #1906, opened 2024-07-15
Split create site

lucasmontoya13:


install/createSite.php
```diff
@@ -171,41 +172,41 @@
...
+	if (empty($variables['databaseHost'])){
+		$variables['databaseHost'] = "localhost";
 	}
 
-	$variables['aspenDBPort'] =  readline("Database host for Aspen (default: 3306) > ");
-	if (empty($variables['aspenDBPort'])){
-		$variables['aspenDBPort'] = "3306";
+	$variables['databasePort'] =  readline("Database host for Aspen (default: 3306) > ");
+	if (empty($variables['databasePort'])){
+		$variables['databasePortPort'] = "3306";
```
mdnoble73: This variable is defined as databasePortPort which is causing errors later on since other places expect just database port.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #1909, opened 2024-07-17
Where is it button renders consistently

Chloe070196:
Description:

Search results for which there are one or more copies available in one or more locations should all include a 'Where is it' button. With the current behaviour, the button only display for some of the search results which meet this criteria. This small patch resolves this by updating the check in copySummary.tpl to include one additional condition.

Test plan:

To replicate the bug:

- Run a search (keep the criteria field empty)
- Look through the search results: some results that have available copies at certain locations will display a 'Where is it?' button, whilst other won't

To test the fix:

- Apply the patch
- Run a search (keep the criteria field empty)
- Look through the search results: all results that have available copies at certain locations should now display a 'Where is it?' button

### DIFF
--- a/code/web/Drivers/Koha.php
+++ b/code/web/Drivers/Koha.php
@@ -3596,14 +3596,36 @@ class Koha extends AbstractIlsDriver {
 		];
 
 		$catalogUrl = $this->accountProfile->vendorOpacUrl;
+
+		$kohaVersion = $this->getKohaVersion();
+		$csrfToken = '';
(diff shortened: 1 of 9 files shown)


mdnoble73:
Could I get more information about the goal here? The definition of option 1 is "Show first 3 available copies  & Where Is It link only if there additional copies" and the definition of option 2 is "Show first 3 available copies & Where Is It link always".  It seems like this PR makes option 1 behave the same as option 2.

PR #1914, opened 2024-07-19
Cookie management updates

AlexanderBlanchardAC:


code/web/interface/themes/responsive/MyAccount/myPreferences.tpl
```diff
@@ -243,9 +226,9 @@
 							$("#usernameHelpButton").attr("aria-expanded","true");
 						}
 						return false;
-					})
+					}) -->
 					{/literal}
-				</script>
+				</script> *}
```
mdnoble73: This is not commented out properly so the trailing *} renders on the screen.

mdnoble73:
The cookie consent should be tied to which modules are active.  Only about 3% of libraries have Axis 360 or cloudLibrary active and EBSCO, EBSCOhost, Summon are similarly low numbers so having these active will be confusing for patrons since we would be asking them to consent to tracking services they don't have.  Tracking side loaded data will also also probably need to be broken up by sideload.  Alternatively, we may want to just have a single check box for tracking all of these.  A user that objects to analytics tracking for Hoopla would also most likely object to tracking analytics for OverDrive and vice versa so from a user perspective they would likely always be checking all of these or none of these.  

On the manage cookie preferences popup, the save preferences button should be placed next to the close button. 

On the manage cookie preferences screen we should align all toggles in a column.

mdnoble73 (changes requested):
See main conversation thread

PR #1920, opened 2024-07-23
Allow custom fields for bib-level call numbers

Abhyastamita:
- Added setting in the Indexing Profile, 'Bib Based Call Number Field', which defaults to the standard Dewey fields (099:092:082), but can be changed to LC fields instead (090:086:050) or whatever the library uses.  These are in order of preference, so only if 099 doesn't exist, use 092, etc.
- Updated database to add new setting field.
- The ILS Record Processor now uses the fields defined in the new setting.  If "use099forBibLevelCallNumbers()"  is set to false (just Arlington), it will skip the 099 even if it's listed.
- Updated release notes.
This is for ticket #133082.


Abhyastamita:
- Removed use099forBibLevelCallNumbers() and added a database update so that if the ILS Class is Arlington Koha, Bib Based Call Number Fields will be set to 092:082, since Arlington doesn't use 099.

mdnoble73 (approved):

PR #1930, opened 2024-07-25
Change of languages in custom forms

lucasmontoya13:
- Allow Aspen to change languages when users fill out forms.

### DIFF
--- a/code/web/index.php
+++ b/code/web/index.php
@@ -1167,7 +1167,13 @@ function loadModuleActionId() {
 			$requestPath = $requestURI;
 			if (strpos($requestPath, '?') > 0) {
 				$requestPath = substr($requestPath, 0, strpos($requestPath, '?'));
+
 			}
+
+			//Store query parameters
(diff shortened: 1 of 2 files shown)


code/web/index.php
```diff
@@ -1211,7 +1217,16 @@ function loadModuleActionId() {
 						$_REQUEST['module'] = 'WebBuilder';
 						$_REQUEST['action'] = 'Form';
 						$_REQUEST['id'] = $form->id;
+<<<<<<< Updated upstream
```
mdnoble73: Looks like this has merge artifacts in it.  Could you complete the merge and resubmit?

mdnoble73 (changes requested):

code/web/release_notes/24.07.00.MD
mdnoble73, line 1: Can you remove this file from your PR

mdnoble73 (changes requested):
Has a lot of changes to the 24.07 release notes which shouldn't be included

PR #1932, opened 2024-07-26
Sort volume numbers correctly

Abhyastamita:
- Changed sorting to natural sort so that volumes would sort Vol 1, Vol 2, Vol 10, Vol 11 instead of Vol 1, Vol 10, Vol 11, Vol 2.
- Updated release notes.


Abhyastamita:
Small unrelated change:
- Added missing parameters (library_id, eula, login_form) to cloudLibrary login to fix the bug with Open in cloudLibrary button not logging into cloudLibrary.

mdnoble73 (approved):

PR #1953, opened 2024-08-02
Change order of permission changes for debian

lizrea:
Logs directory was getting wrong ownership, moving the logs directory ownership command to after the permission changes for the parent should fix it

to test: 
run updateSitePermissions.sh <sitename>, observe that the permissions for var/log/aspen-discovery/<site>/logs are www-data:aspen_apache

apply this patch
run updateSitePermissions.sh <sitename>, observer that the permissions for /var/log/aspen-dsicovery/<site>/logs are now aspen:aspen

### DIFF
--- a/install/updateSitePermissions_debian.sh
+++ b/install/updateSitePermissions_debian.sh
@@ -55,8 +55,8 @@ chown -R www-data:aspen_apache /usr/local/aspen-discovery/tmp
 echo "Updating /var/log directory"
 chmod -R 755 /var/log/aspen-discovery/$1
 chmod -R 755 /var/log/aspen-discovery/$1/logs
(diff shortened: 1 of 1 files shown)


mdnoble73:
Can you add release notes see: https://dev.aspendiscovery.org/release-notes also do you want this in a version earlier than 24.09?  24.09 will go out in September and Debian upgrades will be on versions 24.07.02 or 24.08 primarily.

lizrea:
yep happy to do that

mdnoble73 (changes requested):
Please add release notes

mdnoble73 (approved):

mdnoble73:
I'll add release notes and update centos as well

PR #1954, opened 2024-08-05
feat: usage data download

Chloe070196:
This patch allows admins to download the content of the usage graphs found within the admin dashboard. If it is approved, the feature can then be added to the usage graphs of Summon and ILS, as well as those of any other dashboards to which usage graphs will be added in the future in further PRs.

Clicking the 'Export as CSV' button triggers a fetch of the most recent data for the usage graph from the database as well as its download as a CSV file. A warning message to the user indicates that the data in the file may be more recent than the on-screen data (and result in discrepancies), unless they choose to reload the page immediately after the download.

Test plan:

- once logged in as an Aspen admin, navigate to Aspen Administration > System Reports > Usage Dashboards
- navigate to the various usage graphs within, including section-wide graphs such as 'General Usage'
- notice the added 'Export as CSV' button, and the accompanying warning message
click 'Export as CSV'
- immediately refresh the page
- check that the file has been downloaded
- check that the values within the CSV match the ones in 'raw data table'


code/web/interface/themes/responsive/Admin/usage-graph.tpl
```diff
@@ -28,6 +28,10 @@
 				</tbody>
 			</table>
 		</div>
+		<div>
+			<a id="UsageGraphExport" class="btn btn-sm btn-default" href="/Admin/AJAX?method=exportUsageData&stat={$stat}&instance={$instance}">{translate text='Export To CSV' isAdminFacing=true}</a>
```
mdnoble73: This needs a check for if $instance is empty. Without that, a notice gets generated when the page renders which prevents the button from showing.

code/web/interface/themes/responsive/Admin/usage-graph.tpl
```diff
@@ -28,6 +28,10 @@
...
+			<div id="{$propName}HelpBlock" class="help-block" style="margin-top:0"><small class="text-warning"><i class="fas fa-exclamation-triangle"></i> {translate text="Exporting will retrieve the latest data. To see it on screen, refresh this page." isAdminFacing=true}</small></div>
```
mdnoble73: $propName is not defined so this needs to be switched to explicitly define the id for the help block.  Without that a notice is generated in debug mode which prevents the page from rendering properly.

mdnoble73 (changes requested):
Functionality looks great. Just a couple minor issues to avoid notices in debug mode.

code/web/interface/themes/responsive/Admin/usage-graph.tpl
Chloe070196: Could I ask for guidance on this? I am unsure of what the best way to handle this would be.

The `Admin_AJAX::exportUsageData()` method calls `Admin_UsageGraphs::buildCSV()` which was modelled on the `Admin_UsageGraphs::launch()` method either requesting $instance from the url, or setting $instanceName to ''.  This same control flow logic is implemented in the `Admin_AJAX::exportUsageData()`. 

I looked also at `usage_dashboard.tpl` for guidance (since it contains the links leading to the usage graphs): it seems to be using $selectedInstance for a similar purpose (and $selectedInstance usually appears to be empty), and appears to be doing so without any checks within the .tpl file.

Should $instance remain as an optional parameter in the AJAX CSV download url? Should it be simply removed from the url altogether, if it does turn out that it is always null? 

Implementing a check to see whether $instance is empty in `usage-graph.tpl` could perhaps be done with an 'even if it is empty, do display the button' sort of logic? As far as I understand it, this seems to be the intended behaviour.

Many thanks!

mdnoble73: You can wrap the {$instance} with a check for if it is empty so {if !empty($instance)}{$instance}{/if}

mdnoble73 (changes requested):
$propName still needs to be removed

mdnoble73 (approved):

PR #1963, opened 2024-08-06
Sierra Self Reg

K-Alette:
- Functionality for custom self registration forms and ILS-based self registration via Sierra APIs for Sierra libraries 
- Added TOS functionality for CarlX self registration via ILS 
- Added release notes & DB updates


code/web/sys/SelfRegistrationForms/SelfRegistrationTerms.php
mdnoble73, line 1: Is this no longer used? I think that either WYLD or SWAN really wanted to be able to redirect if Terms were not accepted.

mdnoble73 (approved):

PR #1971, opened 2024-08-09
Add pagination to Usage by User Agent Report

Abhyastamita:
- Added pagination to Usage by User Agent Report
- Updated release notes

### DIFF
--- a/code/web/services/Admin/UsageByUserAgent.php
+++ b/code/web/services/Admin/UsageByUserAgent.php
@@ -18,8 +18,16 @@ class Admin_UsageByUserAgent extends Admin_Dashboard {
 
 		$this->loadDates();
 
+		$page = isset($_REQUEST['page']) ? $_REQUEST['page'] : 1;
+		$pageSize = isset($_REQUEST['pageSize']) ? $_REQUEST['pageSize'] : 5; // to adjust number of items listed on a page
+		$interface->assign('recordsPerPage', $pageSize);
+		$interface->assign('page', $page);
+
 		$userAgent = new UserAgent();
 		$userAgent->orderBy('userAgent');
+		$total = $userAgent->count();
+		$userAgent->limit(($page - 1) * $pageSize, $pageSize);
+		$userAgent->find();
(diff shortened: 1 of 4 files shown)


code/web/services/Admin/UsageByUserAgent.php
```diff
@@ -18,8 +18,16 @@ function launch() {
 
 		$this->loadDates();
 
+		$page = isset($_REQUEST['page']) ? $_REQUEST['page'] : 1;
+		$pageSize = isset($_REQUEST['pageSize']) ? $_REQUEST['pageSize'] : 5; // to adjust number of items listed on a page
```
mdnoble73: Let's make the default page size 30 to match the selections in the menu.

code/web/services/Admin/UsageByUserAgent.php
```diff
@@ -18,8 +18,16 @@ function launch() {
...
+		$interface->assign('recordsPerPage', $pageSize);
+		$interface->assign('page', $page);
+
 		$userAgent = new UserAgent();
 		$userAgent->orderBy('userAgent');
+		$total = $userAgent->count();
+		$userAgent->limit(($page - 1) * $pageSize, $pageSize);
+		$userAgent->find();
 		$allUserAgents = $userAgent->fetchAll('id', 'userAgent');
```
mdnoble73: The fetchAll statement redoes the search from the previous statement.  This just needs to be switched to a loop through the results and store the values appropriately.

mdnoble73 (changes requested):

code/web/services/Admin/UsageByUserAgent.php
Abhyastamita: Thanks Mark!  This is fixed now.

code/web/services/Admin/UsageByUserAgent.php
Abhyastamita: Thanks!  I changed it to 30.

mdnoble73 (approved):

PR #1979, opened 2024-08-14
Added title to cover images

Howell-Carnegie:
Set title to the same text as the alt text so a tooltip shows on hover of the cover image. This was done for both browse category view and search results view of GroupedWork. This fixes  ticket 131697


mdnoble73 (changes requested):
Two quick changes if you don't mind

1. For the title if you change from escape to escapeCSS that will make sure that if there embedded quotes in the title they will be escaped properly. It looks like some of the alt text may need that as well?
2. Add release notes as described here: https://dev.aspendiscovery.org/release-notes

Howell-Carnegie (commented):
I updated the release notes as requested.

mdnoble73 (approved):

PR #1981, opened 2024-08-14
Use CHANGE instead of RENAME

K-Alette:
Fix error when using RENAME COLUMN by using CHANGE COLUMN instead


mdnoble73 (approved):
Syntax for this is not quite right.  It will need the original field definition repeated after the new name, and there is no TO between the old and new.

PR #1984, opened 2024-08-15
Properly close the connections to Koha from Aspen

kylemhall:
It is best practice to close all database connections after the data from a query has been ingested and parsed.

This change calls close() on all mysqli result sets that do not already call close after use.

No changes in Aspen behavior should be noted, but this should result in a reduction of used connections to the database.

### DIFF
--- a/code/web/Drivers/Koha.php
+++ b/code/web/Drivers/Koha.php
@@ -46,6 +46,7 @@ class Koha extends AbstractIlsDriver {
 			$curRow = $results->fetch_assoc();
 			$address = $curRow['address'];
 			$city = $curRow['city'];
+			$results->close();
 		}
 
 		$postVariables = [
@@ -142,6 +143,7 @@ class Koha extends AbstractIlsDriver {
 						$city = $curRow['city'];
 						$lastname = $curRow['surname'];
 					}
+					$results->close();
 				}
 
 				$postVariables = [
@@ -437,7 +439,6 @@ class Koha extends AbstractIlsDriver {
 			if ($renewPrefRow = $renewPrefResults->fetch_assoc()) {
 				$renewPref = $renewPrefRow['autorenew_checkouts'];
 			}
-
 			$renewPrefResults->close();
 		}
 		$timer->logTime("Loaded borrower preference for autorenew_checkouts");
@@ -455,7 +456,6 @@ class Koha extends AbstractIlsDriver {
 					$patronIsExpired = true;
 				}
 			}
-
 			$patronExpirationResults->close();
 		}
 		$timer->logTime("Loaded patron expiration date");
@@ -709,6 +709,7 @@ class Koha extends AbstractIlsDriver {
 
 			$checkouts[$curCheckout->source . $curCheckout->sourceId . $curCheckout->userId] = $curCheckout;
 		}
+		$results->close();
 
 		//Check to see if any checkouts are Claims Returned
 		$allIssueIdsAsString = implode(',', $allIssueIds);
@@ -897,9 +898,11 @@ class Koha extends AbstractIlsDriver {
 					$patronId = $lookupUserRow['borrowernumber'];
 					$newUser = $this->loadPatronInfoFromDB($patronId, null, $barcode);
 					if (!empty($newUser) && !($newUser instanceof AspenError)) {
+						$lookupUserResult->close();
 						return $newUser;
 					}
 				}
+				$lookupUserResult->close();
 			} else if ($this->getKohaVersion() >= 22.1110) {
 				//Authenticate the user using KOHA API
 				$oauthToken = $this->getOAuthToken();
@@ -944,9 +947,11 @@ class Koha extends AbstractIlsDriver {
 
 							$expiredPasswordResult = $this->processExpiredPassword($lookupUserRow['borrowernumber'], $barcode);
 							if ($expiredPasswordResult != null) {
+								$lookupUserResult->close();
 								return $expiredPasswordResult;
 							}
 						}
+						$lookupUserResult->close();
 					}
 					$result['messages'][] = translate([
 						'text' => 'Unable to authenticate with the ILS.  Please try again later or contact the library.',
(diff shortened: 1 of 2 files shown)


mdnoble73 (changes requested):
Please add release notes as well as a test plan

code/web/Drivers/Koha.php
```diff
@@ -737,8 +738,8 @@ public function getCheckouts(User $patron): array {
 						$logger->log("Error parsing claims returned info " . $claimsReturnedResult['created_on'] . " $e", Logger::LOG_ERROR);
 					}
 				}
-				$claimsReturnedResults->close();
 			}
+			$claimsReturnedResults->close();
```
mdnoble73: This will not work in cases where $claimsReturnedResults is false. This line should have been left where it was.

code/web/Drivers/Koha.php
```diff
@@ -900,6 +901,7 @@ public function patronLogin($username, $password, $validatedViaSSO) {
 						return $newUser;
```
mdnoble73, line 907: Before returning $newUser, the database connection should also be closed.

code/web/Drivers/Koha.php
```diff
@@ -947,6 +949,7 @@ public function patronLogin($username, $password, $validatedViaSSO) {
 								return $expiredPasswordResult;
```
mdnoble73, line 956: $lookupUserResult should also be closed here

code/web/Drivers/Koha.php
```diff
@@ -1036,6 +1039,7 @@ public function patronLogin($username, $password, $validatedViaSSO) {
 							}
 						}
 					}
+					$lookupUserResult->close();
```
mdnoble73: There are several returns in the block between where the SQL statement is opened and the close statement where the connection should be closed.

code/web/Drivers/Koha.php
```diff
@@ -1322,8 +1327,11 @@ private function loadPatronInfoFromDB($patronId, $password, $suppliedUsernameOrB
 
 			$timer->logTime("patron logged in successfully");
 
+			$lookupUserResult->close();
```
mdnoble73: We should also close if the $user->insert() statement fails

code/web/Drivers/Koha.php
```diff
@@ -2985,6 +2995,7 @@ public function renewCheckout($patron, $recordId, $itemId = null, $itemIndex = n
 
 			//Parse the result
 			if (isset($renewResponse->success) && ($renewResponse->success == 1)) {
+			    $renewResults = mysqli_query($this->dbConnection, $renewSql);
```
mdnoble73: If this call is moved to after the renewal is done, the renewal count should not be incremented

code/web/Drivers/Koha.php
```diff
@@ -5979,6 +6000,7 @@ public function findUserByField($field, $value) {
 		} else if ($lookupUserResult->num_rows > 1) {
 			return 'Found more than one user.';
 		}
+		$lookupUserResult->close();
```
mdnoble73, line 6013: This will also need to be done if we return $newUser

code/web/Drivers/Koha.php
```diff
@@ -6973,6 +7011,7 @@ private function getKohaSystemPreference(string $preferenceName, $default = '')
 			global $logger;
 			$logger->log("Error loading system preference " . mysqli_error($this->dbConnection), Logger::LOG_ERROR);
 		}
+		$results->close();
```
mdnoble73: $results is already closed by the time we get here. The only way it will not be is if $results is false so this will throw an error

code/web/Drivers/Koha.php
```diff
@@ -7028,6 +7067,7 @@ public function getEditableUsername(User $user) {
 			if ($curRow = $results->fetch_assoc()) {
 				return $curRow['userId'];
```
mdnoble73, line 7077: $results needs to be closed before the return statement as well

code/web/Drivers/Koha.php
```diff
@@ -7049,6 +7089,7 @@ public function updateEditableUsername(User $patron, string $username): array {
 					'message' => 'Sorry, that username is not available.',
 				];
 			}
+			$results->close();
```
mdnoble73, line 7101: $results also needs to be closed if we return in the previous bolck

code/web/Drivers/Koha.php
```diff
@@ -7424,8 +7468,8 @@ public function getCurbsidePickupSettings($locationCode) {
 						}
 					}
 				}
-				$results->close();
 			}
+			$results->close();
```
mdnoble73: Moving this out of the block will cause an error if $results is false

code/web/Drivers/Koha.php
```diff
@@ -7680,6 +7724,7 @@ function validateUniqueId(User $user) {
 		$lookupUserResult = mysqli_query($this->dbConnection, $sql);
 		if ($lookupUserResult->num_rows > 0) {
 			$lookupUserRow = $lookupUserResult->fetch_assoc();
+			$lookupUserRow->close();
```
mdnoble73: This is the wrong variable

code/web/Drivers/Koha.php
```diff
@@ -8469,6 +8523,8 @@ public function updateMessageQueue(): array {
 				'success' => true,
 				'message' => 'Added ' . $numAdded . ' to message queue'
 			];
+
```
mdnoble73: This needs to be before the return statement

code/web/Drivers/Koha.php
```diff
@@ -8524,6 +8580,8 @@ public function updateUserMessageQueue(User $patron): array {
 
 			}
 
+			$results->close();
```
mdnoble73, line 8601: This needs to be within the block checking if results is false

mdnoble73 (changes requested):
See inline notes

code/web/Drivers/Koha.php
kylemhall: Actually, this line is actually superfluous and I'm removing it in the update. That resultset is already closed right after the call to  `fetch_assoc` so there is no need to close if the insert fails, as it has already been closed!

kylemhall:
@mdnoble73 I've updated the branch! I amended the original commit to include those changes. If you'd rather I put them in a separate commit just let me know!

mdnoble73 (changes requested):
Indenting on this file does not match coding style for many of the changed lines. Should use tabs rather than spaces.

mdnoble73 (approved):

Issue #1989, opened 2024-08-16
Hide empty item groups for volume-level holds


kylemhall:
Libraries do not want empty item groups ( e.g. volumes ) in Koha to show in the list of hold-able volumes in Aspen.

PR #1991, opened 2024-08-16
Hide empty item groups for volume-level holds

kylemhall:
In Koha, it is possible to create item groups (i.e. volumes) without attaching items to that group. If an item group has no items, the volume level hold will be unfillable unless items are added to it at a future time. For these reasons, it makes sense to filter these empty volumes out.

A possible future enhancement would be to make the display of empty volumes configurable.

Test Plan:
1) Set up an instance of Koha connected to Aspen Discovery 2) Create 10 items, and four item groups on one record 3) Attach 3 items to each of 3 of the item groups
4) Run koha_export.jar
5) Note the empty volume shows in the list of volumes you can place
   holds on
6) Check out this branch
7) Recompile koha_export.jar
8) Repeat step 4
8) Note the empty volume is no longer in the volume pulldown

Closes Aspen-Discovery/aspen-discovery#1989


mdnoble73 (changes requested):
This change needs release notes

kylemhall:
@mdnoble73 release notes have been added!

mdnoble73 (approved):

PR #1994, opened 2024-08-16
Implement dependent issues for GitHub Issues

kylemhall:
We are currently using GitHub Issues for issue tracking While we
    are deciding on a long term issue tracker for the community ( Issues or Jira ),
    In the meantime it would be nice if we could use the dependent-issues
    action to deal with issue dependencies in the community.

    Once this branch is merged, setting a dependency requires only that
    one leave a comment on the issue in the format:

    Depends on Aspen-Discovery/aspen-discovery#XXXX

    Where XXXX is the dependent issue number.
    Any number of "Depends on" lines can exist in one or more comments.

    https://github.com/marketplace/actions/dependent-issues#dependent-issues

### DIFF
--- a/.github/workflows/dependent-issues.yml
+++ b/.github/workflows/dependent-issues.yml
@@ -1,0 +1,59 @@
+name: Dependent Issues
+
+on:
+  issues:
+    types:
+      - opened
+      - edited
+      - closed
+      - reopened
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - closed
+      - reopened
+      # Makes sure we always add status check for PRs. Useful only if
+      # this action is required to pass before merging. Otherwise, it
+      # can be removed.
+      - synchronize
+
+  # Schedule a daily check. Useful if you reference cross-repository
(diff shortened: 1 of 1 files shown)


mdnoble73:
Let's wait on this until we have a tool selected. Each support organization currently has their own tool that they are using for tracking issues internally currently.

kylemhall:
> Let's wait on this until we have a tool selected. Each support organization currently has their own tool that they are using for tracking issues internally currently.

There is no reason at all to wait. Once the tool is selected, we can re-implement if necessary. All issues will need to be moved over if the tool used is anything except GitHub Issues anyway.

mdnoble73:
Let's talk through it more at the tools meeting

kylemhall:
@mdnoble73 we forget to bring this up in the tools meeting. I'll try to remember to bring it up tomorrow quickly at the start of the guidelines meeting.

mdnoble73:
This is not needed since the community chose Jira for tracking issues

PR #1995, opened 2024-08-16
Fix save user preferences - deprecated

Chloe070196:
DEPRECATED - Up-to-date PR: [https://github.com/Aspen-Discovery/aspen-discovery/pull/2005](https://github.com/Aspen-Discovery/aspen-discovery/pull/2005)

During the process of updating user preferences, $user->getCatalogDriver() may return null, thus the need for a guard clause in $user->updateHomeLibrary. Since this guard clause makes it possible for $user->updatedHomeLibrary to return null as well, an additional guard clause is added to the launch function for 'MyPreferences'.

This should prevent the 'Member function updateHomeLibrary called on Null' from being displayed on the OPAC upon attempts to save user preferences.

**Test plan:**

without the patch:

- login to Aspen
- navigate to Account > Your Preferences
- toggle one or more of the settings
- click 'Update My Preferences'
- notice the error that is then displayed on screen with the message "Member function updateHomeLibrary called on Null. "
- apply the patch, and go through the steps above once more
- notice that this time, upon saving the user preferences, the preferences page is then displayed again with the updated setting(s)


mdnoble73:
It looks like this happens with Aspen Admin and other accounts not tied to the ILS.  The changes there are good guards, but we should also update the my preferences page to not show the home library and alternate pickup locations etc if the user is not associated with the ILS since they are not relevant. Let me know if you need help with how that works.

mdnoble73:
New PR looks great and has been merged

PR #1999, opened 2024-08-19
Update GitHub actions to improve build time and add support for GHCR

kylemhall:
This pull request greatly reduces the build time for the Aspen docker images. Previously each registry ran an independent build and push step. This branch collapses those down to a single build and push step for each Dockerfile while also adding in support for GHCR. Any future container registry additions would only add a few more seconds to the build time.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,10 @@ jobs:
     if: github.repository == 'Aspen-Discovery/aspen-discovery'
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


mdnoble73:
Is this compatible with the current branching structure or does it require tagging?

kylemhall:
> Is this compatible with the current branching structure or does it require tagging?

It will work with the current branching structure just fine, and will run test builds of the container images. Those container images won't be published unless a new tag is pushed.

mdnoble73 (approved):

PR #2001, opened 2024-08-19
Koha api user agent

lucasmontoya13:


code/web/Drivers/Koha.php
```diff
@@ -902,37 +903,16 @@ public function patronLogin($username, $password, $validatedViaSSO) {
...
-						'Cache-Control: no-cache',
-						'Content-Type: application/json;charset=UTF-8',
-						'Host: ' . preg_replace('~http[s]?://~', '', $this->getWebServiceURL()),
-					], true);
-				}
-				$responseBody = $this->apiCurlWrapper->curlSendPage($apiURL, 'POST', json_encode($postParams));
-				$responseCode = $this->apiCurlWrapper->getResponseCode();
-				$jsonResponse = json_decode($responseBody);
-				ExternalRequestLogEntry::logRequest('koha.patronLogin', 'POST', $apiURL, $this->curlWrapper->getHeaders(), json_encode($postParams), $responseCode, $responseBody, ['password' => $password]);
+				$postResponse = $this->kohaApiUserAgent->post("/api/v1/auth/password/validation",$params,"koha.patronLogin",['password' => $password]);
```
mdnoble73, line 910: This needs to deal with error conditions if the post does not succeed. I.e. if the oAuth fails.

mdnoble73:
This is going to need a test plan to go with it since there is a high potential to break things. The more detailed the test plan is the better.

mdnoble73 (changes requested):

PR #2005, opened 2024-08-21
Fix save user preferences

Chloe070196:
An updated version of [PR1995](https://github.com/Aspen-Discovery/aspen-discovery/pull/1995) : 

It has been rebased against the more recent version of 24.09.00, and also adds a filter to 'Home Library', 'Preferred Pickup Location', 'Alternate Pickup Location 1', 'Alternate Pickup Location 2', and 'Bypass pickup location prompt when placing holds' so that these fields only show for users associated with the ILS integration, as they only are useful in the context of ILS. The name of the home library is still displayed but cannot be edited. (Implements the [changes requested](https://github.com/Aspen-Discovery/aspen-discovery/pull/1995#issuecomment-2297439030) on PR1995)

Test plan:

1- in Aspen, as an admin with no ILS association, go to Account > Your Preferences
2- take note of the fields that display
3- masquerade as a user with ILS association, and repeat 1 and 2
4- compare the fields that showed for the user with no ILS association and those that did for the user with ILS association
ILS-related preference settings should all display for the latter, and only for the latter.
Also ensure that the user with no ILS association still is able to update all settings that are relevant to them.

______________________________________________________________________________________________________________

For reference: PR 1995's description:

During the process of updating user preferences, $user->getCatalogDriver() may return null, thus the need for a guard clause in $user->updateHomeLibrary. Since this guard clause makes it possible for $user->updatedHomeLibrary to return null as well, an additional guard clause is added to the launch function for 'MyPreferences'.

This should prevent the 'Member function updateHomeLibrary called on Null' from being displayed on the OPAC upon attempts to save user preferences.

Test plan:

without the patch:

- login to Aspen
- navigate to Account > Your Preferences
- toggle one or more of the settings
- click 'Update My Preferences'
- notice the error that is then displayed on screen with the message "Member function updateHomeLibrary called on Null. "
- apply the patch, and go through the steps above once more
- notice that this time, upon saving the user preferences, the preferences page is then displayed again with the updated setting(s)

### DIFF
--- a/code/web/services/MyAccount/MyPreferences.php
+++ b/code/web/services/MyAccount/MyPreferences.php
@@ -38,11 +38,14 @@ class MyAccount_MyPreferences extends MyAccount {
 				$allowHomeLibraryUpdates = ($patronHomeLibrary->allowHomeLibraryUpdates == 1);
 			}
 
+			$isAssociatedWithILS = !empty($user->ils_username) && !empty($user->ils_barcode);
+
 			$interface->assign('canUpdateContactInfo', $canUpdateContactInfo);
 			$interface->assign('showAlternateLibraryOptions', $showAlternateLibraryOptionsInProfile);
 			$interface->assign('allowPickupLocationUpdates', $allowPickupLocationUpdates);
 			$interface->assign('allowRememberPickupLocation', $allowRememberPickupLocation);
 			$interface->assign('allowHomeLibraryUpdates', $allowHomeLibraryUpdates);
+			$interface->assign('isAssociatedWithILS', $isAssociatedWithILS);
 
 			// Determine Pickup Locations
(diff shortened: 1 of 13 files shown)


code/web/services/MyAccount/MyPreferences.php
```diff
@@ -38,11 +38,14 @@ function launch() {
 				$allowHomeLibraryUpdates = ($patronHomeLibrary->allowHomeLibraryUpdates == 1);
 			}
 
+			$isAssociatedWithILS = !empty($user->ils_username) && !empty($user->ils_barcode);
```
mdnoble73, line 41: The standard way of doing this is with a call to $user->hasIlsConnection();

mdnoble73 (approved):

PR #2021, opened 2024-08-25
Improves QA check for spaces-instead-of-tabs

NashvillePublicLibrary:


mdnoble73 (approved):

kylemhall:
@mdnoble73 A code review was submitted yesterday. I believe  this change breaks the action in such a way that it always passes.

@NashvillePublicLibrary can you tell us what the regular expression should accomplish? I'd be happy to help!

NashvillePublicLibrary:
@kylemhall I goofed and used the wrong regexp! I meant for it to be `^\+\s* {2,}`. I'll correct that now.

The original version of pull_request_qa.yaml would always fail if the "official" file had a space-space anywhere in the code lines that are changed by the pull request. I stumbled across this in User.php. My inclusion of grep should only look at diff output lines beginning with plus, i.e., those in the updated file. It's got the additional benefit of not looking at any line past initial whitespace (e.g., there are those of us who still have some typewriter muscle memory who reflexively type period-space-space)

kylemhall:
Thanks @NashvillePublicLibrary ! I think your line of thinking is spot on! I wonder if you can just push your change to this same pull request. I'd also be happy to make a pull request for the change if you'd like!

NashvillePublicLibrary:
Imma need some git/github coaching. My earlier changes were merged this morning. In order to have this pull request contain the update, do I first "Revert"?
image

NashvillePublicLibrary:
After a little Googling (and stackoverflowing https://stackoverflow.com/questions/12674304/github-reopening-a-merged-pull-request), I think the easiest and cleanest solution is to compose a new pull request. Unless y'all stop me, that's what I'm gonna do.

NashvillePublicLibrary:
https://github.com/Aspen-Discovery/aspen-discovery/pull/2025

PR #2024, opened 2024-08-26
Get volume info for holds from MARC record when there is no volumes.txt for Symphony

Abhyastamita:
Sirsi is not putting volume information in volumes.txt anymore for new Aspen libraries.  The volume description is in the MARC record in the volume subfield, but the call number key (necessary for placing holds) is not.  Existing libraries still use volumes.txt, so that needs to continue to work unchanged.
- Get the volume information from the MARC record from the item field, volume subfield.  
- Prefix volumeId in the ils_volume_info table with LOOKUP: to trigger looking up the call number key while placing the hold.
- Use the Sirsi API to look up all the call number keys for the record, then look up the keys one by one and find the first one that matches the volume description.  There does not appear to be a more efficient way to do this with the Sirsi API.  Holds with lots of items may be slow.  Use this key to actually place the hold.
- I tested this using CLEVNET's records and API connection.
- Updated release notes.

ByWater ticket 134021.

### DIFF
--- a/code/symphony_export/src/com/turning_leaf_technologies/symphony/SymphonyExportMain.java
+++ b/code/symphony_export/src/com/turning_leaf_technologies/symphony/SymphonyExportMain.java
@@ -40,6 +40,8 @@ public class SymphonyExportMain {
 
 	private static boolean hadErrors = false;
 
+	private static boolean volumesInTextFile = true;
+
 	public static void main(String[] args){
 		if (args.length == 0) {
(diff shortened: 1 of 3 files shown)


Abhyastamita:
I don't think there are actually problems with tabs/spaces in the Java file I edited (or I am missing something obvious) but please let me know if I'm wrong and I'll resubmit.

mdnoble73 (approved):

PR #2025, opened 2024-08-26
Correct regexp for spaces-not-tabs detection

NashvillePublicLibrary:


.github/workflows/pull_request_qa.yml
```diff
@@ -62,7 +62,7 @@ jobs:
             echo "Found modified file: $file";
             if [[ $file == *.php || $file == *.js || $file == *.java ]]; then 
               #echo "Checking $file for whitespace issues"
-              DIFF=$(git diff official/$DEFAULT_BRANCH HEAD -- $file | grep "^+\s*  +")
+              DIFF=$(git diff official/$DEFAULT_BRANCH HEAD -- $file | grep "^\+\s* {2,}")
```
kylemhall, line 65: You new regex works well but could be reduced down to `^\+\s*` without any change in behavior.
The `*` is greedy so there will never by any spaces left for the ` {2,}` to match on.

kylemhall (changes requested):

.github/workflows/pull_request_qa.yml
NashvillePublicLibrary: "\s" != " "

NashvillePublicLibrary: \s includes \t. ^\+\s* would hit on any new line with any leading whitespace (e.g., tab or space).

kylemhall: Ah, so the regex is meant to capture any lines starting with whitespace where the whatspace ends with 2 or more spaces? That makes sense to me! I think this is a great example where enhanced commit messages would improve our communication or workflow!

kylemhall (approved):

Issue #2026, opened 2024-08-27
Add a basic check to the QA GitHub Actions for changeds to createSite.php the may need changes to Docker files


kylemhall:
Some times createSite.php is updated without a corresponding update to the Docker scripts. It would be good to have some kind of automated check to help catch these issues before the pull requests are merged.

PR #2027, opened 2024-08-27
Add GH Action to check for changes in createSite.php that may need changes in docker

kylemhall:
This action attempts to do basic checking of changes to createSite.php that would need changes to Docker files, and then checks to see if any files in the docker directory have been changes. I'm sure we can improve this check with experience.

Closes Aspen-Discovery/aspen-discovery#2026

### DIFF
--- a/.github/workflows/pull_request_qa.yml
+++ b/.github/workflows/pull_request_qa.yml
@@ -2,6 +2,7 @@ name: Check Pull Request for QA Issues
 
 on:
   pull_request:
+  push:
 
 jobs:
   check_release_notes:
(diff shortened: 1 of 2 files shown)


kylemhall:
@lucasmontoya13 can you have a look at this and let me know if there is any way we can improve it?

kylemhall:
Tests:
Failure! https://github.com/kylemhall/aspen-discovery/actions/runs/10565908295/job/29271451099
Success! https://github.com/kylemhall/aspen-discovery/actions/runs/10566183988/job/29272363017

lucasmontoya13:
This PR looks very good! I think i could polish this a little bit more. I will make more tests to corroborate it. Thank you

mdnoble73:
Why check for just Aspen or Koha? What we probably want to do to see if new logic may need to go into Docker is check for changes to the createSiteTemplate.ini since that is where new variables would be added.

kylemhall:
> Why check for just Aspen or Koha? What we probably want to do to see if new logic may need to go into Docker is check for changes to the createSiteTemplate.ini since that is where new variables would be added.

That is a really good point @mdnoble73! I had the adb/ktd integration stuck in my head. I'll amend the pull request.

mdnoble73:
The commit message and testing notes for this are no longer accurate, can you please update with the actual changes? Also I think rather than targeting createSite.php, this should target  createSiteTemplate.ini which would indicate new variables being added or existing variables being needed which the docker scripts may or may not need to respond to.

kylemhall:
> The commit message and testing notes for this are no longer accurate, can you please update with the actual changes? Also I think rather than targeting createSite.php, this should target createSiteTemplate.ini which would indicate new variables being added or existing variables being needed which the docker scripts may or may not need to respond to.

Sounds good! I'll get the patch set amended asap!

kylemhall:
Failure! https://github.com/kylemhall/aspen-discovery/actions/runs/10687602144/job/29625572342
Success! https://github.com/kylemhall/aspen-discovery/actions/runs/10687671062/job/29625785651

mdnoble73 (approved):

PR #2029, opened 2024-08-28
cloudLibrary with alternate cards bug fixes

Abhyastamita:
- Fixed check for useAlternateCard during checkout/hold placement - it was always coming back false, which caused the prompt to enter an alternate card to show up inappropriately.
- Made form input and label alignment more consistent with other forms on library card page and alternate card popup.
- Checkouts made with a linked card will now use the cloudLibrary settings that are in the scope of the linked card's home library.
- Updated release notes


Abhyastamita:
Also includes blocking spammy user agents.  
- Added new function isSpammyUserAgent() to bootstrap.php which is almost the same as isSpammySearchTerm() in index.php, but is case insensitive and includes a few extra spammy strings to check for.
- If the user agent is spammy, then return a 404.
- Updated release notes.

mdnoble73 (approved):

PR #2032, opened 2024-08-29
Docker entrypoint updates

tomascohen:
In order to test this enhancements, you will need to build your own image out of this patchset.

```shell
docker build . -f docker/Dockerfile -t aspendiscovery/aspen
```

After that, follow the README to launch Aspen using the latest `docker-compose.yml` file.


docker/files/scripts/createDirs.php
```diff
@@ -3,34 +3,26 @@
 //Capture any error as ErrorException
 set_error_handler("customErrorHandler");
 
+$newOwner = "www-data"; // default user
 //Check how many arguments has been passed to the script
-if (count($argv) == 1) {
```
kylemhall, line 7: Can we use getopt here instead perhaps?

tomascohen: I'd agree. Follow-up pull request?

lucasmontoya13:
I was able to test these changes in a new instance and everything works well. Splitting the start up script resolves the conflict between services and keeps the containers alive

kylemhall (approved):

docker/files/scripts/start.sh
```diff
@@ -1,43 +1,101 @@
...
 if ! php createDirs.php ; then
-	echo "ERROR : FAILED TO CREATE DIRECTORIES OR TRY TO FIX OWNERSHIP AND PERMISSIONS"
+	echo "%   ERROR: Directories creation and permission fixes failed"
 	exit 1
 fi
 
-# Move and create temporarily sym-links to etc/cron directory
-sanitizedSitename=$(echo "$SITE_NAME" | tr -dc '[:alnum:]_')
-cp "$CONFIG_DIRECTORY/conf/crontab_settings.txt" "/etc/cron.d/$sanitizedSitename"
+# FIXME: This seems to be creating dirs like images/images, etc
```
mdnoble73, line 97: @tomascohen do you want a fix in place for this before the PR is merged or do you want to handle this in a followup request?

docker/files/scripts/start.sh
```diff
@@ -47,24 +105,20 @@ ln -s "$dataDir/images" "$localDir/images"
...
-	cron -f -L 2
-else
-/bin/bash -c "trap : TERM INT; sleep infinity & wait"
-fi
+echo "%"
+echo "%   Aspen Discovery ready to use!"
+echo "%"
+echo "%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%"
 
+# FIXME: We should probably run Apache in foreground instead. We need to
```
mdnoble73, line 122: @tomascohen do you want this to be fixed before merge?

docker/files/scripts/start.sh
```diff
@@ -47,24 +105,20 @@ ln -s "$dataDir/images" "$localDir/images"
...
-# Turn on apache
-if [ "$ASPEN_BACKEND" == "yes" ]; then
-	service apache2 start
-fi
-
-# Run pending database updates
+# Run any pending database updates
 echo "127.0.0.1    $SITE_NAME" >> /etc/hosts
+echo "%   * Triggering pending database updates"
 curl -k http://"$SITE_NAME"/API/SystemAPI?method=runPendingDatabaseUpdates
```
mdnoble73, line 115: Are we setting docker up with SSL enabled? If so, this should be https. If not, we will need to let system administrators know how to setup SSL before they use the docker image in production.

mdnoble73 (changes requested):

docker/files/scripts/start.sh
tomascohen: This is a longer term thing. Not a blocker. I just let the FIXME in place so we don't forget.

docker/files/scripts/start.sh
tomascohen: Administrators using the Docker images will need to learn how to set a reverse proxy that would do the SSL termination itself. This is because this is designed to run many backend instances if required in a HA setup.

docker/files/scripts/start.sh
tomascohen: I think this deserves a separate ticket. Because the solution is to have those dirs on a separate path on the filesystem and is a bit of a bolder fix.

tomascohen:
Is my last commit (Release notes) failing because of the last newline that my editor added?

tomascohen:
I've rebased the last patch (using Github's builtin tool). I wanted to mention that the failed check (check_tabs_vs_spaces) is a false positive, as my spaces changes are made on purpose so the output is cleaner to read. I'm indenting things properly, etc.

mdnoble73:
No worries, that tabs vs/spaces check still doesn't work, I've been manually verifying as needed.

PR #2033, opened 2024-08-29
Fix the check_tabs_vs_spaces GitHub Action

kylemhall:
It appears that quoting a regex in bash causes it to use it as a string
literal so the grep chained to the git diff command never worked.

### DIFF
--- a/.github/workflows/pull_request_qa.yml
+++ b/.github/workflows/pull_request_qa.yml
@@ -2,6 +2,7 @@ name: Check Pull Request for QA Issues
 
 on:
   pull_request:
+  push:
 
 jobs:
   check_release_notes:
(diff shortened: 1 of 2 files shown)


kylemhall:
Failure: https://github.com/kylemhall/aspen-discovery/actions/runs/10622415456
Success: https://github.com/kylemhall/aspen-discovery/actions/runs/10622422197

kylemhall:
Looks really good!

Tabs only, passes: https://github.com/kylemhall/aspen-discovery/actions/runs/10634614571
Spaces, fails: https://github.com/kylemhall/aspen-discovery/actions/runs/10634584251
Tabs with spaces, fails: https://github.com/kylemhall/aspen-discovery/actions/runs/10634354386

mdnoble73 (approved):

PR #2039, opened 2024-09-02
See all locations enhancement

Chloe070196:
Patch enhancing the 'Where is it?' modal by adding edition information as well as book cover thumbnails, as request in [ticket](https://app.plane.so/ptfs-europe/projects/b24083da-a488-4563-a393-9101dacefd66/issues/)

**Test plan**

- in Aspen Administration > Catalog/Grouped Works > Grouped Work Display > <setting in use>,  toggle on the 'Show Covers for Editions' checkbox
- it may be work also ensuring that the "Show Where is it?" option is set to either 'Show first 3 available copies & Where Is It link always' or  'Show Where Is It link only'
- run a search
- open the 'Where is it?' link
- notice that the modal that opens includes thumbnails and edition information
- in Aspen Administration > Catalog/Grouped Works > Grouped Work Display > <setting in use>,  toggle off the 'Show Covers for Editions' checkbox
- run a search
- open the 'Where is it?' link
- notice that the modal that opens includes edition information, but no thumbnails

If testing without the patch, then neither the edition information nor the thumbnails will show

### DIFF
--- a/code/web/sys/Grouping/GroupedWorkFormatSortingGroup.php
+++ b/code/web/sys/Grouping/GroupedWorkFormatSortingGroup.php
@@ -356,7 +356,7 @@ class GroupedWorkFormatSortingGroup extends DataObject {
 	public function loadDefaultFormats() : void {
 		//Automatically generate based on the data in the database.
 		global $aspen_db;
-		$loadDefaultFormatsStmt = "SELECT grouping_category, format FROM grouped_work_variation inner join indexed_format on indexed_format.id = formatId inner join grouped_work on grouped_work.id = grouped_work_variation.groupedWorkId group by format, grouping_category order by grouping_category, lower(format);";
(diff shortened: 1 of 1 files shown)


code/web/interface/themes/responsive/GroupedWork/copyDetails.tpl
```diff
@@ -21,6 +30,13 @@
 				{else}
 					<td>{translate text="%1% of %2%" 1=$item.availableCopies 2=$item.totalCopies isPublicFacing=true}{if !empty($item.availableCopies)} <i class="fa fa-check"></i>{/if}</td>
 				{/if}
+				<td class="notranslate">
+					{if !empty($item.edition)}
+						{$item.edition}
+					{else}
+						{translate text='unknown' isPublicFacing=true}
```
mdnoble73: Rather than displaying unknown, I think we may want to leave this blank since that is consistent with what we do in search results.

mdnoble73 (changes requested):
Let's add toggles to both of these new options within Grouped Work Display settings. The cover will be fairly redundant since it is shown for each item. Some libraries also don't shown the edition in all places.

PR #2076, opened 2024-09-27
Indexing Woes

K-Alette:
- Fix issue where method deleteRecord() was missing result.groupedWorkId in PalaceProjectExportMain.java - causing indexers to stall 
- Use language codes for Boundless titles instead of the full language when building Grouped Work permanent_id

Rebuilt all jar files (NOTE: the palace_project_export.jar did not change after rebuilding)

Updated release notes

To test:
Run Palace Project and Boundless exports


K-Alette:
Just a general note that I am very much hoping this is the last hotfix I'll need to do for indexers in 24.09.01 :)

mdnoble73 (approved):

PR #2099, opened 2024-10-11
FIX: 24.10.00 handle EDS search error

Chloe070196:
EDS errors should be logged and not echoed, and a standard Aspen Error should be displayed to the user.

**Displaying the Aspen Error:** 

Bug: When an Aspen error is thrown for EBSCO EDS search and the `searchError.tpl` file is displayed, instead of the original error message, a PHP error is shown: `Cannot use object of type AspenError as array`

Expected behaviour: ~~The error message refers to a specific cause~~.
Instead of the rather obscure 'Cannot use object of type AspenError as array', the message should clearly indicate that it is to do with processing the EDS search, and indicate whether it is related to authentication. 
(EDITED)

Test plan:
- navigate to Aspen Administration > EBSCO EDS >
Settings
- create a new Setting, make sure to use invalid
credentials and save
- navigate to Aspen Administration > Primary
Configuration > Library Systems and select your
current Library
- set its EDS Setting to the setting you just
created
- run a search in Articles and Databases

**Logging EDS errors instead of echoing**

Issue: When an EDS search error is thrown, its message gets printed on screen to the user. This occurs on the search page and is publicly visible.
Expected behaviour: The error message does not show on the UI anymore and is logged instead.

Test plan:
- navigate to Aspen Administration > EBSCO EDS >
Settings
- create a new Setting, make sure to use invalid
credentials and save
- navigate to Aspen Administration > Primary
Configuration > Library Systems and select your
current Library
- set its EDS Setting to the setting you just
created
- log out
- run a search in Articles and Databases

### DIFF
--- a/code/web/interface/themes/responsive/EBSCO/searchError.tpl
+++ b/code/web/interface/themes/responsive/EBSCO/searchError.tpl
@@ -10,10 +10,10 @@
 	</div>
 	{/if}
 
-	{if !empty($searchError) && !empty($searchError.error.msg)}
+	{if !empty($searchError) && !empty($searchError->getMessage())}
 		<h2>{translate text="Error description" isPublicFacing=true}</h2>
 		<div>
-			{$searchError.error.msg}
+			{$searchError->getMessage()}
 		</div>
 	{/if}
 </div>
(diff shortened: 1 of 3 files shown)


K-Alette:
Screenshot 2024-10-11 084954
For both of these scenarios I get the same message - it looks like you'll need to add the correct messaging in processSearch() in EbscoEdsSearcher.php. 
This is on line 666 for me
The errors do get logged correctly, though :)

mdnoble73 (approved):

PR #2106, opened 2024-10-17
Add the ability to skip running all or specific parts of the qa actions

kylemhall:
Right now, all pull requests will trigger the Github Actions for .github/workflows/pull_request_qa.yml. Some developers are using wei/pull to generate pull requests to keep personal branches up to date with the same branches in the official repo. The bot creates pull requests, and these pull requests can fail some actions as they are not like feature branches.

This commit adds the ability to specify one or more github actions variables to disable all or just some of the qa action jobs.

You can set your variables from:
https://github.com/GITHUB_USERNAME/aspen-discovery/settings/variables/actions

SKIP_GHA_PULL_REQUEST_QA will disable all the tests if created and the value is set to 1 ( or anything else that would evaluate to true )

Alternatively, the following variables will disable indvidual tests: SKIP_GHA_PULL_REQUEST_QA_CHECK_RELEASE_NOTES
SKIP_GHA_PULL_REQUEST_QA_CHECK_TABS_VS_SPACES
SKIP_GHA_PULL_REQUEST_QA_CHECK_CREATE_SITE_FOR_DOCKER

### DIFF
--- a/.github/workflows/pull_request_qa.yml
+++ b/.github/workflows/pull_request_qa.yml
@@ -5,7 +5,17 @@ on:
   push:
 
 jobs:
+  debug:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Print GitHub vars
+        run: echo "$GITHUB_CONTEXT"
+        env:
+          GITHUB_CONTEXT: ${{ toJson(vars) }}
+
   check_release_notes:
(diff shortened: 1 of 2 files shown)


mdnoble73:
With the changes Tomas has made, I do not believe these changes are still necessary.  Please let me know if that is not the case.

PR #2107, opened 2024-10-17
Freeze holds issue

lucasmontoya13:


code/web/sys/CurlWrapper.php
```diff
@@ -217,9 +217,9 @@ public function curlSendPage(string $url, string $httpMethod, $body = null) {
 		} else {
 			curl_setopt($this->curl_connection, CURLOPT_CUSTOMREQUEST, $httpMethod);
 		}
-		if ($body != null) {
+
```
mdnoble73, line 220: I think this is going to cause errors in other places. Let's keep the check for if body is null.

code/web/Drivers/KohaApiUserAgent.php
```diff
@@ -77,19 +77,24 @@ public function get(string $endpoint, string $caller, array $dataToSanitize = []
...
 	 * @return  array|bool     An array containing the response body and response code retrieved by the request or false if authorization fails.
 	 * @access  public
 	 */
-	public function post(string $endpoint, array $requestParameters, string $caller, array $dataToSanitize = [], array $extraHeaders = null): array|bool {
+	public function post(string $endpoint, array|null $requestParameters, string $caller, array $dataToSanitize = [], array $extraHeaders = null): array|bool {
 		// Preparing request
 		$apiURL = $this->baseURL . $endpoint;
-		$jsonEncodedParams = json_encode($requestParameters);
 
+		if ($requestParameters == []) {
```
mdnoble73, line 92: This could be checked with empty($requestParameters) which would also check for null.

mdnoble73 (changes requested):

PR #2109, opened 2024-10-18
24.11.00 idor fix

AlexanderBlanchardAC:
A bug has been reported where user can add to the list of another user. 
TEST PLAN: 
Prior to Applying Patch:
Replicate the bug:
1. Log into Aspen and go to your account, select Masquerade and Masquerade as Library card user number 42. 
2. Create a list. 
3. Carry out a search and select a book, click add to list. Use the developer tools to take notice of the id of the list you are adding to and note it down. 
4. End your masquerade and create a list as Aspen_Admin. 
5. Use the search function and add a different book to your list. 
6. In the Dev Tool, use the Network tab to navigate to the saveToList method and right click. 
7. Select Edit and Resend
8. In the Network console, alter the list id number to the one you noted down and click send. 
9. Masquerade as library card number 42 again and check your list, the second book will now be part of this list too. 
APPLY THE PATCH
Repeat the steps above, this time when you select 'send' you should see in the body section that success is false and that you are not authorized to edit this list.

### DIFF
--- a/code/web/services/MyAccount/AJAX.php
+++ b/code/web/services/MyAccount/AJAX.php
@@ -7094,7 +7094,8 @@ class MyAccount_AJAX extends JSON_Action {
 					$listOk = false;
 				} else {
 						//Authorization check: Ensure list belongs to logged in user
-					if ($userList->user_id != UserAccount::getActiveUserId()) {
+					$currentUser = UserAccount::getActiveUserObj();
+					if (!$currentUser->canEditList($userList)) {
 						$result['success'] = false;
 						$result['message'] =translate([
 							'text' => 'You are not authorized to modify this list.',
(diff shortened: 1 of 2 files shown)


code/web/services/MyAccount/AJAX.php
```diff
@@ -7085,14 +7085,24 @@ function saveToList() {
...
 				$userList->id = $listId;
-				$totalRecords = $userList->numValidListItems();
 				if (!$userList->find(true)) {
 					$result['success'] = false;
 					$result['message'] = translate([
 						'text' => 'Sorry, we could not find that list in the system.',
 						'isPublicFacing' => true,
 					]);
 					$listOk = false;
+				} else {
```
mdnoble73, line 7095: There is a permission for administrators to edit all lists ('Edit All Lists') which this prevents as well.

There is a method on the user object to check if a user can edit a list.  $user->canEditList($list). We should use that to verify the list can be modified so if we add additional rules in the future they will be checked.

mdnoble73 (changes requested):

AlexanderBlanchardAC:
Updated to use the canEditList function. Temporarily turned off ability to edit all lists for opac admin to test so that the rest of the test plan remains the same.

mdnoble73 (approved):

PR #2112, opened 2024-10-22
Remove Partner-Specific Directories

K-Alette:
Remove directories for BWS partners from the repository so conf files stop getting overwritten
These directories will have to be added to the server after this code is deployed


mdnoble73:
Can you add release notes for these?

K-Alette:
> Can you add release notes for these?

Done! Sorry for the delay 🫡

mdnoble73 (approved):

PR #2118, opened 2024-10-25
Only count the new commits

tomascohen:


kylemhall:
@tomascohen can you describe the issue, and also is the `asd` commit supposed to be on this?

tomascohen:
Hi. Thanks for looking at this. This pull request was a POC I expected to trigger the actions without the need to integrate the changes into core until polished. Hence the 'asd' commit.

The problem I'm trying to solve is:
1. Change Koha.php
2. Adjust the release notes
3. Generate a commit with those two changes.
4. Generate a PR
=> FAIL: the automatic QA checks alert about totally unrelated files not passing the checks

The base problem is the `git diff --name-only` command is comparing the currently submitted PR with the project's default branch. This is not correct, as it should really be comparing the actually changed files! When the default branch moves a few commits, the `git diff` command will actually account the changes added to the default branch as if they were introduced by the PR. My patch is making it only count the last X commits applied on top of the commit the development was made against. The tests can fail randomly depending on what happens to the default branch being compared to.

This also highlights the fact we are not paying attention to failures in general, as there are `tabs vs spaces` errors introduced in the main branch, but being accounted to the PR. The end result being the bigger problem: we are accepting commits that don't follow the coding guidelines.

tomascohen:
For testing this, I decided to go around and add commits that would make tests pass/fail.

## Release notes checks

With release notes changes: https://github.com/Aspen-Discovery/aspen-discovery/actions/runs/11598581549/job/32294744393?pr=2118
Without release notes changes: https://github.com/Aspen-Discovery/aspen-discovery/pull/2118/checks

## Tabs vs. spaces checks

With spaces added: https://github.com/Aspen-Discovery/aspen-discovery/actions/runs/11598793959/job/32295451911?pr=2118
With tabs replacing wrong spaces: https://github.com/Aspen-Discovery/aspen-discovery/actions/runs/11598811508/job/32295511867?pr=2118

kylemhall (approved):

mdnoble73 (approved):

PR #2134, opened 2024-11-08
Tweak Nashville Student Report

NashvillePublicLibrary:
Feedback from the school librarians includes
+ tweak the appearance of the overdue cards 
    - increase the white space between cover image and text description
+ Return to report via End Masquerade instead of browser back button is slightly confusing and slightly annoying


NashvillePublicLibrary:
I do not yet have a solution for "Return to report via End Masquerade instead of browser back button is slightly confusing and slightly annoying"... but the white space change is approved by school librarians.

mdnoble73 (approved):

PR #2148, opened 2024-11-22
DIS-43: Standarize parameters used for Solr indexing

kylemhall:
Calls to SolrClient commit() use many permutations of the boolean parameters for waitFlush, waitSearcher, and softCommit. The calls are uncommented so the justification for the settings is unknown. Based on recommendations from a Solr expert we should set all these calls to commit should use false,false,true

Test Plan:
1) Apply this patch
2) Run actions that should trigger chnages to solr indexes
3) No changes in search results should be noted


mdnoble73:
This has the potential to affect the visibility of records as they are added to the index. Most likely for lists, events, open archives, etc. 

Need to test:
1) Start solr
2) Index lists, course reserves, etc
3) Apply the change
4) Make a change to a list, course reserve, event, grouped works etc. 
5) After the indexer runs again verify that the new updates show in Aspen. 

The release notes are also not done properly, you need to create a new section for yourself to avoid conflicts with other developers.

kylemhall:
@mdnoble73 can you point me to the documentation for the release notes formet? Thanks!

mdnoble73:
@kylemhall of course it's here: https://dev.aspendiscovery.org/release-notes also linked to from the home page of dev.aspendiscovery.org

kylemhall:
Thanks @mdnoble73!

mdnoble73:
Let me know when the above have been tested and I can merge and rebuild the jars

kylemhall:
> Let me know when the above have been tested and I can merge and rebuild the jars

We tested this on Clevnet where it worked great!

tomascohen:
Hi, after consulting some experts on the subject, I can confirm there is almost never a reason to use hard commits. Soft commits increase performance and do not imply risks of loosing data. Soft commits are more performant for NRT visibility while hard commits reduce startup time for SolR in the event of a crash ( see https://solr.apache.org/guide/solr/latest/configuration-guide/commits-transaction-logs.html ). The latter seems to be exclusive the use case for them.

In turn, all hard commits affect all of indexers because all cores are on a single SolR instance. That is not to say that a hard commit on one core triggers a hard commit on all cores, but that SolR has limited resources and dedicating any amount of them to unnecessary hard commits reduces overall performance of the system.

Aside from this theoretical analysis, truth is we are seeing degraded performance at the bigger scale, which is highlighting the indexing strategy NEEDS to be revisited. Moving into soft commits is one step in that direction which shouldn't cause any behavior regression.

Simply put, we can avoid debugging complicating and perhaps even unpredictable commit conflicts and resource conflicts simply standardizing on soft commits. There is simply no down side.

That doesn't predate the fact this needs to be tested by the team and that release notes patches need to be well formed.

mdnoble73:
Thanks for testing!

mdnoble73 (approved):

PR #2149, opened 2024-11-28
Add installer script for Ubuntu 22.04

kenstir:
### Summary

Adds an installer script for Ubuntu 22 LTS, derived from `installer_debian.sh`.

### Details

This script fixes the two problems with running `installer_debian.sh` on Ubuntu:
* the PHP version is older than Ubuntu
* the Sury PHP repo installation doesn't work

But I'm not sure yet if this is a good direction to go in.  This script is really not a script to set up a development environment, because it ends up running `chown` and `chmod` on a whole bunch of repo files, which makes the clone unusable by git and appear modified.

### DIFF
--- a/install/createSite.php
+++ b/install/createSite.php
@@ -236,7 +236,7 @@ if (!$foundConfig) {
 		$variables['databaseHost'] = "localhost";
 	}
 
-	$variables['databasePort'] =  readline("Database host for Aspen (default: 3306) > ");
+	$variables['databasePort'] =  readline("Database port for Aspen (default: 3306) > ");
 	if (empty($variables['databasePort'])){
 		$variables['databasePort'] = "3306";
 	}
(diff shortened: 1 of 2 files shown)


tomascohen:
Is this still relevant?

mdnoble73:
I have not reviewed since it never moved out of draft and wasn't asked to look in detail, but I'm sure it's a good starting point for an Ubuntu installer.

kenstir:
At the time of the PR, I was looking for an opinion.  I wrote:
>  I'm not sure yet if this is a good direction to go in

Since then, I myself do not need this PR because I switched to using [aspen-dev-box-image with Evergreen](https://github.com/Jacobomara901/aspen-dev-box-image/tree/allow-other-ils).  Sadly that foundered also; it looks like @Jacobomara901 never requested it to be merged back to origin.

PR #2150, opened 2024-11-28
Spammy ip check toggle

lucasmontoya13:


mdnoble73:
What is the need for this? Are we seeing valid requests get flagged?

lucasmontoya13:
This is an implementation requested by ByWater. They are trying to add some more complex tools in front of Aspen to protect it, and they need to be able to disable spam checks if necessary so they don't interfere. 
Ticket  : https://aspen-discovery.atlassian.net/jira/software/c/projects/DIS/issues/DIS-44?jql=project%20%3D%20%22DIS%22%20ORDER%20BY%20created%20DESC

mdnoble73:
I asked for clarity on the ticket. I don't see anything in the code that would interfere with a filter before Aspen.

mdnoble73:
@lucasmontoya13 based on the comments on DIS-44, rather than always returning false for isSpammySearchTerm, let's update to not do the automatic creation of IP Addresses that block traffic. In index.php, we can bypass setting the blockedForSpam for the IP address when the setting is off.

lucasmontoya13:
Alright! I'll make the appropiate changes. Thanks a lot

mdnoble73 (approved):

PR #2158, opened 2024-12-03
DIS-60: LibKey integration

Chloe070196:
Ticket: https://aspen-discovery.atlassian.net/browse/DIS-60

Test plans are found in commit messages

___________________________________________

**Description**:

This integration aims to allow libraries with a LibKey subscription to have direct links to full-text documents displayed on the eContent records shown in search results in Aspen, provided these records have a DOI.

Public

When viewing library catalogue search results of an eContent item-type, either on the search result page or on the detailed view page for a record, users can see an “Access Online” button:

if the LibKey API returned a direct link to the resource itself, then this takes the user directly to the full-text document, opening it in a new tab.

if the LibKey API does not return a direct link, this takes the user to the external page though which access to the resource can be obtained (this is the current default behaviour in Aspen).

On the detailed view page for the record, under the ‘Links’ section (if enabled), both the original link as found on the Marc record and the LibKey-issued link (if found).

Staff

Admin users can grant or remove the permission to administer the LibKey settings.

Admin users with the required permission can administer LibKey (CRUD, and the possibility to assign a specific LibKey set of settings to a specific location or library). The LibKey Settings are found under “Third Party Enrichment” in the Aspen Administration menu.


code/web/Drivers/LibKeyDriver.php
```diff
@@ -0,0 +1,24 @@
...
+
+class LibKeyDriver {
+
+	public function getLibKeyLink(string $doiUrl): string | null {
+		require_once ROOT_DIR . '/sys/LibKey/LibKeySetting.php';
+		$activeLibrary = Library::getActiveLibrary();
+		$settings = new LibKeySetting();
+		$settings->whereAdd("id=$activeLibrary->libKeySettingId");
+		if ($settings->find(true)) {
+			$settings->fetch();
```
mdnoble73: The fetch is not needed with the find(true) that automatically calls a fetch.

code/web/Drivers/LibKeyDriver.php
```diff
@@ -0,0 +1,24 @@
...
+		}
+		$curlWrapper = new CurlWrapper;
```
mdnoble73, line 15: This section should not be done if there are no settings, should go into the if block.

code/web/Drivers/LibKeyDriver.php
```diff
@@ -0,0 +1,24 @@
...
+		$response = $curlWrapper->curlGetPage("https://public-api.thirdiron.com/public/v1/libraries/" . $settings->libraryId  . "/articles/doi/" . $this->extractDoi($doiUrl) . "?access_token=" . $settings->apiKey);
+		if (empty($response)) {
+			return null;
+		}
+		return json_decode($response, true)["data"]["bestIntegratorLink"]["bestLink"];
+	}
+	public function extractDoi(string $url): string {
```
mdnoble73, line 22: Should only doi urls be processed or does everything get processed?

code/web/Drivers/LibKeyDriver.php
```diff
@@ -0,0 +1,24 @@
...
+	public function getLibKeyLink(string $doiUrl): string | null {
+		require_once ROOT_DIR . '/sys/LibKey/LibKeySetting.php';
+		$activeLibrary = Library::getActiveLibrary();
+		$settings = new LibKeySetting();
+		$settings->whereAdd("id=$activeLibrary->libKeySettingId");
+		if ($settings->find(true)) {
+			$settings->fetch();
+		}
+		$curlWrapper = new CurlWrapper;
+		$response = $curlWrapper->curlGetPage("https://public-api.thirdiron.com/public/v1/libraries/" . $settings->libraryId  . "/articles/doi/" . $this->extractDoi($doiUrl) . "?access_token=" . $settings->apiKey);
```
mdnoble73, line 16: Does the doi URL need any sort of encoding? Since that is a full URL I could see having a URL within the URL causing problems.

code/web/RecordDrivers/MarcRecordDriver.php
```diff
@@ -1314,6 +1314,14 @@ function createActionsFromUrls($relatedUrls, $itemInfo = null, $variationId = 'a
 		$i = 0;
 		if (count($relatedUrls) > 1) {
 			//We will show a popup to let people choose the URL they want
+			foreach($relatedUrls as $relatedUrl) {
```
mdnoble73: For performance we should check if LibKey is enabled before checking all of the URLs.

code/web/sys/Account/User.php
```diff
@@ -3872,13 +3872,18 @@ public function getAdminActions() {
 			'Administer All System Messages',
 			'Administer Library System Messages',
 		]);
+		$sections['local_enrichment']->addAction(new AdminAction('System Messages', 'System Messages allow you to display messages to your patrons in specific locations.', '/Admin/SystemMessages'), [
```
mdnoble73: It looks like this was duplicated from the previous lines and is not needed.

code/web/sys/DBMaintenance/version_updates/24.12.00.php
```diff
@@ -80,6 +80,34 @@ function getUpdates24_12_00(): array {
 		//alexander - PTFS-Europe
 
 		//chloe - PTFS-Europe
+		'create_libkey_module' => [
```
mdnoble73: Is the LibKey Module necessary?

code/web/Drivers/LibKeyDriver.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+class LibKeyDriver {
```
mdnoble73, line 3: We may be able to optimize some of the code by moving the functionality to get the LibKey link to the setting.

mdnoble73 (changes requested):
Added several comments inline. My main concern is around performance and behavior if LibKey is not enabled. Currently the LibKey service will still be called even when not enabled.

code/web/Drivers/LibKeyDriver.php
Chloe070196: I would say there is no point in processing urls that do not contain DOIs as LibKey requires DOI numbers.

Would we want to attempt to check whether urls contain DOIs early then, and add control flow so that the getLibKeyLink method is not called for those?

code/web/sys/DBMaintenance/version_updates/24.12.00.php
Chloe070196: I was not entirely sure whether it would be - happy to remove it

code/web/sys/Account/User.php
Chloe070196: This looks like an incorrectly executed merge conflict resolution on my part - I'll remove it

code/web/Drivers/LibKeyDriver.php
Chloe070196: I was hoping the extractDoi() method would negate this by only returning the DOI itself and removing the rest - but this might be an oversimplification of the issue on my end.

I take it the extractDoi() method needs to be updated?

code/web/Drivers/LibKeyDriver.php
Chloe070196: Thank you for reminding me of that, it'll be removed.

code/web/Drivers/LibKeyDriver.php
Chloe070196: I have looked at inverting the logic so that if a setting is not found, then the getLibKeyLink method returns null early (immediately after the check) - would this be an acceptable solution?

mdnoble73: Yep that would work

code/web/Drivers/LibKeyDriver.php
mdnoble73: Either way should be fine. I think we just want to avoid the call to the libkey server.

code/web/Drivers/LibKeyDriver.php
mdnoble73: I'm not 100% sure. What do those URLs look like to start?

Chloe070196: I would expect `https://doi.org/` followed by the DOI number, that's standard as far as I know, so that is the format I would expect all of them to follow.

If an item did contain just a DOI number in their Marc field 856 u, then it would end up as 'http://' followed by the DOI number (not a valid DOI URL format to my knowledge) by the time it hits the LibKey integration (unsure where this occurs). I did account for that potentially happening, but perhaps it is unnecessary.

code/web/sys/DBMaintenance/version_updates/24.12.00.php
```diff
@@ -80,6 +80,27 @@ function getUpdates24_12_00(): array {
 		//alexander - PTFS-Europe
 
 		//chloe - PTFS-Europe
+		'create_libkey_permissions' => [
+			'title' => 'Create LibKey Permissions',
+			'description' => 'Add an LibKey permission section containing the Administer LibKey Settings permission',
+			'sql' => [
+				"INSERT INTO permissions (name, sectionName, requiredModule, weight, description) VALUES ( 'Administer LibKey Settings','LibKey','LibKey', 0, 'Allows the user to administer the integration with LibKey')",
```
mdnoble73: This should be updated to not require the LibKey module since we removed that.

code/web/Drivers/LibKeyDriver.php
```diff
@@ -0,0 +1,29 @@
...
+	public function getLibKeyLink(string $doiUrl): string | null {
+		if (!$this->containsDoi($doiUrl)) {
+			return null;
+		}
+		require_once ROOT_DIR . '/sys/LibKey/LibKeySetting.php';
+		$activeLibrary = Library::getActiveLibrary();
+		$settings = new LibKeySetting();
+		$settings->whereAdd("id=$activeLibrary->libKeySettingId");
```
mdnoble73, line 11: This could be simplified to 
$settings->id = $activeLibrary->libKeySettingId
which will also take care of escaping fields properly rather than just adding a string. Not a big enough deal to prevent merging, but would be good to follow up on.

mdnoble73 (approved):

PR #2160, opened 2024-12-03
Add Mac option to createSite.php for Mac development

Abhyastamita:
- Added Mac option to createSite.php and updated it so that it can run on a Mac being used for development.  The Mac version is different in these ways:
  - Uses '/usr/local/data/aspen-discovery' instead of '/data/aspen-discovery' since Macs will not allow directories to be added to the root directory.
  - Prompts the user to set the group for Apache.  Defaults to _www because this is what is working for me, but it might be different depending on how you're running Apache.  
  - Creates '/usr/local/data/aspen-discovery' and '/var/log/aspen-discovery' if they don't exist and sets the owner:group to [logged-in-user]:[Apache group]
  - Generally based on the Windows setup since these are development, not production setups.
- Added Mac config template files based on the ones that already existed for Windows and Linux.

This worked for my Mac, which is running on MacOS Sonoma.  I have PHP, Apache, and MariaDB installed via Homebrew.  I don't have the resources to test it in other environments, but I think it's an improvement over the existing createSite.php, which won't run at all on a Mac.


Abhyastamita:
I have a failed check for "check create site for docker", but I don't know enough about how the Docker component works to know if this is safe to ignore?  I'm also not sure if it was actually necessary to change createSiteTemplate.ini.  I'm happy to back those changes out or update whatever else needs updating.

mdnoble73:
That is a false positive and really just a warning. No need to change anything.

PR #2166, opened 2024-12-06
Add an initialization condition

lucasmontoya13:
- Check if ils tables have been initialized in the DB


mdnoble73:
@lucasmontoya13 can you add release notes?

mdnoble73 (approved):

PR #2169, opened 2024-12-11
Cron to dismiss old Year in Review messages (DIS-74)

Abhyastamita:
- Added a dismissYearInReviewMessages.php cron which gets today's date and checks to see if there is a match in any Year in Review settings.  If there is, then it sets all undismissed user messages with a type of yearInReview to dismissed.
- Added UpdateCron_24_12_00.php to update the crontab so that the new cron will run every day at 3.
- Added upgrade scripts to run the cron update.
- Updated release notes.


mdnoble73:
@Abhyastamita we no longer need to add update version scripts just to insert new lines into cron. The upgrade script will automatically look for an updateCron_{version}.php file and run it so if you rename updateCron_24_12.php to updateCron_24.12.00.php it should run automatically without the other files.

Abhyastamita:
Thanks @mdnoble73 !  That's good to know.  I renamed the file and took out the extra ones.

mdnoble73 (approved):

PR #2170, opened 2024-12-12
DIS-110 Add clarification to Check Record for Large Print

kidclamp:
This patch adds a few debug statements to the diagnostics when determining format and adds a note to the 'Check Record for Large Print' setting

To test:
 1 - Add two records to ILS:
      One with some 'Books' type items and some 'Large Print' items
      One with an item with an item type that has a blank 'Format' colum in the format map, and some 'Book' type items
      Both records should have 'large print' in the 250 field
 2 - Ensure 'Check Record for Large Print' is checked in:
      ILS Integration->Indexing profiles->Format Information
 3 - View the records, note they are not listed as 'Large Print' only
 4 - Go to Staff View->Diagnostics (may need to force reindex the record)
 5 - Note that there is no info on 'Large Print' check
 6 - Apply patch, rebuild reindexer.jar, reindex the records
 7 - View diagnostic information and note new info about why Large Print check didn't happen
 8 - Change 'Books' to 'Book' and define a Format of 'Book' or 'Large Print' for previously undefined item
 9 - Reindex records, they should now be large print
10 - View diagnostics and note info about the Large Print check


mdnoble73:
@kidclamp can you add release notes for this? Let me know if you need the documentation for how to add release notes.

kidclamp:
Yes, please, documentation on that would helpful

kidclamp:
added, let me know if I got that right ;-)

mdnoble73 (approved):

PR #2174, opened 2024-12-17
Dis 73 25.01.00 custom form permissions

AlexanderBlanchardAC:
This patch limits the custom forms displayed in the dropdown when adding a custom form to a custom page to those of the user's home library if the user's permissions are set to Administer Library Custom Forms. See test plan on JIRA.

https://aspen-discovery.atlassian.net/browse/DIS-73


mdnoble73:
This needs to take into account any additional libraries/locations that the user is allowed to administer.

AlexanderBlanchardAC:
> This needs to take into account any additional libraries/locations that the user is allowed to administer.
I have added a commit to address this and updated the test plan on Jira.

mdnoble73 (approved):

PR #2175, opened 2024-12-17
Koha showopacnote

ammopt:
Previously, when library's 'showOpacNote' was disabled, the Koha user's
opac note was still being displayed. This is because it's getting added
to _web_note regardless of the setting's value.
This patch removes the opacnote from Koha's getILSMessages, and adds the
setting check to when opacnote should be added to loadContactInformation
or not.

### DIFF
--- a/code/web/Drivers/Koha.php
+++ b/code/web/Drivers/Koha.php
@@ -1357,11 +1357,6 @@ class Koha extends AbstractIlsDriver {
 			$user->_state = $userFromDb['state'];
 			$user->_zip = $userFromDb['zipcode'];
 			$user->phone = $userFromDb['phone'];
-
-			$user_library = $user->getHomeLibrary();
-			if ($user_library !== null && $user_library->showOpacNotes) {
-				$user->_web_note = $userFromDb['opacnote'];
-			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


mdnoble73:
Can you create a Jira ticket for this since, this changes when and where the note shows so I'd like to make sure we understand the implications.

ammopt:
Hi Mark, thank you, I 100% agree.
I've created the Jira ticket and provided a more in-depth explanation of the approach and why I went with it:
https://aspen-discovery.atlassian.net/jira/software/c/projects/DIS/boards/2?selectedIssue=DIS-137

code/web/Drivers/Koha.php
```diff
@@ -6939,15 +6942,6 @@ public function getILSMessages(User $user) {
 		$results = mysqli_query($this->dbConnection, $sql);
 		if ($results !== false) {
 			if ($curRow = $results->fetch_assoc()) {
-				if ($library->showOpacNotes) {
```
mdnoble73: This is used in cases where _web_note is not set by calling loadContactInformation so we shouldn't remove this. Instead we could just rely on this function and remove setting _web_note in loadContactInformation. That will prevent showing duplicate messages as well.

mdnoble73 (changes requested):

ammopt:
Thank you for looking here, Mark. Sounds good, I've re-tested it myself, works, follow-up submitted.

mdnoble73:
@ammopt can you update 25.01 release notes rather than 24.12 release notes?

ammopt:
> @ammopt can you update 25.01 release notes rather than 24.12 release notes?

Hi Mark, apologies, yes of course. Done.

mdnoble73 (approved):

PR #2180, opened 2024-12-19
Syndetics upc

ammopt:
On adb, with ktd, we can use record '76' as that has a UPC with a leading zero:
http://localhost:8083/Record/76

Enter valid syndetics credentials or apply the test only patch to override the need for syndetics credentials through the UI, though you still need to add a valid key through the code, your pick.
Access the record:
http://localhost:8083/Record/76
Notice the image is not found, this is because this record's UPC is '021561601628', and processImageURL is not identifying the returned image as a "not found" because it doesn't match any of the checksums it's verifying.
Apply the patch. Repeat the test plan. You'll need to click "Staff View" -> "Reload Cover" on the record page to clear the cover cache and force the request again but now the image should be displaying correctly. This is because the leading zeroes fail check and the returned image is properly identified as a 'not found' image, so getGroupedWorkCover will then check again but without stripping the leading zeroes, and work then.

Finally, I decided to test for the other sizes as well, and noticed all the checksums are wrong, not the just the medium. Possibly caused by a change of default 'not found' images on syndetics' end.

To test this, visit the following locally:
http://localhost:8083/bookcover.php?id=ils:76&size=small&upc=021561601628&reload=1
http://localhost:8083/bookcover.php?id=ils:76&size=medium&upc=021561601628&reload=1
http://localhost:8083/bookcover.php?id=ils:76&size=large&upc=021561601628&reload=1

None of these work before the fix. All work after the fix.

### DIFF
--- a/code/web/interface/themes/responsive/Report/studentBarcodes.tpl
+++ b/code/web/interface/themes/responsive/Report/studentBarcodes.tpl
@@ -50,7 +50,7 @@ var patron = [];
 				textAlign: "left",
 				textMargin: 2,
 				textPosition: "top",
-				height: 40,
+				height: 33,
 				margin: 0,
 				width: barcodeWidth
 			});
@@ -66,6 +66,7 @@ document.addEventListener("DOMContentLoaded", function(event) {
 <style>
 .avery5160 {
 	/* Avery 5160 labels */
+	box-sizing: border-box !important;
 	width: 2.625in !important;
 	height: 1in !important;
 	margin: 0in .125in 0in 0in !important;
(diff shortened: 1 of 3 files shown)


ammopt:
This is [DIS-139](https://aspen-discovery.atlassian.net/browse/DIS-139)

[DIS-139]: https://aspen-discovery.atlassian.net/browse/DIS-139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mdnoble73:
Do we know 100% that the old checksums are not in use? If not, the new ones you found should be added in addition to the old ones rather than replacing them.

ammopt:
> Do we know 100% that the old checksums are not in use? If not, the new ones you found should be added in addition to the old ones rather than replacing them.

Fair. I've amended the patch and submitted.
After force-pushing, this PR ended up with an unintended commit, I force-pushed again and the PR ended up closed.
I've opened a new one https://github.com/Aspen-Discovery/aspen-discovery/pull/2184

PR #2183, opened 2024-12-22
DIS-147 Fix volumeId check for the items without valid item group while placing volume holds

LiYanjun19:
In code/web/Drivers/Koha.php, modified the conditional check for `$volumeId` to use `!empty($volumeId)` instead of `$volumeId != 0`. This ensures that when volumeId is undefined, the check can work successfully and remove the "item_group_id" in $postParams. In this case, the items can be placed successfully even without a valid volume Id. 

Use-Case:
1. Before PR:
   - Placing a hold on a volume with a valid item group worked correctly.
   - Placing a hold without selecting a specific volume caused a server error.
2. After PR: Both holds (with and without selecting a specific volume) are placed successfully without errors.

### DIFF
--- a/code/web/Drivers/Koha.php
+++ b/code/web/Drivers/Koha.php
@@ -1940,7 +1940,7 @@ class Koha extends AbstractIlsDriver {
 		} else {
 			$apiUrl = $this->getWebServiceUrl() . "/api/v1/holds";
 			if ($this->getKohaVersion() >= 22.11) {
-				if ($volumeId != 0){
+				if (!empty($volumeId)){
 					$postParams = [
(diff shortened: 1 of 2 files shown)


mdnoble73:
@LiYanjun19 does this need to be patched for 24.12? If so we will need to create a 24.12.01 branch for it. If not can you submit against 25.01? 

Thanks, 

- Mark

mdnoble73 (approved):

mdnoble73:
@LiYanjun19 these changes look good, but either should be version 24.12.01 or 25.01, not 24.12.000

PR #2187, opened 2024-12-31
DIS-147 Fix volumeId check for the items without valid item group while placing volume holds

LiYanjun19:
In code/web/Drivers/Koha.php, modified the conditional check for $volumeId to use !empty($volumeId) instead of $volumeId != 0. This ensures that when volumeId is undefined, the check can work successfully and remove the "item_group_id" in $postParams. In this case, the items can be placed successfully even without a valid volume Id.

Use-Case:

Before PR:
Placing a hold on a volume with a valid item group worked correctly.
Placing a hold without selecting a specific volume caused a server error.
After PR: Both holds (with and without selecting a specific volume) are placed successfully without errors.


LiYanjun19:
@mdnoble73 Thank you, I resubmitted one against 25.01.00

PR #2189, opened 2025-01-02
DIS-180 Hoopla Daily Indexing Behavior Optimized

LiYanjun19:
Issue:
1. Hoopla is currently indexing every night at 1 am UTC, making multiple API calls within the same hour window. This behavior leads to unnecessary server load and duplicate data extraction. 
2. In Hoopla settings there’s an “index by day” option, libraries are recommended to select this option when implementing Hoopla. Without this setting selected, Aspen makes API calls every 5 min, easily causing Hoopla server overloaded if we miss selecting this option. 
3. In the original Hoopla code, if Hoopla daily indexing is off schedule, we will catch up after 32 hours from the last update time. This catch up mechanism lacks retry litmis. 

Enhancement:
1. Optimized Daily Indexing:

- Modified indexing schedule to run at 1 am Local Time

- Implemented paginated API calls with 500 records per request 

- Added early termination logic for the successful extraction 

- Enhance error handling for daily extraction 

2. Set “index by day” as a default setting 

3. Optimized 32 hours catch up mechanism:

- Implemented a 3-retry limit for 32 hour catch up scenarios 

- Added error logging


code/web/sys/Hoopla/HooplaSetting.php
```diff
@@ -73,7 +73,10 @@ public static function getObjectStructure($context = ''): array {
 				'type' => 'checkbox',
 				'label' => 'Index By Day',
 				'description' => 'Whether or hoopla indexing should only occur once a day',
-				'default' => 0,
+				'default' => 1,
+				'readOnly' => true,
+				'forcedDefault' => true,
```
mdnoble73: forcedDefault isn't a property for fields. If you want all settings to update you will need to do a database update in 25.01.00.php

code/web/sys/Hoopla/HooplaSetting.php
```diff
@@ -73,7 +73,10 @@ public static function getObjectStructure($context = ''): array {
 				'type' => 'checkbox',
 				'label' => 'Index By Day',
 				'description' => 'Whether or hoopla indexing should only occur once a day',
-				'default' => 0,
+				'default' => 1,
+				'readOnly' => true,
```
mdnoble73, line 77: This probably should not be read only

mdnoble73: If we no longer need the field, it could also be removed.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2191, opened 2025-01-10
DIS-211: Ignore 'number one bestseller' subtitle in OverDrive

LeoStoyanovByWater:
This ensures that promotional subtitles like "number one bestsellers" are filtered out in when searching OverDrive records.


mdnoble73:
Hi @LeoStoyanovByWater could you resubmit this against version 25.02? 

Thanks,

- Mark

PR #2192, opened 2025-01-10
DIS-211: Ignore 'number one bestseller' subtitle in OverDrive

LeoStoyanovByWater:
This ensures that promotional subtitles like "number one bestsellers" are filtered out when searching OverDrive records, now on version 25.02.00.


mdnoble73:
@LeoStoyanovByWater can you also move your release notes from the 25.01 release notes to 25.02? 

Thanks,

- Mark

mdnoble73 (approved):

PR #2196, opened 2025-01-14
DIS-224: Change backup cron job to use mydumper

kylemhall:
This commit changs the default backup script to use mydumper.

Test Plan:
1) Set up a new instance of Aspen using createSite.php 2) Note the cron file uses myDumper for backups now!

### DIFF
--- a/code/web/release_notes/25.02.00.MD
+++ b/code/web/release_notes/25.02.00.MD
@@ -69,6 +69,10 @@
 
 // kyle
 
+### Cron Updates
+- When loading patron types and libraries from Polaris, automatically associate them with the correct account profile. (DIS-214) (*MDN*)
+- Switch default backup script to mydumper. (DIS-224) (*KMH*)
+
 //kidclamp
 
 Use mb_substr to presrve diacritics in lists (DIS-178) (*WNC*)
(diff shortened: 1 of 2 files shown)


sites/template.linux/conf/crontab_settings.txt
```diff
@@ -38,7 +38,7 @@
 # MySQL Nightly Dump #
 ######################
 # backup important bits at 12:10am daily
-10 0 * * *    root    php /usr/local/aspen-discovery/code/web/cron/backupAspen.php {sitename}
+10 0 * * *    root   rm /data/aspen-discovery/{sitename}/sql_backup/mydumper/* && php /usr/local/aspen-discovery/code/web/cron/backupAspen-mydumper.php {sitename}
```
mdnoble73, line 41: Mydumper is not installed by default so this won't work on a clean server. The pattern is also to clean up old files within the backup script itself rather than cron. We also should preserve 2 backups on a server. This seems to remove all backups.

code/web/release_notes/25.02.00.MD
```diff
@@ -69,6 +69,10 @@
 
 // kyle
 
+### Cron Updates
+- When loading patron types and libraries from Polaris, automatically associate them with the correct account profile. (DIS-214) (*MDN*)
```
mdnoble73, line 73: Looks like this has extra information in it from another update. Please don't duplicate the extra release note entry.

mdnoble73 (changes requested):

mdnoble73:
Closing per conversation with Leo who said ByWater no longer is interested in this.

PR #2198, opened 2025-01-15
25.02.00 newsletter consent koha integration

Chloe070196:
Test plans in commit messages
Required Koha plugin: https://github.com/PTFS-Europe/koha-plugin-newsletter-consent/releases/tag/v1.0.2
Jira ticket: https://aspen-discovery.atlassian.net/browse/DIS-98


install/aspen.sql
```diff
@@ -2779,6 +2779,7 @@ CREATE TABLE `library` (
   `eCommerceFee` varchar(11) DEFAULT '0',
   `eCommerceTerms` mediumtext DEFAULT NULL,
   `cookieStorageConsent` tinyint(1) DEFAULT 0,
+  `ilsConsentEnabled` tinyint(1) DEFAULT 0,
```
mdnoble73: @Chloe070196 this will need a database update rather than a direct edit of the database script so existing sites will get upgraded.

code/web/sys/LibraryLocation/Library.php
```diff
@@ -909,6 +912,20 @@ static function getObjectStructure($context = ''): array {
 			$validCardRenewalOptions[3] = 'Quipu eRenewal';
 		}
 
+		// prevent administrators from enabling the ILS consent integration if their ILS does not allow for it
+		$driverAllowsForIlsConsentTypes = true;
+		if (
+			$catalog == null ||
+			$catalog->driver == null ||
```
mdnoble73: Rather than checking if the method exists, we've been adding a method to the CatalogConnection that we can check to see if a particular feature is enabled. Then AbstactIlsDriver returns false and the ILS(s) where it is enabled will return true.

code/web/Drivers/Koha.php
```diff
@@ -8377,4 +8395,194 @@ protected function getUserMessageTranslation($code, User $patron): array {
...
+		} else {
+			return translate([
+				'text' => 'Error getting a list of consents for this patron from Koha.',
+				'isPublicFacing' => true,
+			]);
+		}
+
+	}
+
+	private function getConsentTypes() {
```
mdnoble73: @Chloe070196 is there a proactive way to check to see if the plugin is installed or do we just have to check for an error?

mdnoble73 (changes requested):

install/aspen.sql
Chloe070196: Thank you for pointing this out, I'll submit a commit to address that

code/web/sys/LibraryLocation/Library.php
Chloe070196: Thank you for the pointers on this, I will write a commit adding an `hasIlsConsentSupport` method on the relevant classes then, and remove the current check

code/web/Drivers/Koha.php
Chloe070196: Yes, there is. Aspen can query the Koha db to do just that. Happy to sort out a method to implement this so that Aspen does not make any calls to the API endpoints that are plugin-only if that plugin is not found.

code/web/services/MyAccount/MyPrivacySettings.php
```diff
@@ -0,0 +1,164 @@
...
+
+require_once ROOT_DIR . '/services/MyAccount/MyAccount.php';
+
+class MyAccount_MyPrivacySettings extends MyAccount {
+	function launch(): void {
+		global $interface;
+		global $library;
+
+		if (!UserAccount::isLoggedIn()) {
+			$interface->assign('error', 'You must be logged in to access the Administration Interface');
```
mdnoble73: This is not an admin interface so this message is not correct. By extending MyAccount, should already be enforced.

code/web/interface/themes/responsive/account-menu.tpl
```diff
@@ -100,7 +100,7 @@
 				</div>
 			{/if}
 			{if !empty($showUserPreferences)}<div class="header-menu-option" ><a href="/MyAccount/MyPreferences">{translate text='Your Preferences' isPublicFacing=true}</a></div>{/if}
-			{if $cookieConsentEnabled} <div class="header-menu-option"><a href="/MyAccount/MyCookiePreferences">{translate text="Your Privacy Settings" isPublicFacing=true}</a></div>{/if}
+			{if $privacyConsentEnabled} <div class="header-menu-option"><a href="/MyAccount/MyPrivacySettings">{translate text="Your Privacy Settings" isPublicFacing=true}</a></div>{/if}
```
mdnoble73, line 103: Is this functionality replacing cookie consent? The reason for removing the MyCookiePreference page is unclear based on the Jira as well as comments on the PR.

code/web/services/MyAccount/MyCookiePreferences.php
mdnoble73, line 1: Is this functionality replacing cookie consent? The reason for removing the MyCookiePreference page is unclear based on the Jira as well as comments on the PR.

mdnoble73 (commented):
Is this functionality replacing cookie consent and/or merging Cookie Consent with the other notifications? The reason for removing the MyCookiePreference page is unclear based on the Jira as well as comments on the PR.

code/web/services/MyAccount/MyCookiePreferences.php
Chloe070196: [ce43a8a](https://github.com/Aspen-Discovery/aspen-discovery/pull/2198/commits/ce43a8aee80986f238b3a38304129d295d2f5564) : the Cookie Consent functionality is preserved. The MyCookiePreference page was effectively renamed to MyPrivacySettings in order to become a one-stop shop for patrons consents as it was already referred to as 'Your Privacy Settings' in patron menus. 

The relevant classes and templates files have been refactored so that the 'Your Privacy Settings' point to the Privacy Settings page (formerly cookie consent only) where patrons can view and edit all consents (which as of now is both the existing cookie consents, and the newsletter consent in Koha as introduced by a recent plugin.)

I hope this clarifies the intent behind the refactor!

code/web/interface/themes/responsive/account-menu.tpl
Chloe070196: please see https://github.com/Aspen-Discovery/aspen-discovery/pull/2198#discussion_r1940753998

code/web/services/MyAccount/MyPrivacySettings.php
Chloe070196: Apologies for the oversight - I'll add a commit to remove this

Chloe070196: Here it is: [476273d](https://github.com/Aspen-Discovery/aspen-discovery/pull/2198/commits/476273d15fc815c81a883d8dd7f6f9ff070f93b9)

PR #2200, opened 2025-01-15
DIS-228 Add stack traces to the default log4j configurations

kylemhall:
Test plan:
1) Create a new instance in your test aspen
2) Note the log4j config files include "%throwable" at the end of the
   pattern lines


mdnoble73:
Closing based on Leo's new PR for DIS-228

PR #2201, opened 2025-01-16
Dis 196 indexing 856 for records with e content

AlexanderBlanchardAC:
https://aspen-discovery.atlassian.net/jira/software/c/projects/DIS/boards/2?selectedIssue=DIS-196


mdnoble73 (changes requested):
Rather than creating a second field for this that could conflict with index856Links (i.e. what do you do if index856 links is off, but index856LinkOnlyIfNoEcontentRecord is on). Let's change index856Links to an enum with the values
0 = Never
1 = Always
2 = Only when no eContent Items are found

PR #2209, opened 2025-01-22
Check the 'run full update' option

lucasmontoya13:


mdnoble73:
@lucasmontoya13 this shouldn't be necessary because the last index timestamps will be blank. What behavior are you trying to correct?

lucasmontoya13:
It could resolve my problem, but i'm thinking that it is not the best place to make changes. I will be back soon with a better thought-out proposal.

PR #2212, opened 2025-01-22
DIS 250: Fixed Usage Tables Not Aggregating Data Properly

LeoStoyanovByWater:
- Added a unique composite index on `(userAgentId, year, month, instance)` for `usage_by_user_agent`.
- Added `$usageByUserAgent->find(true)` before incrementing request counters.
- Added `$usageByIPAddress->find(true)` in various files before DB updates.
- Added update call in `bootstrap.php` to update IP Address request counter.

To test:
1. Open Aspen in a browser or local environment. Click around the site to accumulate user requests.
2. Query the tables to see that before the patch, numerous rows would be created for the same user/IP.
3. Once this patch is applied, the same (user/IP, month, year) row as before should change (e.g., accumulate requests).

### DIFF
--- a/code/web/interface/themes/responsive/Greenhouse/ajaxScheduledUpdateDetails.tpl
+++ b/code/web/interface/themes/responsive/Greenhouse/ajaxScheduledUpdateDetails.tpl
@@ -14,11 +14,11 @@
 		</div>
 		<div class="col-sm-3">
 			<strong>{translate text='Date Scheduled' isAdminFacing=true}</strong>
-			<div class="row"><div class="col-sm-12">{$updateScheduled}</span></div></div>
+			<div class="row"><div class="col-sm-12">{$updateScheduled|date_format:"%D %T"}</span></div></div>
 		</div>
 		<div class="col-sm-3">
 			<strong>{translate text='Date Ran' isAdminFacing=true}</strong>
-			<div class="row"><div class="col-sm-12">{if $updateRan}{$updateRan}{else}{translate text='Not Yet Ran' isAdminFacing=true}{/if}</div></div>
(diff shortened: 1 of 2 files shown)


code/web/Action.php
```diff
@@ -117,6 +117,8 @@ protected function forbidAPIAccess()
 		try{
 			$usageByIPAddress->numBlockedApiRequests++;
 			if (SystemVariables::getSystemVariables()->trackIpAddresses) {
+                // This will match that specific IP/year/month/instance row.
```
mdnoble73: @LeoStoyanovByWater it looks like this is using incorrect indenting. Can you correct to use tabs?

code/web/Action.php
```diff
@@ -117,6 +117,8 @@ protected function forbidAPIAccess()
...
+                $usageByIPAddress->find(true);
```
mdnoble73: @LeoStoyanovByWater this search will need to be done before the number of blocked requests is incremented

install/aspen.sql
```diff
@@ -5732,7 +5732,7 @@ CREATE TABLE `usage_by_user_agent` (
   `numRequests` int(11) NOT NULL DEFAULT 0,
   `numBlockedRequests` int(11) NOT NULL DEFAULT 0,
   PRIMARY KEY (`id`),
-  KEY `userAgentId` (`userAgentId`,`year`,`instance`,`month`)
+  UNIQUE KEY `userAgentId` (`userAgentId`,`year`,`instance`,`month`)
```
mdnoble73: These changes need to be done with a database update rather than by simply modifying the default database

mdnoble73 (changes requested):
In addition to the inline updates requested, this will also need a function in the database updates to aggregate all the existing rows prior to making the unique key

PR #2213, opened 2025-01-22
DIS 250: Fixed Usage Tables Not Aggregating Data Properly

LeoStoyanovByWater:
- Added a unique composite index on (userAgentId, year, month, instance) for usage_by_user_agent.
- Added $usageByUserAgent->find(true) before incrementing request counters.
- Added $usageByIPAddress->find(true) in various files before DB updates.
- Added update call in bootstrap.php to update IP Address request counter.

To test:
1. Open Aspen in a browser or local environment. Click around the site to accumulate user requests.
2. Query the tables to see that before the patch, numerous rows would be created for the same user/IP.
3. Once this patch is applied, the same (user/IP, month, year) row as before should change (e.g., accumulate requests).


kylemhall:
Could you also prefix all the commit messages in this pull request with "DIS 250: "? Thanks!

code/web/Action.php
```diff
@@ -114,7 +114,9 @@ protected function forbidAPIAccess()
 		$aspenUsage->blockedApiRequests++;
 		$aspenUsage->update();
 		global $usageByIPAddress;
-		try{
+		try {
```
mdnoble73, line 117: @LeoStoyanovByWater this should not be needed after your change in bootstrap that will load the row by default.

code/web/bookcover.php
```diff
@@ -27,6 +27,8 @@
 	}
 	if (SystemVariables::getSystemVariables()->trackIpAddresses) {
 		if ($usageByIPAddress->id) {
+			// This will match that specific IP/year/month/instance row.
```
mdnoble73: @LeoStoyanovByWater this also should not be needed if we are looking for the correct row in bootstrap where $usageByIPAddress was initialized

code/web/bootstrap.php
```diff
@@ -178,6 +180,7 @@
 }
 $usageByIPAddress->lastRequest = time();
 $usageByIPAddress->numRequests++;
+$usageByIPAddress->update();
```
mdnoble73: What is the update doing here. Are you finding cases where the usage was not being incremented?

code/web/index.php
```diff
@@ -1029,6 +1029,8 @@
 
 	if (SystemVariables::getSystemVariables()->trackIpAddresses) {
 		if ($usageByIPAddress->id) {
+			// This will match that specific IP/year/month/instance row.
+			$usageByIPAddress->find(true);
```
mdnoble73: The find has already been done as part of bootstrap (which is what sets the id). This should not be needed.

code/web/index.php
```diff
@@ -1592,6 +1594,8 @@ function trackSpammyRequest() {
 				$ipAddress->update();
 			}
 		}
+		// This will match that specific IP/year/month/instance row.
```
mdnoble73: This lookup is also part of bootstrap so does not need to be repeated

code/web/sys/AspenError.php
```diff
@@ -183,6 +183,8 @@ function handleAspenError() {
 			try {
 				if (SystemVariables::getSystemVariables()->trackIpAddresses) {
 					if ($usageByIPAddress->id) {
+						// This will match that specific IP/year/month/instance row.
```
mdnoble73: This is already done in bootstrap so does not need to be repeated

code/web/sys/DBMaintenance/version_updates/25.02.00.php
```diff
@@ -67,6 +67,39 @@ function getUpdates25_02_00(): array {
...
+				// 4) Re-insert aggregated rows into original table.
+				"INSERT INTO usage_by_user_agent
+				 SELECT userAgentId, year, month, instance, totalRequests, totalBlocked
+				 FROM usage_by_user_agent_temp",
+
+				// 5) Drop the temp table.
+				"DROP TABLE usage_by_user_agent_temp",
+
+				// 6) Finally add the unique index
+				"ALTER TABLE usage_by_user_agent
```
mdnoble73, line 98: We should also drop the original index here so we don't have 2 duplicate keys, one unique and one non-unique

mdnoble73 (changes requested):

code/web/sys/DBMaintenance/version_updates/25.02.00.php
```diff
@@ -67,6 +67,40 @@ function getUpdates25_02_00(): array {
...
+						SUM(numBlockedRequests) AS totalBlocked
+				 FROM usage_by_user_agent
+				 GROUP BY userAgentId, year, month, instance",
+
+				// 3) Clear out original table.
+				"TRUNCATE TABLE usage_by_user_agent",
+
+				// 4) Re-insert aggregated rows into original table.
+				"INSERT INTO usage_by_user_agent
+				 SELECT userAgentId, year, month, instance, totalRequests, totalBlocked
```
mdnoble73: @LeoStoyanovByWater this fails because the columns are numRequests and numBlockedRequests rather than totalRequests and totalBlocked. Those were just temporary column names when you filled the temp table initially.

mdnoble73 (changes requested):

install/aspen.sql
```diff
@@ -5732,7 +5732,7 @@ CREATE TABLE `usage_by_user_agent` (
   `numRequests` int(11) NOT NULL DEFAULT 0,
   `numBlockedRequests` int(11) NOT NULL DEFAULT 0,
   PRIMARY KEY (`id`),
-  KEY `userAgentId` (`userAgentId`,`year`,`instance`,`month`)
+  UNIQUE KEY `userAgentId` (`userAgentId`,`year`,`instance`,`month`)
```
mdnoble73, line 5735: In the future you don't need to modify aspen.sql, that gets built automatically with a script.

mdnoble73 (approved):

PR #2216, opened 2025-01-28
DIS-219: Filter Hold Requests

AlexanderBlanchardAC:
https://aspen-discovery.atlassian.net/jira/software/c/projects/DIS/boards/2?search=219&selectedIssue=DIS-219


code/web/release_notes/25.02.00.MD
```diff
@@ -125,6 +125,9 @@
 - Eliminate the time portion of display dates from Aspen Open Archives search results (DIS-242) (*JStaub*)
 
 //alexander
+### Holds Updates
```
mdnoble73, line 128: @AlexanderBlanchardAC this needs the ticket number in parenthesis before your initials for each change.

mdnoble73 (changes requested):
Can you fix release notes? Also the name of the Pull Request should be DIS-219 Filter hold requests with the dash so that Jira can link properly. 

Thanks

code/web/release_notes/25.02.00.MD
AlexanderBlanchardAC: Sorry Mark, missed that. Updated the release notes.

mdnoble73 (changes requested):
Hi Alexander, 
This development does not follow the design patterns that are already established within Aspen with regards to Linked Accounts. In other locations we have the ability to select the active Linked User or all are shown. This type of filter should follow that pattern. 

The input to seelect which user to select should be above the tabs (Libby, Boundless, Physical, etc).  so that it applies to all tabs without the user having to reselect as they move between tabs and so they don't need to set it for available vs. pending separately. 

This functionality should also be included in checked out titles for consistency. 

Let me know if you have questions or want to discuss more. 

- Mark

AlexanderBlanchardAC:
> Hi Alexander, This development does not follow the design patterns that are already established within Aspen with regards to Linked Accounts. In other locations we have the ability to select the active Linked User or all are shown. This type of filter should follow that pattern.
> 
> The input to seelect which user to select should be above the tabs (Libby, Boundless, Physical, etc). so that it applies to all tabs without the user having to reselect as they move between tabs and so they don't need to set it for available vs. pending separately.
> 
> This functionality should also be included in checked out titles for consistency.
> 
> Let me know if you have questions or want to discuss more.
> 
> * Mark

Hi Mark, 
Please could I have an example of where we filter by linked users so I know what I am aiming at?
Thanks, 
Alexander

mdnoble73:
Look for cases that use the switch-linked-user-form.tpl template. It looks like we deprecated most of those cases, but that allows the user to change back and forth between which account is active. By default you would want to show all accounts so you would likely need to extend the template to optionally include all accounts.

AlexanderBlanchardAC:
> Look for cases that use the switch-linked-user-form.tpl template. It looks like we deprecated most of those cases, but that allows the user to change back and forth between which account is active. By default you would want to show all accounts so you would likely need to extend the template to optionally include all accounts.

Cheers Mark, 
The partner wants to be able to see the holds of multiple accounts at once, while also filtering out the holds of some linked users which would mean changing this form to use multiselect, is that going to cause issues in the RosenLevelUp or if any of the depreicated use cases are reinstated in the future?

mdnoble73:
Do you have more information on the request, there isn't a lot of info in the Jira ticket. Is the use case that I want to see my holds as well as the holds of my child, but not the holds for my partner?

If that is the case, having filters at the top where patrons could choose the accounts they want to see (rather than select the ones to hide) would be better. Similar to what we do for materials requests. Then they would default to seeing all accounts in this case.

AlexanderBlanchardAC:
> Do you have more information on the request, there isn't a lot of info in the Jira ticket. Is the use case that I want to see my holds as well as the holds of my child, but not the holds for my partner?
> 
> If that is the case, having filters at the top where patrons could choose the accounts they want to see (rather than select the ones to hide) would be better. Similar to what we do for materials requests. Then they would default to seeing all accounts in this case.

Hi Mark, 
Yes, that use case would be correct. I have added a specification PDF to the Jira ticket. (I will update this tomorrow to the new filtering logic - include rather than hide).

AlexanderBlanchardAC:
Hi Mark, I have moved the filter to the top of the page and reversed the logic so that it filters in rather than filtering out. By default the page shows the holds of the main user and all linked users. I have been asked to open a separate ticket for adding this functionality to checked out titles.

PR #2219, opened 2025-01-29
DIS-192: Native Events (Part 1 - admin interface for adding events)

Abhyastamita:
- Added new permissions for events (some of which don't do anything yet)
- Added new settings options to administer Event Fields, Event Field Sets, Event Types, and Events
- Added new tables: event, event_event_field, event_field, event_field_set, event_field_set_field, event_instance, event_type, event_type_library, event_type_location
- Added Is Valid Event Location option to sublocation (but this doesn't do anything yet)
- This is very much a work in progress and lots of things don't work correctly yet (permissions, rules for being able to delete things, some form behavior, and importantly the events don't do anything yet once you add them).  But a user should be able to:
  - create events fields
  - add them to an event field set
  - pick one of the fields sets to use while adding an event type
  - create an event and use one of the event types and see all the correct options based on the event type/field set/fields they already set up
  - Choose a recurrence pattern for the event (e.g. 3rd Wednesdays of every month) and have event instances generated on the correct dates for that recurrence pattern
  - View and manage the event instances created for each event


Abhyastamita:
- Added indexer and record driver. 
- Also added System Variable setting to enable and disable Aspen Events.  
- Started renaming Native Events to Aspen Events and reorganized the admin menu to separate out Aspen Events from the other third party events settings.  
- Uploaded event covers now display on the event record, however there are still some changes that need to be made for this - one is that uploaded files should get renamed so they don't overwrite each other.

mdnoble73 (approved):

PR #2220, opened 2025-02-01
DIS-309: Filter student checkout report by system

NashvillePublicLibrary:


code/web/interface/themes/responsive/Report/studentReport.tpl
```diff
@@ -241,15 +242,62 @@
 		</table>
 		<script type="text/javascript">
 			{literal}
-				$(document).ready(function(){
-					$('#studentReportTable').tablesorter({
-						widgets: ["filter"],
-						widgetOptions: {
-							filter_hideFilters : false,
-							filter_ignoreCase: true
+			$(document).ready(function() {
```
mdnoble73, line 245: @NashvillePublicLibrary tablesorter based filtering does not work well if there is more than one page of results and can be a bit slow with a large number of results. It does not look like the report can have multiple pages so that doesn't seem to be an issue. I'm not sure how many rows normally exist within the table. 

This could also be implemented similarly to the showOverdueOnly filter, but this approach should work as well based on my understanding of your use case

NashvillePublicLibrary: @mdnoble73 The Student Report as-is will only render one page. It usually contains hundreds of records.

I wanted to use tablesorter rather than an approach that makes an additional call to the CarlX database for end-user speed sake.

I don't think you mention anything above that should stop my PR from being accepted. If a change is actually needed, just tell me again!

mdnoble73: Nope nothing blocking merge, just wanted to point out some of the potential pros/cons.

mdnoble73 (approved):

PR #2221, opened 2025-02-02
DIS-244 Carl.X: Lost items should appear in Checked Out page

NashvillePublicLibrary:
+ add property ilsStatus to Checkout object
+ add column ilsStatus to user_checkout table
+ alter Lost records on /MyAccount/CheckedOut to display "Pay Fee Online" button linked to /MyAccount/Fees
+ alter Lost records on /MyAccount/CheckedOut to display "Lost" instead of "Overdue"

### DIFF
--- a/code/web/interface/themes/responsive/MyAccount/ilsCheckedOutTitle.tpl
+++ b/code/web/interface/themes/responsive/MyAccount/ilsCheckedOutTitle.tpl
@@ -1,3 +1,4 @@
+{debug}
 {strip}
 	<div id="record{$record->source}_{$record->sourceId|escape}" class="result row{if $record->isOverdue()} bg-overdue{/if}">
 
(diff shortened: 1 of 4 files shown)


code/web/interface/themes/responsive/MyAccount/ilsCheckedOutTitle.tpl
```diff
@@ -250,6 +251,9 @@
 									{translate text="Sorry, this title cannot be renewed" isPublicFacing=true}
 								{/if}
 							{/if}
+							{if !empty($record->ilsStatus) && $record->ilsStatus == 'Lost'}
```
mdnoble73: @NashvillePublicLibrary I'm thinking we should add a property to the checkout called showFineLink that could be set within the driver rather than having the logic within the template. That will make it easier to add this functionality for other ILSs that have different ways of indicating that a title is lost and has accumulating fines.

mdnoble73:
@NashvillePublicLibrary see the comment within the code. This also needs release notes. 

Thanks,
- Mark

mdnoble73 (approved):

PR #2224, opened 2025-02-04
DIS-179: Fixed Duplicated Titles in Reading History

LeoStoyanovByWater:
- Added a SQL query to delete old duplicates and to prevent future duplicates by creating a unique index.
- Modified `CatalogConnection.php` to prevent duplicates during checkout-based history updates.
- Updated `AJAX.php` to check `userId` and `sourceId` instead of `groupedWorkPermanentId` when adding manual entries.

The cron job `UpdateReadingHistory.java` eventually calls `updateReadingHistoryBasedOnCurrentCheckouts()` in `CatalogConnection.php`, which has new validation to prevent duplicates. This means the cron job should not create any duplicates as well.

To test:
1. Checkout a couple of items for a patron from an ILS (e.g., Koha).
2. Force reload the reading history for that patron.
3. On the patron's reading history page, and in the `user_reading_history_work` table, there will likely be duplicated items.
4. Apply the patch, run the appropriate DB maintenance update, and force reload the patron's reading history.

### DIFF
--- a/code/web/CatalogConnection.php
+++ b/code/web/CatalogConnection.php
@@ -664,7 +664,11 @@ class CatalogConnection {
 									$userReadingHistoryEntry->isIll = 1;
 								}
 								$userReadingHistoryEntry->deleted = 0;
-								$userReadingHistoryEntry->insert();
+								if ($userReadingHistoryEntry->find(true)) {
+									$userReadingHistoryEntry->update();
+								} else {
+									$userReadingHistoryEntry->insert();
+								}
 								$userReadingHistoryEntry = null;
 								//}
 							}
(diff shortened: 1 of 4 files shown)


code/web/sys/DBMaintenance/version_updates/25.02.00.php
```diff
@@ -156,6 +156,25 @@ function getUpdates25_02_00(): array {
...
+			'description' => 'Combine multiple rows per (userId, sourceId) into one row before adding unique key.',
+			'continueOnError' => true,
+			'sql' => [
+				// 1. Keep the latest entry of the duplicated titles.
+				"DELETE t1 FROM user_reading_history_work t1
+				INNER JOIN user_reading_history_work t2 
+				WHERE 
+					t1.id < t2.id AND
+					t1.userId = t2.userId AND
+					t1.sourceId = t2.sourceId;",
```
mdnoble73, line 171: @LeoStoyanovByWater this will also need to check the source in addition to the sourceId since you could have record id 1 from the ILS as well as event 1.

code/web/services/MyAccount/AJAX.php
```diff
@@ -7444,12 +7444,9 @@ function saveToReadingHistory(): array {
 						$readingHistoryEntry->isIll = 0;
 						$readingHistoryEntry->isManuallyAdded = 1;
 						//No cost savings updates since this is outside the library
+						$readingHistoryEntry->userId = $user->id;
```
mdnoble73, line 7447: @LeoStoyanovByWater these are already set above so they don't need to be set again

mdnoble73: If we're seeing duplicates while manually adding titles, we probably need to do the find prior to setting checkout and check in dates

code/web/CatalogConnection.php
```diff
@@ -664,7 +664,11 @@ function getReadingHistory($patron, $page = 1, $recordsPerPage = 20, $sortOption
 									$userReadingHistoryEntry->isIll = 1;
 								}
 								$userReadingHistoryEntry->deleted = 0;
-								$userReadingHistoryEntry->insert();
+								if ($userReadingHistoryEntry->find(true)) {
```
mdnoble73, line 667: @LeoStoyanovByWater if we are seeing duplicates here we probably need to move this check before setting the check in and check out dates as well as possibly ill and deleted. The find will do a query based on all of the properties that are set so we likely will still see some duplicates.

mdnoble73 (changes requested):

PR #2229, opened 2025-02-06
DIS-179: Fixed Duplicated Titles in Reading History

LeoStoyanovByWater:
- Added a SQL query to delete old duplicates and to prevent future duplicates by creating a unique index.
- Modified `CatalogConnection.php` to prevent duplicates by checking `(userId, sourceId, source)` during checkout-based history updates.
- Updated `AJAX.php` to prevent duplicates when adding manual entries.

The cron job `UpdateReadingHistory.java` eventually calls `updateReadingHistoryBasedOnCurrentCheckouts()` in `CatalogConnection.php`, which has new validation to prevent duplicates. This means the cron job should not create any duplicates as well.

To test:
1. Checkout a couple of items for a patron from an ILS (e.g., Koha).
2. Force reload the reading history for that patron.
3. On the patron's reading history page, and in the `user_reading_history_work` table, there will likely be duplicated items.
4. Apply the patch, run the appropriate DB maintenance update, and force reload the patron's reading history.


code/web/sys/DBMaintenance/version_updates/25.03.00.php
```diff
@@ -130,6 +130,29 @@ function getUpdates25_03_00(): array {
 		], //portal_cell_show_hide_description
 
 		// Leo Stoyanov - BWS
+		'deduplicate_reading_history' => [
+			'title' => 'De-Duplicate Reading History & Add Unique Key',
+			'description' => 'Combine multiple rows per (userId, sourceId, source) into one row before adding unique key.',
+			'continueOnError' => true,
+			'sql' => [
+				// 1. Keep the latest entry of the duplicated titles.
+				"DELETE t1 FROM user_reading_history_work t1
```
mdnoble73, line 176: @LeoStoyanovByWater this does not account for cases where source and sourceId are both blank. In that case we should look for duplicates based on groupedWorkId to avoid deleting different grouped works. We may want to add groupedWorkId has part of the query in all cases.

mdnoble73 (changes requested):

code/web/sys/DBMaintenance/version_updates/25.03.00.php
mdnoble73: This also should account for setting the initial check out date

ghost: I don't understand why initial checkout date is relevant here.

mdnoble73 (changes requested):
Let's talk through this more. The previous code was allowing an entry to be stored in the database multiple times. So we would have a complete picture of when a patron used a title. This pull request only stores the title once which is similar to how we display things to the patron. But loses the granularity of knowing when things were checking out multiple times. 

I think we probably need to break this into 2 tables one to store grouped work usage which would be unique for the patron and groupedWorkPermanentId and then a second table which stores the checkout history which would be the source, sourceId, format, checkOutDate, and checkInDate.

Cost savings can then be calculated based on the formats checked out for each groupedWorkPermanentId.

We'll need much more expansive testing notes to be sure that we're covering all the cases for different ILSs. And that we are testing new checkouts, checkouts of titles checked out previously, manually added titles etc. Some of those tests will probably need to be verified in the database.

PR #2230, opened 2025-02-06
DIS-301: Added Option to Use Original Cover URLs

LeoStoyanovByWater:
- Added an option under System Variables to allow libraries to enable the use of the direct cover URLs, rather than saving and pulling cover images from the file system.
- Improved image fetching by bypassing the logic to build the cover image URL if the URL has already been built and relevant cover information has not changed.

To test:
1. Navigate to Aspen Administration -> System Variables under System Administration -> Click to edit whichever library system -> Scroll to the bottom of the page -> Enable "Use Original Cover URLs."
2. Clear browser cache.
3. Navigate to a page that displays cover images (e.g., browse page).
4. In the browser's inspect mode, cover images have the same _src_ as before, but if you right click and open the image in a new tab, it is being pulled directly from its appropriate provider.

EDIT: This is a preliminary improvement, a step before a potentially larger change to completely avoid using `bookcover.php`.


mdnoble73:
@LeoStoyanovByWater It looks like this doesn't do any checking to see if the URL gives back a valid response so if the library has a chain of book cover sources setup this will only use the first one always. We would want to accommodate the existing chain even if the full cover is not being cached. This also does not handle cases where no image is available from any source.

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -2054,4 +2064,101 @@ protected function searchOmdbForCover($year, string $encodedShortTitle, OMDBSett
...
+	 */
+	private function initUrlCache(): void
+	{
+		// Create a subdirectory "urlCache" under the main cover path.
+		$urlCachePath = $this->bookCoverPath . '/urlCache';
+		if (!is_dir($urlCachePath)) {
+			mkdir($urlCachePath, 0777, true);
+		}
+		// Use a hashed key for the file name so that the file system has fewer characters to work with.
+		$this->urlCacheFile = $urlCachePath . '/' . md5($this->cacheName) . '.txt';
```
mdnoble73: Why store these as files rather than as entries in the database?

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -2054,4 +2064,101 @@ protected function searchOmdbForCover($year, string $encodedShortTitle, OMDBSett
...
+	/**
+	 * If the system is configured to use the original cover URLs and the source is not an upload,
+	 * check for a cached URL in the file system (using the cache file initialized by initUrlCache()),
+	 * write it if not present, and then (if the URL appears to be valid and not a dummy) immediately issue a 301 redirect.
+	 *
+	 * @param string $source The source label.
+	 * @param string $url The URL to process.
+	 * @return string Returns the (possibly unmodified) URL.
+	 */
+	private function handleOriginalCoverUrl(string $source, string $url): string {
```
mdnoble73, line 2245: This will always store and return the first URL that is checked rather than attempting to see if a valid file is returned (many sources do not return a valid image). Storing the URL rather than the actual image could be ok, but it should not circumvent the logic to find a good image for the work.

mdnoble73 (changes requested):

code/web/sys/Covers/BookCoverProcessor.php
ghost: Reading a file tends to be less overhead than SQL queries for each lookup and the URLs themselves would rarely change (if ever), so caching a file and looking it up is more efficient. Each file is also now a much smaller size because it holds one line, the URL, so it will also utilize much less storage than before.

mdnoble73: @LeoStoyanovByWater we're already doing the database lookup though so the call to the file system is an extra check than just adding the original URL to the book cover table.

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -2096,4 +2107,165 @@ protected function searchOmdbForCover($year, string $encodedShortTitle, OMDBSett
...
 		return $foundTitle;
 	}
+
+	/**
+	 * Validates a cover URL to ensure it returns a valid, usable image
+	 *
+	 * @param string $url The URL to validate
+	 * @return bool True if the URL is valid and returns a usable image, false otherwise
+	 */
+	private function validateCoverUrl(string $url, string $source): bool {
```
mdnoble73: @LeoStoyanovByWater with this extra validation the new method seems to be more expensive in terms of bandwidth and processing than the original method. This new method first gets the headers and then downloads the image and validates, but then returns the original url to the browser so it can be downloaded again. The amount of storage on disk is reduced though.

Previously we validated the image once and then cached it. So on the first call there was one call to fetch the image. And then we returned the cached data to the user. 

Now, there are 3 calls to the remote server each time an image is displayed, the first gets the header, the second gets the full image, and the third is from the user's browser to fetch the image. 

If we want to store the URL I think we'd want to assume it is correct when using the cached version, but make sure to expire the cache and make sure that the reload cover option works.

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -2130,4 +2141,185 @@ protected function searchOmdbForCover($year, string $encodedShortTitle, OMDBSett
...
+
+		// Check the color of the bottom left and bottom right corners.
+		if ($imageResource = @imagecreatefromstring($image)) {
+			$width = $imageInfo[0];
+			$height = $imageInfo[1];
+			$rgbLeft = imagecolorat($imageResource, 0, $height - 1);
+			$rgbRight = imagecolorat($imageResource, $width - 1, $height - 1);
+			imagedestroy($imageResource);
+			if ($rgbLeft == 8421504 && $rgbRight == 8421504) {
+				$this->log("Cover URL $url returned an image with partial gray at the bottom; falling back.", Logger::LOG_ERROR);
```
mdnoble73: Do we need this level of logging or can this be LOG_DEBUG?

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -2130,4 +2141,185 @@ protected function searchOmdbForCover($year, string $encodedShortTitle, OMDBSett
...
+				// Force validation if reload flag is set
+				$forceValidation = $this->reload;
+				$validationResult = $this->validateCoverUrl($url, $this->bookCoverInfo->imageSource, $forceValidation);
+				if ($validationResult !== false) {
+					header("HTTP/1.1 301 Moved Permanently");
+					header("Location: $url");
+					$this->addCachingHeader();
+					exit;
+				} else {
+					$this->log("URL validation failed for stored URL: $url", Logger::LOG_ERROR);
```
mdnoble73: Do we need this level of logging or can this be LOG_DEBUG?

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -2130,4 +2141,185 @@ protected function searchOmdbForCover($year, string $encodedShortTitle, OMDBSett
...
+				}
+			} else {
+				$this->log("Debug - Hash validation failed: " . ($currentHash === $storedHash ? "Hashes match" : "Hashes don't match") . ", URL empty: " . (empty($url) ? "Yes" : "No"), Logger::LOG_ERROR);
```
mdnoble73: Do we need this level of logging or can this be LOG_DEBUG?

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -2130,4 +2141,185 @@ protected function searchOmdbForCover($year, string $encodedShortTitle, OMDBSett
...
+			}
+		} else {
+			$this->log("Debug - Early conditions failed: BookCoverInfo exists: " . ($this->bookCoverInfo ? "Yes" : "No") .
```
mdnoble73, line 2332: Do we need this level of logging or can this be LOG_DEBUG?

mdnoble73 (changes requested):
With this logic it looks like anything that has a default cover will never have any sort of caching so if a title uses a default cover, the full logic will run every time which is expensive. 

Also check logging to ensure that LOG_ERROR is not being used when LOG_DEBUG should be

PR #2232, opened 2025-02-07
DIS-338: Validation for Uploading Images in Web Builder

LeoStoyanovByWater:
- Introduced frontend and backend validation for uploading images with a note displaying the valid file types.

The backend validation also closely mirrors how regular file validation is done in the same `DataObjectUtil.php`.

To test:
1. Navigate to Aspen Administration → Web Builder → Images and upload an invalid file type. The page will output a SQL error at the top, and the save will fail.
2. Apply the patch and try to upload an invalid file type again. An alert will stop and notify you that the file type is invalid.
3. If you remove the JS snippet in `property.tpl` to test backend validation, upload another invalid image, and you will be directed to an error page.


code/web/sys/DataObjectUtil.php
```diff
@@ -348,12 +348,14 @@ static function processProperty(DataObject $object, $property, $fieldLocks) {
 				} elseif (isset($_FILES[$propertyName]["error"]) && $_FILES[$propertyName]["error"] > 0) {
 					//return an error to the browser
 					$logger->log("Error in file upload for $propertyName", Logger::LOG_ERROR);
-				} elseif (in_array($_FILES[$propertyName]["type"], [
-					'image/gif',
-					'image/jpeg',
-					'image/png',
-					'image/svg+xml',
-				])) {
+				} elseif (!empty($property['validTypes']) && !in_array($_FILES[$propertyName]["type"], $property['validTypes'])) {
```
mdnoble73: This removes checking of file type for anything where validTypes are not set for image properties of which there are several. Handling of those default cases need to be considered while adding the ability to define specific allowable types.

ghost: It's now been a while, but I believe I removed that check for defaults because it was ineffective in preventing invalid image types from being uploaded. It's also not clear that the default check was ever being used anywhere else in the codebase. Nonetheless, I'll add a default fallback.

mdnoble73: Thanks. It looks like the behavior before was to fail silently if it was the wrong file type. Would be great if there was a message.

ghost: Is the new commit sufficient with the Aspen Error? This check is the "last line of defense" if the user somehow bypasses the frontend JS validation.

PR #2238, opened 2025-02-12
DIS-250: Fixed User Usage Checks & Added Atomic SQL Updates

LeoStoyanovByWater:
- Added checks about whether a user usage agent exists; if it does not, insert one.
- Implemented atomic SQL updating of user usage agent requests and blocked requests.

### DIFF
--- a/code/web/bootstrap.php
+++ b/code/web/bootstrap.php
@@ -64,6 +64,9 @@ ob_start();
 
 initMemcache();
 initDatabase();
+$timer->logTime("Initialized Database");
+requireSystemLibraries();
+initLocale();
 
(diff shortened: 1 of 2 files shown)


mdnoble73:
@LeoStoyanovByWater can you put this into 25.02 since we're in the testing period and it's directly related to the code changes for 25.02? 

Also for the functions, I was thinking that you could put them on the UsageByUserAgent class. So you would call them with $usageByUserAgent->incrementUserUsageBlockedRequests()

We also probably need the same thing for incrementing UsageByIPAddress since that would have similar issues if requests come in at the same time.

PR #2239, opened 2025-02-13
DIS-250: Fixed User Usage Checks & Added Atomic SQL Updates

LeoStoyanovByWater:
- Added checks about whether a user usage agent exists; if it does not, insert one.
- Implemented atomic SQL updating of user usage agent requests and blocked requests.
- Moved the initialization of local time in `bootstrap.php` to occur before any usage table operations.

### DIFF
--- a/code/web/bootstrap.php
+++ b/code/web/bootstrap.php
@@ -192,13 +192,13 @@ $usageByIPAddress->month = date('n');
 $usageByIPAddress->ipAddress = IPAddress::getClientIP();
 $usageByIPAddress->instance = $aspenUsage->getInstance();
 
-$usageByIPAddress->lastRequest = time();
-if ($usageByIPAddress->find(true)) {
-	$usageByIPAddress->incrementNumRequests();
-} else {
-	$usageByIPAddress->numRequests = 1;
-	$usageByIPAddress->insert();
+try {
+	$usageByIPAddress->find(true);
+} catch (Exception $e) {
+	//Table has not been created yet, ignore it
 }
+$usageByIPAddress->lastRequest = time();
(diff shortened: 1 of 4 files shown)


code/web/bootstrap.php
```diff
@@ -192,7 +207,12 @@
 	try {
 		$usageByIPAddress->numBlockedRequests++;
 		if (SystemVariables::getSystemVariables()->trackIpAddresses) {
-			$usageByIPAddress->update();
+			if ($usageByIPAddress->find(true)) {
```
mdnoble73: @LeoStoyanovByWater there is another call to update requests for the IP around lines 1030-1036 which needs to be converted to incrementing request

code/web/bootstrap.php
```diff
@@ -173,17 +192,13 @@
...
-} catch (Exception $e) {
-	//Table has not been created yet, ignore it
-}
 $usageByIPAddress->lastRequest = time();
-$usageByIPAddress->numRequests++;
-
-$timer->logTime("Initialized Database");
-requireSystemLibraries();
-initLocale();
+if ($usageByIPAddress->find(true)) {
```
mdnoble73: This is also done later in the file (lines 1030-1036). This code is not needed if we update the code later.

mdnoble73: We also will need to account for incrementing the last request as part of incrementing the counter atomically

mdnoble73 (changes requested):
@LeoStoyanovByWater a couple more changes for IP Addresses, thanks! This is looking really good.

mdnoble73 (approved):

PR #2241, opened 2025-02-14
DIS-364: Added Image Source Check for Uploaded List Covers

LeoStoyanovByWater:
- Added check to `BookCoverProcessor::getUploadedListCover()` to make sure the list's cover is actually from an 'upload' source.

Note to anyone who stumbles upon this ticket/pull request:
After this patch has been applied, if there is an existing list that incorrectly uses a cover from a record that has the same ID, you must SQL query to change the `imageSource` of the existing list to `default` (e.g., `UPDATE bookcover_info SET imageSource = 'default' WHERE recordType = 'list' AND recordId = '664'`). This is because `BookCoverProcessor.php` will now properly check the source of the list's book cover before assuming its source and thus create a default cover image. Any future list creations that happen to have the same ID as an existing record will not need this additional step because of this patch.


mdnoble73 (changes requested):
This is a partial fix, but does not preclude uploaded list images from conflicting with records. If you have a record with an ID of 3 and a list with an ID of 3, both will be uploaded to 3.png. When uploading images we should prefix them with the word list_ so the ILS record would be 3.png and the list would be list_3.png. Then we'll need to review the covers that have already been loaded to see if they should be lists or records.

ghost:
@mdnoble73 Although there is an easy fix to have new uploaded list images and record images be prefixed accordingly (along with backwards compatibility, but wouldn't be fail proof), I have no idea how we would determine if existing uploaded cover images are for lists or records, even with a migration script created.

PR #2244, opened 2025-02-18
DIS-384: Added Required and Default Grouped Work Format String

LeoStoyanovByWater:
- Upon a new install, a default format string entry is inserted into the `grouped_work_format_sort_group` table.
- The Format String in a Grouped Work Display setting is now required, has a description, and is automatically filled with the default entry.


mdnoble73 (changes requested):
Do we need existing sites that do not have grouped work format sorting to be handled? 

This also doesn't implement everything described in the Jira ticket which is probably fine, but a full explanation of what was actually implemented is appropriate.

ghost:
@mdnoble73 I don't believe existing sites will need any more handling than this patch provides because the field is now required, so they have to create a grouped work format string before attempting to save on the grouped work display setting page.

I showed Esther, who created the issue, the simple fix, and she approved of it. In terms of "why the Grouped Work Format Sorting is not a part of the Grouped Work Display settings proper" from the Jira issue, that's beyond the scope of my understanding. That is why I focused on the main problem of how the field was required, but not labeled as such, and convenience (i.e., default to an automatically created grouped work format string unless the user wishes to create and choose his own).

mdnoble73:
It's separate because it's much more than a label once you get into the actual settings and they can be shared between different grouped work settings. The scope of what you added looks good to me, thanks.

mdnoble73 (approved):

PR #2250, opened 2025-02-19
DIS-397 Add logging to INN-Reach results fetch method

tomascohen:
This patch adds

* logging of the external request made to INN-Reach, through `ExternalRequestLogEntry::logRequest`
* a check for the HTTP response status code, and if it is not 200, an error log will be generated

Test plan:
1. Enable the INN-Reach integration
2. Have random credentials for the OAuth2 section
3. Perform a search
4. Make sure your IP gets external requests logged => An error log should be generated
=> There's a log on the external requests logs


tomascohen:
For the record, I ran the QA bash script lines locally and cannot reproduce the error tabs vs. spaces error. Any advise is welcome.

mdnoble73 (approved):

PR #2251, opened 2025-02-19
Dis 403 preferred pickup branch ignored

lathomas64:
to test
1. enable setting preferred pickup locations for users
2. search for a record with items at several locations.
3. set your preferred pickup location to one of the locations of the item
4. the item at the preferred pickup location should now be in in the top 3'
5. confirm this behavior is still preserved for alternate pickup locations 1 and 2


mdnoble73 (changes requested):
@lathomas64 this looks like it will work, but it will be confusing to future developers that the preferred pickup location has been substituted for home location. It also means that nothing will sort according to the patron's home location which for some libraries is probably good, but it will be unexpected for others. 

Could you update the Jira ticket with more information on the original problem as well as the fix and update the release notes to describe the new behavior so libraries can test properly. 

Thanks
- Mark

lathomas64:
added extra clarification to the release notes.

mdnoble73 (approved):

PR #2253, opened 2025-02-20
Allow ils authentication

lucasmontoya13:


code/web/sys/UserAccount.php
```diff
@@ -555,7 +555,7 @@ public static function login($validatedViaSSO = false) {
 			/** @var AccountProfile $accountProfile */
 			$accountProfile = $driverData['accountProfile'];
 			$okToAuthenticate = false;
-			if ($accountProfile->id == $library->accountProfileId || $accountProfile->authenticationMethod == 'db') {
+			if ($accountProfile->id == $library->accountProfileId || $accountProfile->authenticationMethod == 'db' || $accountProfile->authenticationMethod == 'ils') {
```
mdnoble73, line 558: There should only be one account profile for the associated library with an authentication method of ILS. Did you intend this to be authenticationMethod == 'sso'?

mdnoble73 (changes requested):

code/web/sys/UserAccount.php
lucasmontoya13: Actually I'm not able to login as a Koha patron. I realize that the patronLogin in Koha.php is ignored in the flux of the code when i try to logging in. It seems like if authentication method is not 'ils' then a user from Koha is not 'okToAuthenticate'. I'm not sure if i'm missing something, but the use of the authentication method 'ils' was intentional. I'm sure there we could take another way

mdnoble73: Is your library attached to an account profile?

lucasmontoya13: In the account profile settings, adb sets a new one, but the checkboxes for the libraries are not marked, so i assume that the libraries aren't attached to an account profile

mdnoble73:
@lucasmontoya13 can you check this commit to see if it meets your needs? https://github.com/Aspen-Discovery/aspen-discovery/commit/86ea0f79b8da0c74d79acaba7e857680b58831f0

lucasmontoya13:
Ok! I'll check that! Thanks

lucasmontoya13:
No, It didn't work. I will collect more information about the bug

mdnoble73:
Does your library have an account profile attached to it?

lucasmontoya13:
In the account profile settings, adb sets a new one, but the checkboxes for the libraries are not marked, so i assume my library is not attached to an account profile

mdnoble73:
If you attach it to an account profile does login work?

lucasmontoya13:
Yes, it is working now. I'll close the PR! Thank you Mark

PR #2257, opened 2025-02-21
Dis 239 evergreen username and email update

Chloe070196:
Enable patrons to update their Evergreen username and email address on Aspen Discovery.

Jira ticket: [DIS-239](https://aspen-discovery.atlassian.net/browse/DIS-239)


code/web/Drivers/Evergreen.php
```diff
@@ -778,11 +778,69 @@ function changeHoldPickupLocation(User $patron, $recordId, $itemToUpdateId, $new
...
+		 * is incorrect.) However, a successful update will always result
+		 * in the "payload" property being set to 1. Check for this instead.
+		*/
+		$responseData = json_decode($apiResponse, true);
+		$success = isset($responseData['payload']) && $responseData['payload'][0] == 1;
+		return $success;
+	}
+
+
+	private function isDuplicateUserName($username):bool {
```
mdnoble73, line 837: @Chloe070196 This will only look at users that have logged into Aspen before which will still mean the username within Evergreen can be duplicated.

code/web/Drivers/Evergreen.php
```diff
@@ -778,11 +778,69 @@ function changeHoldPickupLocation(User $patron, $recordId, $itemToUpdateId, $new
...
+		global $aspen_db;
+		$condition = "ils_username=" . "'" . $username . "'";
```
mdnoble73, line 839: @Chloe070196 this has a potential SQL injection. You should use the standard data object methods with the count function

mdnoble73 (changes requested):

PR #2259, opened 2025-02-21
DIS-401: Added Sierra Statgroup Parameter to API Payment Requests

LeoStoyanovByWater:
Added `statgroup` parameter to Sierra API payment requests using active branch location, library's main location, or patron's home library as per configuration.

This was tested on a production server with a partner who requested the change. Follows [Sierra's API documentation](https://www.saclibrarycatalog.org/iii/sierra-api/swagger/index.html#!/patrons/Pay_a_patron_s_fines_put_28).


code/web/Drivers/Sierra.php
```diff
@@ -1912,6 +1912,12 @@ public function completeFinePayment(User $patron, UserPayment $payment) {
 			'payments' => [],
 		];
 
+		// Use patron's home location stat group by default.
+		$homeLocation = $patron->getHomeLocation();
```
mdnoble73: @LeoStoyanovByWater this should respect the libraryToUseForPayments System Variable. Some libraries prefer to have the payments tracked to the active catalog rather than the patron home library.

ghost: From looking around, my best guess was to use the location singleton's active location for the "Active Catalog."

code/web/Drivers/Sierra.php
```diff
@@ -1912,6 +1912,32 @@ public function completeFinePayment(User $patron, UserPayment $payment) {
...
+
+		if ($systemVariables && $systemVariables->libraryToUseForPayments == 1) {
+			// Get active library location using the singleton pattern.
+			$activeLocation = $locationSingleton->getActiveLocation();
+			global $logger;
+			if ($activeLocation) {
+				$paymentLocation = $activeLocation;
+				$logger->log("Using active catalog location {$activeLocation->code} for payments.", Logger::LOG_NOTICE);
+			} else {
+				$paymentLocation = $patron->getHomeLocation();
```
mdnoble73, line 1948: activeLocation will not be set unless the patron is inside a branch (based on IP) or at a branch specific URL. In that case we should look at the global $library and get the default location for the library (based on main location if one is set or first location sorted alphabetically)

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2260, opened 2025-02-21
DIS-197 Prefer ILS Description

LiYanjun19:
Add system preference for "Prefer ILS Description" in Grouped Work Display settings. With this preference selected, grouped work records will display ILS Description over other eContent Description. 

Test plan:
1. Find a record that has eContent and physical records 
2. Apply this pull request 
3. Force Reindex the record in staff view 
4. Go to system administration > Grouped Work Display Settings > Prefer ILS Description
5. Between the changes of record description display in Search Results and Grouped Work page, with "Prefer ILS Description" on/off


LiYanjun19:
Tho I added the ils_description related code in AnodeAPI.php and SearchAPI.php, for the consistency of the description in different cases, I doubt if it's necessary to add them in there.

code/web/services/API/SearchAPI.php
```diff
@@ -3338,7 +3343,11 @@ function searchLite() : array {
 					$items[$recordKey]['author'] = $record['author_display'];
 					$items[$recordKey]['image'] = $configArray['Site']['url'] . '/bookcover.php?id=' . $record['id'] . '&size=medium&type=grouped_work';
 					$items[$recordKey]['language'] = $record['language'][0];
-					$items[$recordKey]['summary'] = $record['display_description'];
+					if ($library->getGroupedWorkDisplaySettings()->preferIlsDescription == 1) {
```
mdnoble73: @LiYanjun19 do we need logic here if there is no ils_description? I.e. records only owned within an eContent collection that do not have physical titles attached?

code/web/services/API/SearchAPI.php
```diff
@@ -2792,7 +2793,11 @@ function getAppSearchResults(): array {
 					$format = $format . ' - ' . $ccode;
 				}
 
-				$summary = utf8_encode(trim(strip_tags($item['display_description'])));
+				if ($library->getGroupedWorkDisplaySettings()->preferIlsDescription == 1) {
```
mdnoble73: @LiYanjun19 do we need logic here if there is no ils_description? I.e. records only owned within an eContent collection that do not have physical titles attached?

code/web/services/API/AnodeAPI.php
```diff
@@ -148,7 +149,9 @@ function getAnodeGroupedWorks($result, $branch) {
 				if (!isset($groupedWorkRecord['image'])) {
 					$groupedWork['image'] = '/bookcover.php?id=' . $groupedWork['id'] . '&size=medium&type=grouped_work';
 				}
-				if (isset($groupedWorkRecord['display_description'])) {
+				if ($library->getGroupedWorkDisplaySettings()->preferIlsDescription == 1 && isset($groupedWorkRecord['ils_description'])) {
```
mdnoble73: @LiYanjun19 do we need logic here if there is no ils_description? I.e. records only owned within an eContent collection that do not have physical titles attached?

code/reindexer/src/org/aspen_discovery/reindexer/AbstractGroupedWorkSolr.java
```diff
@@ -1230,6 +1231,15 @@ void addKeywords(HashSet<String> keywords) {
 		this.keywords.addAll(keywords);
 	}
 
+	void addIlsDescription(String description) {
```
mdnoble73: @LiYanjun19 we probably want to preserve the functionality to check the format category to prefer the "Books" format category and the longest description. That way the audiobook or playaway doesn't take precedence over the book. Right now the first record would always take precedence.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2264, opened 2025-02-24
DIS-192: Aspen Events reporting, Calendar display and printing improvements

Abhyastamita:
- Switched Aspen Events to use minutes instead of hours.  This includes a database update to multiply existing event lengths by 60, but this might not be necessary if no one has been using Aspen Events yet.
- Fixed indexer bug that prevented full indexing from taking place because it was set to find events after the event date cutoff instead of before.
- Fixed bug that was preventing creating new events if they didn't have a sublocation.
- Added Event Reports, which is a graph that displays event hours over time and can be filtered in a variety of ways.
- Added option to display event calendar by week instead of by month.  Updated all event indexers and event schema to index event_week in the same way they already index event_month, year, etc.
- Made improvements to calendar CSS for printing so that the full calendar should be displayed on a single page with appropriate font-sizes for the paper side.  Printing with landscape orientation should always yield a calendar view.  Printing with portrait orientation may yield more of an agenda or list view, which skips dates that have no events to save space.  
  - Calendar printing should probably be tested with a variety of browsers and with a real printer.  I used Print to PDF and tested in Chrome, Firefox, and Safari, all on a Mac, but it wouldn't surprise me if there are different, less ideal results with different different computers and configurations.
- Added new setting, Calendar Display Settings, which so far only allows a user with 'Print Calendars with Header Images' to upload a header image for printing calendars.  If somebody with this permission views the event calendar, they will be able to see the header image and it will be included and sized appropriately when they print.


code/web/interface/themes/responsive/Events/calendar.tpl
```diff
@@ -1,10 +1,19 @@
 {strip}
+{if !empty($loggedIn) && in_array('Print Calendars with Header Images', $userPermissions)}
+	<div class="calendar-header-image">
+		<img src="{if !empty($headerImage)}{$headerImage}{/if}" alt="{translate text=$headerAlt inAttribute=true isPublicFacing=true}" title="{translate text=$headerAlt inAttribute=true isPublicFacing=true}" id="calendar-header">
```
mdnoble73: headerAlt needs a check to make sure it isn't empty (which it will be until calendar settings are created)

code/web/interface/themes/responsive/DataObjectUtil/duration.tpl
```diff
@@ -0,0 +1,27 @@
...
+<fieldset class="form-inline">
+	<div class="form-group col-sm-6">
+		<label for='{$propName}_hours'>
+			Hours:
+		</label>
+		<input type="number" name='{$propName}_hours' id='{$propName}_hours' min='0' class='form-control duration-input' {if !empty($property.required) && (empty($objectAction) || $objectAction != 'edit')}required{/if} {if !empty($property.readOnly)}readonly disabled{/if} {if !empty($property.autocomplete)}autocomplete="{$property.autocomplete}"{/if} value="{$hours}">
+	</div>
+	<div class="form-group col-sm-6">
+		<label for='{$propName}_minutes'>
+			Minutes:
```
mdnoble73: Let's translate this and remove the trailing colon

code/web/interface/themes/responsive/DataObjectUtil/duration.tpl
```diff
@@ -0,0 +1,27 @@
+{strip}
+{assign var=hours value= (int) ($propValue / 60)}
+{assign var=minutes value=$propValue % 60}
+<fieldset class="form-inline">
+	<div class="form-group col-sm-6">
+		<label for='{$propName}_hours'>
+			Hours:
```
mdnoble73: Let's translate this and remove the trailing colon

code/web/services/Events/EventGraphs.php
```diff
@@ -0,0 +1,273 @@
...
+
+		$dataSeries = [];
+		$columnLabels = [];
+		$userHours = new EventInstance();
+		$userHours->selectAdd();
+		$userHours->whereAdd("event_instance.deleted = 0");
+		$userHours->whereAdd("event_instance.status = 1"); // Exclude cancelled events
+		if (!empty($query) || !empty($fields)) {
+			$eventField = new EventEventField();
+			foreach ($fields as $key => $value) {
```
mdnoble73, line 150: If we're building this query dynamically we need to make sure there isn't a possibility of sql injection

code/web/services/Events/Calendar.php
```diff
@@ -158,18 +215,38 @@ function launch() {
...
+				break;
+			} else if (!$useWeek && $dayNum > $maxDay) {
 				break;
 			}
 		}
 		$interface->assign('weeks', $weeks);
 
-		$this->display('calendar.tpl', 'Events Calendar ' . $formattedMonthYear, '');
+		$headerImage = $this->getHeaderImage();
+		if (!empty($headerImage)) {
```
mdnoble73: We might want to assign these explicitly to blank if there isn't a header?

mdnoble73 (changes requested):

Abhyastamita:
Thanks Mark!  Should be fixed now.

mdnoble73 (approved):

PR #2265, opened 2025-02-24
DIS-425: Added Existence Checks for Polaris API Response Properties

LeoStoyanovByWater:
- Added check for existence of `$response->AccessToken` and `$response->PatronID`; set either to `null` and log it if either do not exist.
- Underlying problem seems to be that Aspen was trying to make Polaris API calls with a null or empty password during masquerade mode, so undefined properties were being returned in a JSON response from the Polaris authentication API request. Thus, those properties must be set to `null` if they do not truly exist.

### DIFF
--- a/code/web/release_notes/25.03.00.MD
+++ b/code/web/release_notes/25.03.00.MD
@@ -36,7 +36,7 @@
 ### Other Updates
 - Remove unused AspenLiDASettings class. (DIS-393) (*MDN*)
 - Added check to make sure `$catalog` is not null before calling `hasIlsConsentSupport()` in `Library.php`. (DIS-394) (*LS*)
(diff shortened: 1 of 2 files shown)


code/web/release_notes/25.03.00.MD
```diff
@@ -36,6 +36,7 @@
 ### Other Updates
 - Remove unused AspenLiDASettings class. (DIS-393) (*MDN*)
 - Added check to make sure `$catalog` is not null before calling `hasIlsConsentSupport()` in `Library.php`. (DIS-394) (*LS*)
+- Added check for existence of `$response->AccessToken` and `$response->PatronID`; set either to `null` and log it if either do not exist. (DIS-425) (*LS*)
```
mdnoble73: @LeoStoyanovByWater could you move this down under the section 
//leo 
That will make merges easier.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2272, opened 2025-02-28
DIS-287 Debian Installer Improvements

HitScan:
Default to php 8.4 and add a helper script to install or upgrade php as needed. Also added some necessary packages that are no longer installed by default (cron, rsyslog, etc.)
Addresses DIS-287

### DIFF
--- a/code/web/release_notes/25.03.00.MD
+++ b/code/web/release_notes/25.03.00.MD
@@ -42,7 +42,7 @@
 ### Install Updates
 - Properly install the logrotate configuration file on debian systems. (DIS-366) (*MDN*)
 - Increase max_input_vars for php to 5000 on debian systems. (DIS-431) (*MDN*)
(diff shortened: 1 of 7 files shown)


mdnoble73 (changes requested):
@HitScan we had split out the clam av install because some support orgs were wanting to use a different anti-virus scanner. If we want/need to enforce that could you update the Jira with an explanation of why that should be added. Also it looks like logrotate is being setup automatically which is great. Can you document that as well within the Jira and the release notes.

HitScan:
I see, I wasn't sure why there was a separate installer just for the AV stuff. Is there a preference for keeping that separate vs asking a question in the main installer?

mdnoble73:
> I see, I wasn't sure why there was a separate installer just for the AV stuff. Is there a preference for keeping that separate vs asking a question in the main installer?

I'm ok with it being a question in the main installer. We may also want to look at alternative AV scanners, ClamAV can be fairly resource intensive.

mdnoble73 (approved):

PR #2273, opened 2025-02-28
DIS-239: Evergreen patron email update

Chloe070196:
Jira ticket: [DIS-239](https://aspen-discovery.atlassian.net/browse/DIS-239)

[DIS-239]: https://aspen-discovery.atlassian.net/browse/DIS-239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/code/web/Drivers/Evergreen.php
+++ b/code/web/Drivers/Evergreen.php
@@ -2593,6 +2593,9 @@ class Evergreen extends AbstractIlsDriver {
 						if (!empty($mappedPatronData['suffix'])) {
 							$user->_fullname .= ' ' . $mappedPatronData['suffix'];
 						}
+						if (!empty($mappedPatronData['email'])) {
+							$user->email = $mappedPatronData['email'];
+						}
 						$user->_fullname = trim($user->_fullname);
 
 						if (!empty($mappedPatronData['expire_date'])) {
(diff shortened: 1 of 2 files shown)


code/web/Drivers/Evergreen.php
```diff
@@ -778,11 +778,74 @@ function changeHoldPickupLocation(User $patron, $recordId, $itemToUpdateId, $new
...
+
+		/** It seems the response sent back by the Evergreen API will have a
+		 * status code of 200 even upon failure (eg. if the patron password
+		 * is incorrect.) However, a successful update will always result
+		 * in the "payload" property being set to 1. Check for this instead.
+		*/
+		$response['success'] = $responseData['payload'][0] == 1;
+
+		if (!$response['success']) {
+			global $logger;
```
mdnoble73, line 846: @Chloe070196 if the call is successful here we should update the email that we have stored for the patron within Aspen.

code/web/Drivers/Evergreen.php
```diff
@@ -2530,6 +2593,9 @@ public function loadContactInformation(User $user) {
 						if (!empty($mappedPatronData['suffix'])) {
 							$user->_fullname .= ' ' . $mappedPatronData['suffix'];
 						}
+						if (!empty($mappedPatronData['email'])) {
```
mdnoble73: @Chloe070196 this should not be needed if we update the email after the patron updates their email address. loadContactInformation should only need to load transient data (prefixed by _ ) that we don't store in the database.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2277, opened 2025-02-28
Dis 450 clean orphan libcal events

lathomas64:
Creating a static method to clean up orphan events
to test:
* set up springshare libcal settings in aspen and pull events
* delete the springshare settings
* run the eventsIndexer
* the imported events should now be marked as deleted


mdnoble73 (changes requested):
@lathomas64 do we also need to mark the events as deleted within springshare_libcal_events database table? If not, the events will get deleted from Solr at each pass (even though they are already deleted). 

This could also probably be optimized to delete all of the events with one query similar to: 
solrUpdateServer.deleteByQuery("type:event_libcal AND source:" + settingsId).

Then we don't need the orphanEventCleanup property. If we go that route also should add some logging to the event indexing log so adminstrators can see what has been deleted?

mdnoble73:
Oops sorry, didn't see this was still a draft. Sorry for commenting early

lathomas64:
> @lathomas64 do we also need to mark the events as deleted within springshare_libcal_events database table? If not, the events will get deleted from Solr at each pass (even though they are already deleted).
> 
> This could also probably be optimized to delete all of the events with one query similar to: solrUpdateServer.deleteByQuery("type:event_libcal AND source:" + settingsId).
> 
> Then we don't need the orphanEventCleanup property. If we go that route also should add some logging to the event indexing log so adminstrators can see what has been deleted?

Good idea. I wasn't sure how to approach all the solr changes which was the only reason I was doing things piecemeal like that but with the deletebyQuery option we can make this a lot simpler.

mdnoble73 (approved):

PR #2278, opened 2025-02-28
DIS-452 Boundless: Pass email during Checkout-to-placeHolds transition

LiYanjun19:
When a Boundless record is showing “available” in Aspen, but actually “waitlist” on Boundless website, if a patron wants to checkout the record, Aspen will redirect to place holds. But it will fail and shows “invalid email address”.

To replicate the bug:
1. Find a Boundless record that “available” in Aspen but “waitlist” on Boundless. 
2. Checkout the record 
3. A pop-up displaying “Sorry, we could not checkout this Boundless title to you. Would you like to place a hold instead?”
4. Click Ok and place a hold 
5. Get the error “Could not place Boundless hold, Invalid Email Address” 

The error comes from when redirecting from Checkout to Placeholds, the email is not being passed, so, add the parameters to the doHold method.

### DIFF
--- a/code/web/Drivers/Axis360Driver.php
+++ b/code/web/Drivers/Axis360Driver.php
@@ -597,10 +597,6 @@ class Axis360Driver extends AbstractEContentDriver {
 						'text' => 'Place a Hold',
 						'isPublicFacing' => true,
 					]);
-					$email = empty($patron->axis360Email) ? $patron->email : $patron->axis360Email;
-					$result['api']['holdParams'] = [
-						'email' => $email
-					];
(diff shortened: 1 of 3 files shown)


mdnoble73 (changes requested):
It looks like this still won't work if the patron does not have an email set for their account. We should check to see if promptForAxis360Email is on or off and if the patron does not have an Axis360Email or promptForAxis360Email is on, we should prompt them to enter their email as if it was a normal hold.

mdnoble73 (approved):

PR #2283, opened 2025-03-03
DIS-192: iCalendar export, private events, archived event types

Abhyastamita:
- Added Export button for both event search results and for individual event record pages which generates an iCalendar file.  The button appears for events from any vendor, but for repeating Aspen Events, there's an option to just export a single event or all upcoming events in the series.  
  - Tested iCalendar files with Apple Calendar, Google Calendar, and Outlook.com Calendar
- Fixed private events so that they are limited according to the user's permissions for Manage Events.
- Also limited private events that show up in search results and the event calendar, however doesn't fully work as it should.  If the user has the permission View Private Events For All Locations, private events will show up in search and on the calendar.  If the user has more granular permissions just to view private events for home library or home location, then they will not see any private events in the search results or calendar, just in Manage Events.  I had trouble finding a Solr query that would work for that.


Abhyastamita:
- Fixed private event indexing so that search results can be limited to only include private events at the user's home library or home location.
- Added label for private events in search interface.

mdnoble73 (approved):

PR #2290, opened 2025-03-05
DIS-188: Adding series member covers and adding, deleting, and excluding series and series members

Abhyastamita:
- Added expanded edit screen for series members that allows uploading a cover and changing more series member metadata if it's something the user added.
  - No edit or delete button for series members that were added automatically by indexing.  The user can still change some fields like volume for these.
- Added Exclude checkbox for series members so that the user can remove titles in the series that don't belong and they aren't deleted, so the indexer won't re-add them.
  - Added button to Show Excluded Series Titles.  These are hidden by default when editing a series in order to make the display less overwhelming if there are lots of series members.
- Added display of either default or uploaded cover for placeholder series.
- Added option to add a series from the Administer Series page.  User created series do not have the Grouped Work Series Title filled out, so running the indexer will not add or modify these series records at all.  This way no records that don't belong will be accidentally added.
- Changed series author field so that it can be edited but it still has a warning that it might be modified by indexing.


data_dir_setup/.idea/workspace.xml
mdnoble73, line 1: @Abhyastamita can you exclude this form your commit and add to git ignore? It will conflict with other developers so best not to include it.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2295, opened 2025-03-06
DIS-416: Debugging ecommerce applications

lucasmontoya13:
Test plan :

- Login as a staff user with permissions to manage ecommerce payments settings
- Set your prefer ecommerce to manage your payments like Paypal, ProPay, NCR, PaypalPayflow with valid credentials
- By default the option 'Enable Payments Debugging' is active
- Enable your IP to be able to log debugging information
- Go to Greenhouse -> External Log Entry
- Login with a user who has fines to pay (use an incognito window for testing)
- Pay the fines
- The staff user should be able to watch debugging information about movements related with the ecommerce

### DIFF
--- a/code/web/release_notes/25.04.00.MD
+++ b/code/web/release_notes/25.04.00.MD
@@ -27,6 +27,8 @@
 //jason (equinox)
 
 //lucas
+### Other Updates
+- Add an option to allow staff users to get debugging information about payments. (DIS-416) (*LM*)
 
 //tomas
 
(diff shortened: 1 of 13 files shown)


mdnoble73 (changes requested):
I don't think we need the extra settings for each eCommerce setting. Using the debugging by IP should be sufficient as long as no PII is logged.

lucasmontoya13:
I've created another PR with some enhacements. I'm going to close this one!

PR #2299, opened 2025-03-07
DIS-489 Add default value to browseCategoryGroupId

LiYanjun19:
In Sierra and Symphony, in Primary Configuration > Libraries System and Location, it has a button “Update from ILS” and it can import the libraries and locations from ILS directly. 
When clicking this button, it will get an Error `Error inserting Library<br/> SQLSTATE[HY000]: General error: 1364 Field ‘browseCategoryGroupId' doesn't have a default value `

Adding default value 1 to browseCategoryGroupId to prevent this error.


code/web/release_notes/25.04.00.MD
```diff
@@ -23,6 +23,8 @@
 //imani
 
 //yanjun
+### Other Updates
+- Assign default value to browseCategoryGroupId for Sierra and Symphony when importing Libraries and Locations. (DIS-489) (*YL*)
```
mdnoble73: In the changes, only Sierra is handled. Should Symphony have been handled as well? Do we need to ensure that browse category group 1 exists?

LiYanjun19: Symphony only loads locations from ILS but not libraries, so it shouldn't have the same issue as Sierra. I edited the release note and the code for handling browse category group id better.

mdnoble73 (approved):

PR #2301, opened 2025-03-10
DIS-429: Added Cover Manager to Greenhouse

LeoStoyanovByWater:
- Added Cover Manager under Maintenance Tools to allow administrators to force a reload on various image covers by source.

The subsection name is intentionally broad, allowing for potential expansion with additional functionalities in the future.

Test Plan:

1. Navigate to the Greenhouse and select “Cover Manager” under the “Maintenance Tools” section.
2. Select whatever number of sources you would like to force reload and click “Process Selected Sources”.
3. If you query the `bookcover_info` table in the DB, you will notice that the image sources no longer exist.
   - Note 1: If users are actively using the site, any image sources that were deleted could easily be repopulated between the point at which they were deleted and the point of querying the table for them.
   - Note 2: If you reload the Cover Manager page after force reloading covers, those cover sources you reloaded will likely disappear from the list, unless they were immediately repopulated (see Note 1). This fact is explained in the blue, information box at the top of the page.


mdnoble73 (changes requested):
See: https://aspen-discovery.atlassian.net/browse/DIS-429?focusedCommentId=12020

mdnoble73:
Can you resubmit this for 25.05

ghost:
Just to confirm, if I resubmit the PR for 25.05.00, is there any guarantee it will be approved and merged into that release?

mdnoble73:
It's still subject to the normal review process, but yes the plan would be to have it merged into 25.05.

PR #2302, opened 2025-03-11
DIS-493: Added Option to Prioritize Titles from Available Koha Records for Grouped Works Display Titles

LeoStoyanovByWater:
- Added the Prioritize Available Records option for Title Selection option the Indexing Profile.
- If the above option is enabled, titles from available records will be prioritized as the display title for their grouped work.

Note: Intellij fixed tabbing for lines of code inside files I edited.

Test Plan:
1. This is only applicable to Aspen instances with a Koha ILS. Find a grouped work in the catalog that has at least one on-order record, and it will likely have a title in all capital letters. This is how Koha typically titles on-order records.
2. Apply the patch and force reindex the grouped work. After a couple of minutes, and if it has other records, one of their titles will have been chosen, according to the hierarchy listed above, for the grouped work to use as its display title.


mdnoble73 (changes requested):
See https://aspen-discovery.atlassian.net/browse/DIS-493?focusedCommentId=12021

mdnoble73:
Can you resubmit this for 25.05 and we can discuss with catalogers

PR #2309, opened 2025-03-13
DIS-505: Palace Project Settings Cleanup and More Informative Logging

LeoStoyanovByWater:
- Fixed the `runFullUpdate` flag not resetting after successful updates.
- Removed the unused `regroupAllRecords` flag from Palace Project settings to eliminate confusion. Regrouping happens automatically.
- Added more extensive logging so users and developers alike understand what the exporter is doing.
- Disabled editing of the Last Update of Changed Records and Last Update of All Records fields because they are populated by the exporter.


code/web/sys/PalaceProject/PalaceProjectSetting.php
```diff
@@ -62,13 +54,15 @@ public static function getObjectStructure($context = ''): array {
 				'label' => 'Last Update of Changed Records',
 				'description' => 'The timestamp when just changes were loaded',
 				'default' => 0,
+				'readOnly' => true,
```
mdnoble73, line 57: It is at times useful to reset this if problems occurred during an export so we can manually go back a few days or weeks to pickup changes without having to do a full update.

code/web/sys/PalaceProject/PalaceProjectSetting.php
```diff
@@ -62,13 +54,15 @@ public static function getObjectStructure($context = ''): array {
...
 			],
 			'lastUpdateOfAllRecords' => [
 				'property' => 'lastUpdateOfAllRecords',
 				'type' => 'timestamp',
 				'label' => 'Last Update of All Records',
 				'description' => 'The timestamp when just changes were loaded',
 				'default' => 0,
+				'readOnly' => true,
```
mdnoble73, line 65: It is at times useful to reset this if problems occurred during an export so we can manually go back a few days or weeks to pickup changes without having to do a full update.

mdnoble73 (changes requested):

code/web/sys/PalaceProject/PalaceProjectSetting.php
ghost: PalaceProjectExportMain.java never uses any values the user may pass in from the interface regarding these two timestamps. All it does is update both timestamps, so I don't see how updating either from the interface would affect the export process.

mdnoble73: It is used to set stopProcessingDueToLastUpdateTime for collections that have circulation and it's used to determine when to run the index for collections that do not have circulation.

ghost: After analyzing the `PalaceProjectExportMain.java` code again, and after Yanjun and Nick both looked at it, the `lastUpdateOfAllRecords` and `lastUpdateOfChangedRecords` timestamp fields are still never actually read or used to determine processing behavior.
The variable `stopProcessingDueToLastUpdateTime` is determined entirely by comparing the last modified date (`lastTitleModifiedDate`), which is collection metadata from the Palace Project API response, to the collection's last index time (`collection.lastIndexed`) with a buffer, which is a value from the `palace_project_collections` table. And, collection processing decisions are based solely on `collection.hasCirculation` and `collection.lastIndexed < yesterdayInSeconds`. While other exporters may use the two aforementioned timestamps, or at least something similar, the Palace Project exporter does not, either because it was never implemented (the more likely case) or because it was never necessary.

mdnoble73: Oops sorry you are correct, the setting moved down to collection from settings.

mdnoble73 (approved):

PR #2313, opened 2025-03-17
DIS-504: Filtering messaging settings

lucasmontoya13:


code/web/Drivers/Koha.php
```diff
@@ -5921,7 +5933,26 @@ public function getMessagingSettingsTemplate(User $patron): ?string {
 			$messageAttributes[] = $messageType;
 		}
 		$messageAttributesRS->close();
-		$interface->assign('messageAttributes', $messageAttributes);
+		$activeMessagesAttributes = [];
+		
+		foreach($messageAttributes as $messageAttribute){
+
+			$conditions =[
```
mdnoble73: Why do this as an array of checks? Using a set of if then statements would be easier to read. Can you also add information about what the checks are doing so we don't have to reference koha documentation when they need to change?

mdnoble73 (changes requested):

code/web/Drivers/Koha.php
lucasmontoya13: I thought it could be a good option, but i think now it is clearer. I've updated the PR!

mdnoble73 (approved):

PR #2323, opened 2025-03-18
DIS-518: Improved Data Object Property Handling and Error Prevention

LeoStoyanovByWater:
- Added the `handlePropertyChangeEffects()` method to serve as a central mechanism for managing and executing side effects triggered by property changes across Aspen.
  - Addressed issues with deletions not triggering reindexing and the need for regrouping, such as when format-related fields change.
  - More in-depth in the Jira issue.
- Created a `forceRegrouping()` method in the `SystemVariables` class to set the `regroupAllRecordsDuringNightlyIndex` flag when record regrouping is required.
- Added the `dynamic_label` property type to the list of skipped types in `oneToMany` relationships to prevent error messages in logs when saving Indexing Profiles.
- Added a `readOnly` attribute to the `groupingCategory` property in `FormatMapValue` class to further clarify its read-only nature. 
- Implemented an override of the delete method in `FormatMapValue` class to ensure regrouping is properly triggered when format maps are deleted.

Test Plan:
1. Create a dummy Format Map under the “Format Information” section of the Indexing Profile and save.
2. Have a new tab open on the System Variables page and make sure to disable “Run full index tonight” because it was enabled by the creation of the dummy Format Map and save.
3. Delete the dummy Format Map and save, reload the System Variables page, and notice that the “Run full index tonight” option is not enabled.
4. Apply this patch and repeat steps 1-3. Reload the System Variables page and notice that the “Run full index tonight” will be enabled after the deletion of the dummy Format Map.
   - Additionally, assuming “Regroup all records during nightly index” is disabled under System Variables, change the format or format category of a Format Map and save. Reload the System Variables page and notice that the “Regroup all records during nightly index” option will be enabled.
   - For further confirmation, tail follow messages.log to see the logging of the the enabling of the “Run full index tonight” and “Regroup all records during nightly index” options.


code/web/sys/Indexing/FormatMapValue.php
```diff
@@ -229,4 +232,21 @@ public function __get($name) {
...
+	 * Override delete to ensure regrouping is triggered when format map values are deleted.
+	 *
+	 * @param bool $useWhere
+	 * @return int
+	 */
+	public function delete($useWhere = false) : int {
+		// Check if we're deleting a specific record (i.e., not a bulk delete).
+		if (!$useWhere && !empty($this->id)) {
+			// Trigger regrouping if a format value is being deleted.
+			$this->handlePropertyChangeEffects('format', $this->format, null, ['forcesRegroup' => true], 'deleted');
```
mdnoble73: Should this have a reindex as well? The old formats will need to be moved to something new most likely. Why not use the actual object structure rather than one faked in here?

mdnoble73 (changes requested):

code/web/sys/Indexing/FormatMapValue.php
ghost: The force-reindex flag is properly set upon deletions of the format map values because the `formatMap` property under `IndexingProfile.php` has the `'forcesReindex' => true` field. The logging indicates that this is true:
`Forcing nightly index because formatMap on IndexingProfile (1) was deleted from 'object' to 'empty' by user 1 (oneToMany entry).`

I don't know what you mean by "old formats will need to be moved to something new most likely."

You have a good point, so I've changed to use the actual object structure. 

Overall, this overridden deletion method was added because the `FormatMapValue` objects are stored as a `oneToMany` relationship within `IndexingProfile`, but the `forcesRegroup` flag is defined on the `format` and `formatCategory` properties within `FormatMapValue`, not on the `formatMap` property of `IndexingProfile`. I've only added the flag for those properties because, after asking around, those are the two relevant properties for regrouping. This solves the issue where when a `FormatMapValue` is deleted, the standard `DataObject` deletion process has no visibility into the individual property flags of the deleted object.

mdnoble73 (approved):

PR #2329, opened 2025-03-20
DIS-521: Enhanced Filtering to Exclude Volumes with All Suppressed Items from the Holds Dropdown

LeoStoyanovByWater:
- Added functionality to filter out volumes where all associated items have suppressed statuses from the volume holds dropdown.

Minor testing on a server with Koha ILS and on a server with Symphony ILS.

### DIFF
--- a/code/web/services/Record/AJAX.php
+++ b/code/web/services/Record/AJAX.php
@@ -758,6 +758,9 @@ class Record_AJAX extends Action {
 
 			$numItemsWithVolumes = 0;
 			$numItemsWithoutVolumes = 0;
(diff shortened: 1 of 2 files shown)


code/web/services/Record/AJAX.php
mdnoble73, line 1: @LeoStoyanovByWater we already have a list of un-suppressed items within the MarcRecordDriver. Adding this here is going to lead to maintenance issues since we'll need to realize that the suppression rules are being checked in 2 different places. We should use the list of items that has already exists.

mdnoble73:
Closing after discussing with Leo we came up with an alternate approach.

PR #2330, opened 2025-03-20
DIS-520: Implement Item Barcode Fallback for Invalid Call Numbers for Symphony Holds

LeoStoyanovByWater:
- Added fallback to use item barcode when call number is invalid according to the place-hold API response in `placeSirsiHold()` for `SirsiDynixROA.php`.
- Explicit string casting should no longer be needed because of the `'.'` string concatenation.

Minor testing on a Symphony server.

### DIFF
--- a/code/web/Drivers/SirsiDynixROA.php
+++ b/code/web/Drivers/SirsiDynixROA.php
@@ -1430,12 +1430,19 @@ class SirsiDynixROA extends HorizonAPI {
 				$volume = $this->getMissingVolumeKey($webServiceURL, $shortId, $sessionToken, $displayVolume);
 			}
 
+			// Symphony has an odd behavior where a call key (i.e., call number) can be valid when checked directly,
+			// but still fail with "no call number found" when used in hold placement.
+			// Thus, use the call key first, but keep the item barcode for fallback if needed.
+			$originalItemId = null;
+
 			if (!empty($volume)) {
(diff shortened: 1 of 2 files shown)


mdnoble73 (changes requested):
Is there an example of this? The change looks like it will place an item level hold rather than a volume hold which would change hold functionality significantly.

ghost:
@mdnoble73 This would place an item-level hold rather than a volume hold. The logical flow is that it would attempt to place a hold on the volume whose call number is, for whatever reason, invalid (1), and then an item-level hold would be attempted, finally returning success (2). If the same patron attempts to place a hold on the same volume again, it would go through the same sequence and properly return a message indicating that the patron already has a hold on the item (3).

(1) {"messageList":[{"code":"placeHold.noCallnumber","message":"Cannot place a hold on this title (no call number found). Please contact a librarian."}]}

(2) {"holdFee":null,"holdRecord":{"resource":"/circulation/holdRecord","key":"70501507"}}

(3) {"messageList":[{"code":"hatErrorResponse.722","message":"User already has a hold on this material"}]}

mdnoble73:
Let’s talk about this at the symphony group or in the symphony channel to be sure this is the behavior that all Symphony libraries will want in this situation.

mdnoble73:
Can you resubmit this for 25.05 and we can discuss with the Symphony group

PR #2331, opened 2025-03-21
DIS-219: Filter Holds

AlexanderBlanchardAC:


AlexanderBlanchardAC:
Tested - the if check still runs with the current casing as it is a function

mdnoble73 (approved):

PR #2333, opened 2025-03-21
DIS-533: Fixed NYT Lists Visibility in Consortia

LeoStoyanovByWater:
- Fixed NYT lists visibility in consortia when using 'Lists from library list publishers Only' option for a library system.


code/web/sys/DBMaintenance/version_updates/25.04.00.php
```diff
@@ -31,6 +31,14 @@ function getUpdates25_04_00(): array {
 		//Yanjun Li - ByWater
 
 		// Leo Stoyanov - BWS
+		'fix_nyt_user_home_location' => [
+			'title' => 'Fix NYT User Home Location',
+			'description' => 'Set nyt_user home location to -1 to ensure NYT lists are visible in consortia when "Lists from library list publishers Only" is selected.',
+			'continueOnError' => true,
+			'sql' => [
+				"UPDATE user SET homeLocationId = -1 WHERE username = 'nyt_user'",
```
mdnoble73: When should make sure the source of the user is admin as well just in case a real user had a username of nyt_user. Which is unlikely, but possible.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2335, opened 2025-03-24
DIS-491: Fixed CloudLibrary Renewals Sequence

LeoStoyanovByWater:
- Fixed renewals of CloudLibrary titles and updated the UI during the renewal process to be more responsive and user-friendly.

Test Plan:
1. Assuming you already checked out a CloudLibrary title and you are within the three-day window of its expiry date, navigate to your Checked Out Titles.
2. Click to renew the CloudLibrary title, wait a couple of seconds, and notice that the title has been renewed.


mdnoble73 (changes requested):
See https://aspen-discovery.atlassian.net/browse/DIS-491?focusedCommentId=12023

mdnoble73 (approved):

PR #2340, opened 2025-03-27
DIS-76: HeyCentric integration

Chloe070196:
Jira ticket at: https://aspen-discovery.atlassian.net/browse/DIS-76

### DIFF
--- a/code/web/sys/Account/UserPayment.php
+++ b/code/web/sys/Account/UserPayment.php
@@ -24,6 +24,7 @@ class UserPayment extends DataObject {
 	public $deluxeRemittanceId;
 	public $deluxeSecurityId;
 	public $ncrTransactionId;
+	public $heyCentricPaymentReferenceNumber;
 	public $requestingUrl;
 
 	public static function getObjectStructure($context = '') {
@@ -117,6 +118,13 @@ class UserPayment extends DataObject {
 				'type' => 'checkbox',
 				'label' => 'Cancelled?',
 				'description' => 'Whether or not the user cancelled the payment',
+				'readOnly' => true,
+			],
+			'declined' => [
+				'property' => 'declined',
+				'type' => 'checkbox',
+				'label' => 'Cancelled?',
+				'description' => 'Whether or not the payment was declined',
 				'readOnly' => true,
 			],
 			'error' => [
--- a/code/web/sys/Account/UserPayment.php
+++ b/code/web/sys/Account/UserPayment.php
@@ -123,7 +123,7 @@ class UserPayment extends DataObject {
 			'declined' => [
 				'property' => 'declined',
 				'type' => 'checkbox',
-				'label' => 'Cancelled?',
+				'label' => 'Declined?',
 				'description' => 'Whether or not the payment was declined',
 				'readOnly' => true,
 			],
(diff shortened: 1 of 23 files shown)


code/web/sys/Account/UserPayment.php
```diff
@@ -119,6 +120,13 @@ public static function getObjectStructure($context = '') {
 				'description' => 'Whether or not the user cancelled the payment',
 				'readOnly' => true,
 			],
+			'declined' => [
+				'property' => 'declined',
+				'type' => 'checkbox',
+				'label' => 'Cancelled?',
```
mdnoble73: The label on this looks incorrect.

code/web/sys/SystemVariables.php
```diff
@@ -395,6 +395,24 @@ public static function getSystemVariables() {
 		return SystemVariables::$_systemVariables;
 	}
 
+	public function getCurrencySymbol() {
```
mdnoble73, line 400: We shouldn't use a hard coded list here. We can use NumberFormatter to get the correct code see: https://stackoverflow.com/questions/13897516/get-currency-symbol-in-php. Also, we're already in the active SystemVariable so there is no need to look up the SystemVariable again to get the currency code.

mdnoble73 (changes requested):
See comments inline as well as https://aspen-discovery.atlassian.net/browse/DIS-76?focusedCommentId=12049

code/web/sys/Account/UserPayment.php
Chloe070196: It does - I will check what it is meant to be, and update accordingly

code/web/sys/SystemVariables.php
Chloe070196: Thank you for this; I will update the method shortly

Chloe070196: Looking into this, I think that to achieve it, the currency codes currently in use in Aspen would need to change. What I found was that the currency formatter expects 'en_GB' instead of 'GB', 'en_US' instead of 'USD', etc... (happy to be proven wrong if this is incorrect, or if there is a workaround!)

I could programmatically add 'en_' to the currency codes currently in Aspen, but this would limit internationalisation (just as the current hardcoded list does), as we would not have a way to handle codes that start with other suffixes. It also wouldn't resolve the US / USD incompatibility.

The other option would be to use the user locale, but I'm wondering if that may clash with the Currency Code system setting from a user standpoint (as donations would follow the system's currency symbol, but fines paid through HeyCentric would take the user locale into account, and the two might differ at times). Do you think this could possibly prove an issue?

Lastly, since `getCurrencySymbol()` already exists in `Donation.php` (its addition to `SystemVariables.php` aimed to make the existing method more widely available so that the HeyCentric integration could make use of it), do we want to consider this change out of scope? If so, I will raise a separate Jira ticket for this issue.

mdnoble73:
Can you resubmit this for 25.05

Chloe070196:
As requested, the HeyCentric integration has been re-submitted for `25.05.00`: I have rebased the changes, swapped the Aspen Discovery branch this PR targets, and done some preliminary testing.

While this is marked as draft, the code itself is ready for review.
That said, we will be conducting some additional internal post-rebase testing before marking this as ready to be merged into 25.05.00.

mdnoble73 (approved):

PR #2348, opened 2025-03-29
DIS-540 Sierra Contact Info State and ZIP handling

LiYanjun19:
For libraries with Sierra, patrons' contact information, city, state, and zip are always displayed together. The current code splits city, state, and zip by checking commas. However, some libraries’ API responses for patron information don’t have the comma between city, state, and zip, but split them with only spaces, or missing ZIP. 

With this fix, different cases of address line 2 can be handled well. 
- City, State ZIP
- City, State
- City State ZIP

### DIFF
--- a/code/web/Drivers/Sierra.php
+++ b/code/web/Drivers/Sierra.php
@@ -2128,10 +2128,21 @@ class Sierra extends Millennium {
 				if (strpos($line2, ',')) {
 					$user->_city = substr($line2, 0, strrpos($line2, ','));
 					$stateZip = trim(substr($line2, strrpos($line2, ',') + 1));
-					$user->_state = substr($stateZip, 0, strrpos($stateZip, ' '));
-					$user->_zip = substr($stateZip, strrpos($stateZip, ' '));
+					if (strpos($stateZip, ' ')) {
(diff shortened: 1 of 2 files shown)


code/web/Drivers/Sierra.php
```diff
@@ -2128,10 +2128,21 @@ private function loadContactInformationFromApiResult(User $user, stdClass $patro
...
-					$user->_zip = substr($stateZip, strrpos($stateZip, ' '));
+					if (strpos($stateZip, ' ')) {
+						$user->_state = substr($stateZip, 0, strrpos($stateZip, ' '));
+						$user->_zip = substr($stateZip, strrpos($stateZip, ' '));
+					} else {
+						$user->_state = trim($stateZip); 
+					}
 				} else {
-					$user->_city = $line2;
+					$parts = preg_split('/\s+/', $line2);
```
mdnoble73, line 2138: It doesn't look like this would handle cities that have more than one word or possibly states with more than one word. 

Perhaps we need to define the format the library uses and parse based on that?

mdnoble73 (changes requested):

mdnoble73:
Can you resubmit this for 25.05 after you look into the City/State split with multi-word cities/states

PR #2356, opened 2025-04-01
DIS-575: Add latest tag to github actions for docker image push 

Jacobomara901:
This bug is causing indexing to fail on all ADB instances. Since there is no 'latest' tag, the development environments (which rely on moving images and not static ones) have been getting outdated for a while. This means that the SOLR schema has now become out of date, causing the indexing to fail for any users of an image without a specified tag. Docker is designed to automatically pull 'latest' if no tags are given but this image is not updated when a static tag is pushed (as we do currently). We need to do this separately in the github actions and push both to the static tag and latest.

Since this is such a small patch that doesn't touch or affect the codebase at large, I have not updated the release notes here. Let me know if you would prefer this to be included in there and I can get a quick commit pushed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,6 +51,9 @@ jobs:
           file: docker/Dockerfile
           push: ${{ startsWith(github.ref, 'refs/tags/') }}
           tags: |
+            aspendiscovery/aspen:latest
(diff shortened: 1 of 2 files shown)


mdnoble73:
Adding release notes would be great, thanks.

PR #2357, opened 2025-04-01
DIS-573: Added Patron Type Assignment Upon Login & Added SSO Information

LeoStoyanovByWater:
- Upon login, and with `ssoAuthOnly` enabled, assign the user's patron type to the "Patron type given to staff users," which takes logical precedence, or to "A fallback value for category ID."
- Added alerts to the SSO Setting page to inform administrators that that patron types must be correctly assigned to the `admin_sso` profile when `ssoAuthOnly` is enabled, with specific warnings for LDAP, SAML, and OAuth services to ensure proper staff permissions and access rights.

Test Plan:
1. Navigate to your SSO Settings, set one up if one does not exist, and select the “SAML 2” service.
2. Enable “Only authenticate users with single sign-on” and you will notice a yellow alert below it.
3. Select a patron type that is assigned to the admin_sso account profile into either the “Patron type given to staff users” field or the “A fallback value for category ID” field, where the former has logical precedence.
   - If you do not have a patron type that is assigned to the admin_sso account profile, create one under Patron Types of the admin interface.
5. Use SSO to log in with a user account that is assigned to a patron type that is associated with the ILS account profile.
6. After a successful login, query the user table to find the user account for which you used to log in. Notice that its patron type has changed to one of the following you set in either the “Patron type given to staff users” field or the “A fallback value for category ID” field.


code/web/interface/themes/responsive/css/alerts.css
mdnoble73, line 1: What are these changes for? They aren't documented in the pull request or the ticket and we shouldn't change style for some alerts. These also don't respect the color customization that admins can make within theme settings. This should be another Jira ticket/pull request if desired.

mdnoble73 (changes requested):
The check to see if the admin_sso account profile exists does not seem to be running for new SSOSettings. 

Also see below

code/web/interface/themes/responsive/js/aspen/admin.js
```diff
@@ -2322,7 +2322,192 @@ AspenDiscovery.Admin = (function () {
...
+			// Change listener to service selector to rerun validation.
+			$serviceSelect.on('change.ssoService', function() {
+				setTimeout(() => AspenDiscovery.Admin.checkSSOAuthOnlyPatronTypes(), 100);
+			});
+
+			if (!isSSOEnabled || !ssoSettingId) return true;
+
+			// Show loading message.
+			$formGroup.after(`
+				<div id="ssoAuthOnly_warning" class="alert alert-bs53-warning mt-2">
```
mdnoble73: Can we use a regular alert-warning here for consistencey

code/web/services/Admin/AJAX.php
```diff
@@ -1683,5 +1683,83 @@ public function exportUsageData() {
...
+				$library = new Library();
+				$library->libraryId = $libraryId;
+				if ($library->find(true) && !empty($library->accountProfileId)) {
+					$logger->log("$library->accountProfileId", Logger::LOG_ERROR);
+					$primaryProfileIds[$library->accountProfileId] = $library->accountProfileId; // Use keys for uniqueness.
+				}
+			}
+		} else {
+			// If no libraries are explicitly linked, default to the global library's profile.
+			$logger->log("2", Logger::LOG_ERROR);
```
mdnoble73: It looks like there are several leftover logging statements that can be removed

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2358, opened 2025-04-02
DIS-576: Fix Validation Error When Using First Available Item in Volume Hold Form

LeoStoyanovByWater:
- Modified the volume hold form to bypass volume selection validation when the "First Available Item" radio button is selected.
- Enhanced the UI to use Bootstrap alerts rather than JS alerts.

Test Plan:
1. Place a hold on a volume that calls `AspenDiscovery.Record.showPlaceHoldVolumes()`.
2. That calls `getPlaceHoldVolumesForm()` and pops up a form modal.
3. You have a choice between placing a hold on the "First Available Item" or by selecting a "Specific Volume."
4. Within `getPlaceHoldVolumesForm()`, `alwaysPlaceVolumeHoldWhenVolumesArePresent()` returns false (e.g., Koha driver) and you choose to place a hold on the "First Available Item."
5. There is a call to `AspenDiscovery.Record.placeVolumeHold()`, but no volume was selected as intended, so you see a JS alert saying "You must select a volume before continuing."


code/web/interface/themes/responsive/js/aspen/record.js
```diff
@@ -446,7 +465,7 @@ AspenDiscovery.Record = (function () {
 			if (holdType.length > 0) {
 				params['holdType'] = holdType.val();
 			} else {
-				if ($('#holdTypeBib').is(':checked')) {
+				if (holdTypeBib.is(':checked')) {
```
mdnoble73: @LeoStoyanovByWater this should be $holdTypeBib or the variable should be renamed to just holdTypeBib

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2379, opened 2025-04-11
Dis 352 community engagement two

AlexanderBlanchardAC:
https://aspen-discovery.atlassian.net/jira/software/c/projects/DIS/boards/2?selectedIssue=DIS-352


install/updateCron_25.04.php
mdnoble73, line 1: This needs to get renamed to updateCron_25.05.php to run for the correct version

code/web/interface/themes/responsive/layout.tpl
```diff
@@ -38,6 +38,15 @@
 		{if !empty($og_url)}
 			<meta property="og:url" content="{$og_url|escape:html}"/>
 		{/if}
+		{if !empty($twitter_card)}
```
mdnoble73, line 41: These look like they are part of another Jira ticket, are these intended to be here?

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2385, opened 2025-04-15
DIS-624: updating release notes for LiDA changes.

lathomas64:
release notes for LiDA update: https://github.com/Aspen-Discovery/aspen-lida/pull/9


mdnoble73:
These should be added directly to the LiDA repository now

PR #2391, opened 2025-04-18
DIS-672: updated release notes for Lida Updates.

lathomas64:
Test plan at: https://github.com/Aspen-Discovery/aspen-lida/pull/10


mdnoble73:
These should now be committed directly to the Aspen LiDA repository

PR #2394, opened 2025-04-21
DIS-678: Changed Indexing Profile Name Field to a Dropdown of Account Profile Names

LeoStoyanovByWater:
- When adding a new Indexing Profile, the "Name" field has been changed to a dropdown of Account Profile options.
- Users may no longer create an Indexing Profile before at least one, valid ILS Account Profile exists.

Test Plan:
1. Apply the patch and navigate to Indexing Profiles.
2. Create a new Indexing Profile and select a valid IL account profile from the “Name” field dropdown.
   - If you attempt to create another Indexing Profile and one already exists, the creation will fail because that has not been implemented yet.


code/web/sys/Account/AccountProfile.php
```diff
@@ -568,6 +568,23 @@ function getIndexingProfile() : ?IndexingProfile {
 		return $this->_indexingProfile;
 	}
 
+	/**
+	 * Checks if there are any Account Profiles other than 'admin' or 'admin_sso'.
+	 *
+	 * @return bool True if at least one ILS profile exists, false otherwise.
+	 */
+	public static function hasValidILSProfiles(): bool {
```
mdnoble73, line 576: This should be hasValidAccountProfiles right?

mdnoble73 (approved):

PR #2406, opened 2025-04-24
DIS-700: Hoopla Flex

LiYanjun19:
This PR adds support for Hoopla Flex titles to Aspen, allowing Aspen to handle instant titles and flex titles independently, allowing patrons to interact with Flex titles in the catalog. 

Hoopla extraction, settings, and indexing updates: 
- Hoopla Flex titles extract at 1 am local time with Instant titles 
- Hoopla Flex titles availability updates every 5 min 
- Handle Instant and Flex title's daily incremental extractions, and full updates separately 
- Add include Instant and include Flex options to Hoopla scopes to allow for filtering of Hoopla Flex and Hoopla Instant titles
- Update dashboard for holds 
- Update export log for better handling of instant and flex titles 
- Remove indexByDay

Hoopla patron circulation updates:
- Allow patrons to checkout, return, place holds, cancel holds for Hoopla Flex titles 
- Update registration prompts


mdnoble73 (changes requested):
I started testing this and hit the following error right away. 

1. Updated to the new code
2. Ran database updates
3. Did not run hoopla updates (which will be typical after deploy until the nightly hoopla index runs)
4. Logged in with a user that is active in Hoopla
5. Tried to checkout a hoopla title. 
6. With debug on, got an error that hooplaType was not defined. Looks like in Hoopla_AJAX you need a check for if hooplaType is defined and if not default it to instant. With debug off, the checkout prompts never displayed.

code/web/interface/themes/responsive/MyAccount/account-sidebar.tpl
```diff
@@ -111,6 +111,14 @@
 											</a>
 										</li>
 									{/if}
+									{if $user->isValidForEContentSource('hoopla')}
```
mdnoble73: The hold functionality should be limited to only patrons/libraries that have access to flex

code/web/Drivers/HooplaDriver.php
```diff
@@ -192,21 +205,46 @@ public function getAccountSummary(User $user): AccountSummary {
...
+					$summary->numCheckoutsRemaining = $hooplaPatronStatusResponse['body']->borrowsRemaining;
 				} else {
 					global $logger;
-					$hooplaErrorMessage = empty($hooplaPatronStatusResponse->message) ? '' : ' Hoopla Message :' . $hooplaPatronStatusResponse->message;
+					$hooplaErrorMessage = empty($hooplaPatronStatusResponse['body']->message) ? '' : ' Hoopla Message :' . $hooplaPatronStatusResponse['body']->message;
 					$logger->log('Error retrieving patron status from Hoopla. User ID : ' . $user->id . $hooplaErrorMessage, Logger::LOG_NOTICE);
 					$this->hooplaPatronStatuses[$user->id] = false; // Don't do status call again for this user
 				}
+				// Get Holds status 
+				$holdsUrl = $patronURL . '/holds/current';
```
mdnoble73: Do we need to check this only if the patron/library has access to flex? Otherwise it's an extra call that we know will fail.

code/web/migrations/importPikaData.php
```diff
@@ -263,7 +263,8 @@ function loadUserInfoFromCSV(?array $userRow, User $existingUser = null): User {
 	$userFromCSV->initialReadingHistoryLoaded = $userRow[$curCol++];
 	$userFromCSV->noPromptForUserReviews = $userRow[$curCol++];
 	$userFromCSV->source = cleancsv($userRow[$curCol++]);
-	$userFromCSV->hooplaCheckOutConfirmation = $userRow[$curCol];
+	$userFromCSV->hooplaCheckOutConfirmation = $userRow[$curCol++];
+	$userFromCSV->hooplaHoldQueueSizeConfirmation = $userRow[$curCol++];
```
mdnoble73: Do we know that these will be exported from Pika?

mdnoble73 (changes requested):

code/web/interface/themes/responsive/MyAccount/account-sidebar.tpl
LiYanjun19: changes made

code/web/Drivers/HooplaDriver.php
LiYanjun19: changes made

code/web/migrations/importPikaData.php
LiYanjun19: Reverted this part

LiYanjun19:
Also limit the Hoopla Options 'Display hold queue size before placing holds' to only patrons can access Flex

mdnoble73 (approved):

mdnoble73 (changes requested):
@LiYanjun19 can you resolve the merge conflicts and then we can merge?

mdnoble73 (approved):

PR #2441, opened 2025-05-01
DIS-313: Fix: amend wording and check selected user exists prior to filtering for exports

AlexanderBlanchardAC:
Amend wording of label 
Ensure selected user exists when filtering exports - if on "all" page filter by record only.

### DIFF
--- a/code/web/services/MyAccount/AJAX.php
+++ b/code/web/services/MyAccount/AJAX.php
@@ -3819,7 +3819,7 @@ class MyAccount_AJAX extends JSON_Action {
 		foreach ($allCheckedOut as $key => $checkout) {
 			$checkout->recordId = $this->normalizeRecordId($checkout->recordId);
 
-			if (!empty ($slectedUser) && $checkout->userId != $selectedUser) {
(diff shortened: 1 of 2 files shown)


code/web/services/MyAccount/AJAX.php
```diff
@@ -3819,7 +3819,7 @@ public function filterCheckoutsBySelected(array $allCheckedOut, $selectedCheckou
 		foreach ($allCheckedOut as $key => $checkout) {
 			$checkout->recordId = $this->normalizeRecordId($checkout->recordId);
 
-			if ($checkout->userId != $selectedUser) {
+			if (!empty ($slectedUser) && $checkout->userId != $selectedUser) {
```
mdnoble73: @AlexanderBlanchardAC there is a typo in this. It should be $selectedUser

mdnoble73 (changes requested):

code/web/services/MyAccount/AJAX.php
AlexanderBlanchardAC: Sorry, corrected.

mdnoble73 (approved):

PR #2445, opened 2025-05-02
DIS-757: Initialize Placard Image Max Height Variable

LeoStoyanovByWater:
- Fixed an issue when attempting to create a new theme and an initialization error would appear.

### DIFF
--- a/code/web/sys/Theming/Theme.php
+++ b/code/web/sys/Theming/Theme.php
@@ -548,7 +548,7 @@ class Theme extends DataObject {
 
 	/** @noinspection PhpUnused */
 	public static int $defaultPlacardImageMaxHeight = 0;
-	public int $placardImageMaxHeight = 0;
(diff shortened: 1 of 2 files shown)


code/web/sys/Theming/Theme.php
```diff
@@ -548,7 +548,7 @@ class Theme extends DataObject {
 
 	/** @noinspection PhpUnused */
 	public static int $defaultPlacardImageMaxHeight = 0;
-	public int $placardImageMaxHeight;
+	public int $placardImageMaxHeight = 0;
```
mdnoble73: This can't be initialized in this way or the data object will search for it since the value is set.

ghost: Remove the variable type to follow DataObject conventions.

mdnoble73 (approved):

PR #2447, opened 2025-05-05
DIS-599: Apply restrictions

lucasmontoya13:


mdnoble73 (changes requested):
@lucasmontoya13 Why does this need to be stored in the account summary? It doesn't look like it's being displayed with the account summary. If it doesn't need to be stored there, the changes could be limited to Koha.php

lucasmontoya13:
I've made the requested changes! I'm not sure if github notifies you about it. Thank you

mdnoble73 (approved):

PR #2448, opened 2025-05-05
DIS-760: Fix script upgrade_debian_25.02.00.sh to correctly update se…

kylemhall:
…ssion.gc_probability in php.ini

This script simply does nothing when run.

Test Plan:
1) Set up aspen on a debian system
2) Check out default branch of aspen
3) Ensure the correct php.ini file has:
   session.gc_probability = 0
4) Run install/upgrade_debian_25.02.00.sh
5) Verify php.ini still has session.gc_probability = 0 6) Check out this branch
7) Run install/upgrade_debian_25.02.00.sh
8) Verify php.ini still has session.gc_probability = 1


mdnoble73 (changes requested):
We are now on php 8.3 so the php_version should be changed to 8.3. 
The fix could just use the && operator on line 8 rather than ||
If this is intended to run with 25.06 a new upgrade_debian_25.06.00.sh file should be created. 
The proposed fix would insert a second session.gc_probability line if the existing value is not 0. 
This fix needs release notes.

kylemhall:
> We are now on php 8.3 so the php_version should be changed to 8.3.

Do you mean 8.4?

> The fix could just use the && operator on line 8 rather than || 

I think the additional code adds a lot of clarity to what the script is doing.

> If this is intended to run with 25.06 a new upgrade_debian_25.06.00.sh file should be created.

Can do! Will you be backporting this bug fix to 25.02, 25.03, 25.04, and/or 25.05?

> The proposed fix would insert a second session.gc_probability line if the existing value is not 0. 

You are incorrect.

> This fix needs release notes.

Can do!

mdnoble73 (approved):

PR #2450, opened 2025-05-05
DIS-761: Add code from Koha to Unknown Error message for more helpful…

lathomas64:
… debugging.

Get Koha and aspen into any state where getRenewErrorMessage would normally return unknown Error and if you are passing in a value for either the 1st or 3rd parameter you should see that included in the response. It is difficult to make this occur reliable precisely because Unknown Error should not be occuring most of the time but we are simple updating the information printed when it does so screenshots can be more helpful in diagnosing what is going on.


code/web/Drivers/Koha.php
```diff
@@ -7362,6 +7362,10 @@ protected function getRenewErrorMessage(?SimpleXMLElement $error, string $messag
 			$message .= 'Item is an onsite checkout';
 		} elseif ($code == "has_fine") {
 			$message .= 'Item has an outstanding fine';
+		} elseif (!empty($code))
+		{
+			$message = 'Unknown error:' . $code;
+		}
```
mdnoble73: @lathomas64 it looks like there is an extra } in the proposed changes that make the file not valid. Can you remove that.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2451, opened 2025-05-05
DIS-759: Validate Before Removing Themes and Libraries from Themes

LeoStoyanovByWater:
- The essential default theme can no longer be deleted, as there must be at least one theme for site stability.
- If deleting a theme would leave a library with no theme, deletion is prevented.
- If deleting a parent theme on which other child themes depend, deletion is prevented.
- If a theme cannot be deleted because it's actively in use, there is a clear alert after saving that explains exactly why it cannot be removed.
- The "Delete All" button has been removed from the Themes administration page to prevent the deletion of all themes.

Test Plan:
1. Navigate to the Themes admin interface and delete a Theme that is the parent theme of another theme (i.e., another themes extends the about-to-be-deleted theme). I recommend creating backup tables of the following tables so they can be re-inserted after testing: `themes`, `library_themes`, and `location_themes`. 
2. If you navigate to the catalog that the orphaned theme is responsible for displaying, the catalog will pop up AJAX request errors and template errors and be generally unusable.
3. If you created backups, re-insert the data from the backup tables into the original tables. If you did not create backups, run SQL statements to remove the deleted theme’s ID from the `library_themes` and `location_themes` tables because these dependencies will cause conflicts.
4. Apply the patch and redo steps 1-3. Notice that an error will display if you attempt to delete themes with child and library dependencies.
   - In terms of library dependencies, if a library from Library Systems has no associated themes, its catalog will contain template errors. After the patch, you cannot delete themes from a library, or delete libraries from a theme, if that library has no other associated themes; it must have at least one theme.

### DIFF
--- a/code/web/sys/LibraryLocation/Library.php
+++ b/code/web/sys/LibraryLocation/Library.php
@@ -4972,8 +4972,52 @@ class Library extends DataObject {
 		return $this->_themes;
 	}
 
+	/**
+	 * Checks if this library has multiple themes assigned to it.
+	 *
+	 * @return bool True if the library has more than one theme, false otherwise.
+	 */
+	public function hasMultipleThemes(): bool {
+		$themes = $this->getThemes();
+		return count($themes) > 1;
+	}
+
 	public function saveThemes() : void {
 		if (isset ($this->_themes) && is_array($this->_themes)) {
+			// First, check if all themes are being deleted.
+			$themesToDelete = 0;
+			$totalThemes = count($this->getThemes());
+
+			foreach ($this->_themes as $obj) {
+				/** @var LibraryTheme $obj */
+				if ($obj->_deleteOnSave) {
+					$themesToDelete++;
+				}
+			}
+
+			// If all themes would be deleted, prevent it.
+			if ($themesToDelete > 0 && $themesToDelete >= $totalThemes) {
+				$preventionMessage = translate([
+					'text' => 'Cannot delete all themes from a library. Each library must have at least one theme assigned to it.',
+					'isAdminFacing' => true
+				]);
+
+				$user = UserAccount::getActiveUserObj();
+				if ($user) {
+					$user->updateMessage = $preventionMessage;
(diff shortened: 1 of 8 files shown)


code/web/sys/LibraryLocation/Library.php
```diff
@@ -4972,8 +4972,52 @@ public function getThemes(): array {
...
+					$user->updateMessage = $preventionMessage;
+					$user->updateMessageIsError = true;
+					$user->update();
+				}
+
+				// Prevent the deletion by unsetting the deleteOnSave flag.
+				foreach ($this->_themes as $obj) {
+					if ($obj->_deleteOnSave) {
+						$obj->_deleteOnSave = false;
+						break; // Only need to preserve at least one theme.
```
mdnoble73, line 5016: If someone tries to delete all, they probably made a mistake so preserving all rather than preventing one at random seems like it would be best?

code/web/sys/Theming/Theme.php
```diff
@@ -3218,10 +3218,63 @@ public function __set($name, $value) {
 		}
 	}
 
-	public function saveLibraries() {
+	public function saveLibraries(): void {
```
mdnoble73, line 3221: This can go into the function getDeletionBlockInformation which would prevent deletion if deleting the theme would leave any library without a theme.

code/web/sys/LibraryLocation/Library.php
mdnoble73: Can this also go in getDeletionBlockInformation?

mdnoble73 (changes requested):

code/web/sys/Theming/Theme.php
ghost: From my understanding of `getDeletionBlockInformation()`, it is invoked when the entire theme object is being deleted. In this case, we want to prevent users from deleting the selected libraries within the theme object if they would be left without any themes.

PR #2462, opened 2025-05-09
DIS-786: Show List Titles & Descriptions When Printing Lists

LeoStoyanovByWater:
- Enhanced list printing to ensure the list's title is clearly displayed on the printout and all item descriptions are no longer hidden.
- Fixed an issue where printing lists in landscape mode would only print the first page; all items will now be included across multiple pages as needed.

Tested on Chrome, Mozilla, and Edge.

Test Plan:
1. Navigate to any list in the catalog and click the “Print List” button. Notice that the list title is not displayed and that list descriptions are hidden.
2. Switch to “Landscape” layout mode in the print preview. Notice that there is only one sheet and list items are cut off.
3. Apply the patch and repeat steps 1-2.

### DIFF
--- a/code/web/interface/themes/responsive/css/print.less
+++ b/code/web/interface/themes/responsive/css/print.less
@@ -1,10 +1,10 @@
 @media print {
-  /* Make sure descriptions are visible when printing */
+  /* Make sure descriptions are visible. */
   .hidden-xs {
     display: block !important;
   }
 
-  /* Make sure list title is prominently displayed */
+  /* Make sure list title is prominently displayed. */
   #listTitle {
     font-size: 24px;
     margin-top: 20px;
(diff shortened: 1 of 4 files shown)


code/web/interface/themes/responsive/css/print.less
```diff
@@ -0,0 +1,52 @@
+@media print {
+  /* Initially, make sure descriptions are visible. */
+  .hidden-xs {
+    display: block !important;
```
mdnoble73, line 6: We'll want this tied to a specific class. We definitely don't want all hidden information to be printed.

code/web/interface/themes/responsive/css/print.less
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  /* Hide buttons and controls that aren't needed for print. */
+  .btn-toolbar, .btn-group, .resultActions {
+    display: none !important;
+  }
+
+  /* Hide mobile-only toggle rows (e.g. "Description" and "Click to view") in print. */
+  .visible-xs {
```
mdnoble73: Why do we want all visible things to be hidden? can this be done by more specific id or class?

mdnoble73 (changes requested):
It looks like several things are overly general for printing. Since print.less will be applied to all printed pages, this should be restricted with more specificity in terms of what is and isn't printed.

code/web/interface/themes/responsive/css/print.less
ghost: All CSS rules inside `print.less` have been scoped to `body#MyAccount-MyList`.
`.hidden-xs` has been scoped to new `.list-entry-hidden-desc` and `.list-entry-hidden-format`.

code/web/interface/themes/responsive/css/print.less
ghost: `.visible-xs` replaced with `.list-entry-desc-toggle` class for scoping.

mdnoble73 (approved):

PR #2466, opened 2025-05-09
DIS-560: Enhanced Side Load Permissions

K-Alette:
Add new settings for sharing and owning libraries for side load collections 
Add new permissions for side load administration for home library admins 
Update display of side loads and side load scopes for those with only home library permissions

Update release notes


mdnoble73 (changes requested):
Found a few things testing, see DM

PR #2467, opened 2025-05-09
DIS-785 Remove Inactive Boundless Records

LiYanjun19:
Inactive Boundless records couldn't be removed from the database because records' related availability didn't get deleted successfully from the database. This patch corrected the query.

### DIFF
--- a/code/axis_360_export/src/com/turning_leaf_technogies/axis360/Axis360Extractor.java
+++ b/code/axis_360_export/src/com/turning_leaf_technogies/axis360/Axis360Extractor.java
@@ -475,8 +475,9 @@ public class Axis360Extractor {
 			for (Axis360Title axis360Title : existingRecords.values()) {
 				if (!axis360Title.isDeleted() && !axis360Title.isProcessed()) {
 					//Remove Axis360 availability
-					deleteAxis360AvailabilityStmt.setString(1, axis360Title.getAxis360Id());
+					deleteAxis360AvailabilityStmt.setLong(1, axis360Title.getId());
 					deleteAxis360AvailabilityStmt.setLong(2, setting.getId());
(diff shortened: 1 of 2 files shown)


code/axis_360_export/src/com/turning_leaf_technogies/axis360/Axis360Extractor.java
```diff
@@ -475,8 +475,9 @@ private int deleteItems(Axis360Setting setting, HashMap<String, Axis360Title> ex
 			for (Axis360Title axis360Title : existingRecords.values()) {
 				if (!axis360Title.isDeleted() && !axis360Title.isProcessed()) {
 					//Remove Axis360 availability
-					deleteAxis360AvailabilityStmt.setString(1, axis360Title.getAxis360Id());
+					deleteAxis360AvailabilityStmt.setLong(1, axis360Title.getId());
```
mdnoble73, line 478: @LiYanjun19 this doesn't look right to me since the query is looking for the Axis360 id rather than the Aspen ID. I think this was just missing the executeUpdate a few lines down.

mdnoble73 (changes requested):

code/axis_360_export/src/com/turning_leaf_technogies/axis360/Axis360Extractor.java
LiYanjun19: @mdnoble73 The query is "DELETE FROM axis360_title_availability where titleId = ? and settingId = ?,  the `titleId` here is the foreign key referencing axis360_title.id instead of the axis360_title.axis360Id.

mdnoble73: Oh yep, you're correct.

PR #2500, opened 2025-05-20
DIS-698: using getMainLocation() in self registration because branchc…

lathomas64:
* call /API/RegistrationAPI?method-processSelfRegistration (easiest way is via LiDA and the self register link) without sending up borrower_branchcode
* your registration should still go through and be associated with your default library's main location. 
* before this change you would get an error back from Koha about expecting a string but getting null.

### DIFF
--- a/code/web/Drivers/Koha.php
+++ b/code/web/Drivers/Koha.php
@@ -4699,7 +4699,7 @@ class Koha extends AbstractIlsDriver {
 			$postVariables = $this->setPostFieldWithDifferentName($postVariables, 'gender', 'borrower_sex', $library->useAllCapsWhenSubmittingSelfRegistration);
 			$postVariables = $this->setPostFieldWithDifferentName($postVariables, 'initials', 'borrower_initials', $library->useAllCapsWhenSubmittingSelfRegistration);
 			if (!isset($_REQUEST['borrower_branchcode']) || $_REQUEST['borrower_branchcode'] == -1) {
-				$postVariables['library_id'] = $library->code;
+				$postVariables['library_id'] = $library->getMainLocation()->code;
 			} else {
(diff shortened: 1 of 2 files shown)


code/web/Drivers/Koha.php
```diff
@@ -4699,7 +4699,7 @@ function selfRegister(): array {
 			$postVariables = $this->setPostFieldWithDifferentName($postVariables, 'gender', 'borrower_sex', $library->useAllCapsWhenSubmittingSelfRegistration);
 			$postVariables = $this->setPostFieldWithDifferentName($postVariables, 'initials', 'borrower_initials', $library->useAllCapsWhenSubmittingSelfRegistration);
 			if (!isset($_REQUEST['borrower_branchcode']) || $_REQUEST['borrower_branchcode'] == -1) {
-				$postVariables['library_id'] = $library->code;
+				$postVariables['library_id'] = $library->getMainLocation()->code;
```
mdnoble73: @lathomas64 we should check to see if the library ilsCode is empty before going to the main location. The main location could also be null so we should check that case as well. If we don't have a library code or a main location we should return an error to the user. 

It does look like this was originally checking code rather than ilsCode which was incorrect (since library does not have a code).

mdnoble73 (changes requested):

code/web/Drivers/Koha.php
lathomas64: ilsCode was not working for the partner I had been looking at but it looks like that was a user Error and it should have been blank. updating to look at that field first. We would be getting an error from the processSelfRegistration after the request to Koha failed but updating here to fail earlier and save the trip to Koha servers.

mdnoble73 (approved):

PR #2526, opened 2025-05-28
DIS-552: Fixed Admin Search to Include All Menu Items

LeoStoyanovByWater:
- Improved the admin settings search from the sidebar to include all menu items.


mdnoble73:
@LeoStoyanovByWater this fix does not match the description of DIS-552. DIS-552 is talking about the search bar on a specific page rather than the sidebar which is what this PR seems to address. Also I don't see testing notes here or in the Jira item.

PR #2531, opened 2025-05-30
DIS-903: Fixed Welcome Email Not Being Sent When Registering New User (Symphony)

LeoStoyanovByWater:
- Resolved an issue where when registering a new user via Aspen with a Symphony ILS, no welcome email would be sent.

Test Plan:
1. Assuming you have enabled and set up welcome emails, register a new patron account via Aspen.
   - For clarity, log the API response under `findNewUser()` of `SirsiDynixROA.php` and notice the empty `results` array.
2. Notice that the user is successfully registered, but you receive no email. If you also query the `user` table in the DB, you will see that the new user has not been added.
3. Apply the patch and repeat step 1.
   - If you added the log, notice the now populated `results` array.


code/web/Drivers/SirsiDynixROA.php
```diff
@@ -90,13 +90,16 @@ public function getWebServiceResponse($requestType, $url, $params = null, $sessi
...
+	function findNewUser($patronBarcode, $patronUsername): User|bool {
 		// Creates a new user like patronLogin but looks up user by barcode.
 		// Note: The user pin is not supplied in the Account Info Lookup call.
 		$sessionToken = $this->getStaffSessionToken();
 		if (!empty($sessionToken)) {
 			$webServiceURL = $this->getWebServiceURL();
 			$includeFields = urlEncode("firstName,lastName,privilegeExpiresDate,preferredAddress,preferredName,address1,address2,address3,library,primaryPhone,profile,pin,blockList{owed}");
+			// Symphony needs time to process the full information of the user when registering,
+			// so add a delay to ensure the results array is populated.
+			usleep(500000);
```
mdnoble73: It would be better to do this after submitting the new registration rather than every time we call findNewUser which can happen from other places, i.e. masquerade

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2547, opened 2025-06-03
DIS-915: Checking if the result to getPatronInformation has a Patron

lathomas64:
bypassReadingHistoryUpdate in CarlX.php was displaying some warnings when there was no patron which broke the javascript parsing of updateReadingHistoryTask.java

To test these changes.
* have an Aspen connected to CarlX
* Have at least 1 patron in Aspen that does not exist in CarlX with trackReadingHistory set to 1
* run cron.jar in code/cron
* you should not see errors in the stdout or in the cron log on the website for updateReadingHistory

Prior to this update you would see something like: "2025-06-02 08:55:21 - ERROR: Unable to load patron information for 26185240 exception loading response: org.json.JSONException: A JSONObject text must begin with '{' at 1 [character 2 line 1]"

Sitting on this as a draft until 25.07 is available to make a PR against.


lathomas64:
TODO: when 25.07 is available update the release notes and switch this to be based off that branch and take it out of draft.

mdnoble73:
@lathomas64 can you add release notes for this

mdnoble73 (approved):

PR #2558, opened 2025-06-06
DIS-585: Create Permission Groups to Prevent Permission Conflicts

LeoStoyanovByWater:
- Resolved conflicts between overlapping permissions in Aspen by enforcing mutual exclusivity.
  - Permissions that were previously presented as checkboxes are now shown as dropdowns to allow only one selection.
  - This change prevents roles from being assigned conflicting levels of access for the same functionality.
- Database updates performed to clean up conflicting permissions.
  - Retains only the broadest applicable permission per group.
  - Ensures backward compatibility for users who previously had multiple overlapping permissions.
  - Example: If a role had both "Administer Library Collection Spotlights" and "Administer All Collection Spotlights", only the broader permission (i.e., "Administer All Collection Spotlights") is retained.
- Renamed the "Administer Side Loads" permission to "Administer All Side Loads" for clarity.

Test Plan:
1. Rather than applying the entire patch, upload only `25.07.00.php` to your test instance or server. This is to avoid the new logic in `Permissions.php` that could affect testing when saving permissions changes.
2. First, choose a test role on which you will select all permissions on the Permissions page in the admin interface.
3. Run the “Cleanup Mutually Exclusive Permissions” update under the DB Maintenance page. On the Permissions page, notice that the broadest permissions remain selected while the inferior, related permissions have been unselected.
   - **Note**: Do not reload the Permissions page because that is considered a resubmission by its design, which means that any DB changes made will be overridden by the submission of the previously selected changes. Instead, use the admin sidebar to re-navigate to the Permissions page.
   - Example: If both “Administer Library Collection Spotlights” and “Administer All Collection Spotlights” were selected, only “Administer All Collection Spotlights” will be selected after the DB update.
4. Under the “Cleanup Mutually Exclusive Permissions” update in `25.07.00.php`, you will notice that there are a couple of SQL statements that consider cases of three mutually exclusive permissions. For the following (see list below), to test them, select the second most broad permission and the least broad permission.
   - List by group key: `adminBrowseCategories`, `adminPlacards`, `adminLocations`, `adminEventsAdmin`, and `adminPrivateEvents`.
5. Once again, run the “Cleanup Mutually Exclusive Permissions” update under the DB Maintenance page. On the Permissions page, notice that the second broadest permissions remain selected, while the inferior, related permissions (i.e., least broad) have been unselected.
   - Example: If both “Administer Selected Browse Category Groups” and “Administer Library Browse Categories” were selected, only “Administer Library Browse Categories” will be selected after the DB update.
6. Now, apply the full patch; that is, either checkout the full PR or run the “Create Permission Groups and Mappings” under the DB maintenance page and upload the changed files to your test instance or server. Notice that the mutually exclusive permissions have been grouped into dropdown selectors.
   - For additional testing, select any of the grouped permissions and ensure that those permissions apply.


mdnoble73 (changes requested):
It looks like with this as is, there is no way to set "not allowed" or no permission for a permission group. So for every permission where there is a permission group a role must have one of the values. We need to add a way to deselect all permissions within a permission group.

PR #2592, opened 2025-06-13
DIS-968: Removed Unsupported Scoping Options & `admin_sso` Profile for Two-Factor Auth Settings

LeoStoyanovByWater:
- Removed the “Libraries” and “Patron Types” labels and “Select All” options because users cannot scope Local Administrators to Libraries and Patron Types.
- Removed the option of selecting the “admin_sso” Account Profile, as the authentication flow is handled entirely by the identity provider.

Test Plan: Present in the description of the Jira issue.


code/web/sys/TwoFactorAuthSetting.php
```diff
@@ -79,17 +81,23 @@ static function getObjectStructure($context = ''): array {
...
 		}
 
-		if (!UserAccount::userHasPermission('Administer Two-Factor Authentication')) {
-			unset($structure['libraries']);
-		}
 		return $structure;
 	}
 
 	public function updateStructureForEditingObject($structure) : array {
+		require_once ROOT_DIR . '/sys/Account/AccountProfile.php';
```
mdnoble73, line 92: You shouldn't need to reload the profile here since $this is the active library. You can just check that $this->name == 'admin'

PR #2598, opened 2025-06-16
DIS-980: Replaced Legacy Wikipedia Scraper with REST-API-Based Parser

LeoStoyanovByWater:
WIP


mdnoble73 (changes requested):
Rather than passing if debugging is enabled to the front end. The appropriate information can be returned as part of the AJAX response. With an error response that is either blank or contains data when debugging is on. 

We aren't exposing any sensitive data with the debugging here, but it would be pretty easy to have the JavaScript enable debugging when it shouldn't be if it's just a global javascript variable.

ghost:
@mdnoble73 Debug messages are now returned through the AJAX response.

PR #2611, opened 2025-06-17
DIS-979_milestone_email_conditions

AlexanderBlanchardAC:
https://aspen-discovery.atlassian.net/browse/DIS-979


mdnoble73:
Can you add the release notes to 25.06.01.MD and check the whitespace to be tabs rather than spaces?

PR #2633, opened 2025-06-20
DIS-1007: Add "Author" Sort Option for User Lists

LeoStoyanovByWater:
- Added the option to sort user lists by "Author".

Test Plan:
1. Apply the patch.
2. Navigate to a user list, click “Sort by,” and select “Author.” Notice that the list is sorted alphabetically by the author’s name.

### DIFF
--- a/code/web/sys/UserLists/UserList.php
+++ b/code/web/sys/UserLists/UserList.php
@@ -58,8 +58,8 @@ class UserList extends DataObject {
 		'title' => 'title ASC',
 		'dateAdded' => 'dateAdded ASC',
 		'recentlyAdded' => 'dateAdded DESC',
-		'custom' => 'weight ASC',
-		// this puts items with no set weight towards the end of the list
+		'custom' => 'weight ASC', // Puts items with no set weight towards the end of the list.
+		'author' => '',
 	];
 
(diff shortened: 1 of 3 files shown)


code/web/sys/UserLists/UserList.php
```diff
@@ -371,24 +371,66 @@ function removeAllListEntries(bool $updateBrowseCategories = true) : void {
...
-		//Get all entries for the list
+	public function getListRecords(
+		int $start, int $numItems, bool $allowEdit, string $format, string $citationFormat = null,
+		string $sortName = null, bool $forLiDA = false, float|int $appVersion = 0
+	): array {
 		if ($sortName == null) {
 			$sortName = $this->defaultSort;
 		}
+
+		// Because the DB does not persist an "author" column on the list itself,
```
mdnoble73, line 392: Has this been tested with larger lists (100s of entries) with a mix of different sources, i.e. Grouped Works, Events, Lists, etc. This could be a pretty time consuming operation to search for them all by ID for sorting.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2634, opened 2025-06-23
DIS-999 prevent theme id error

Chloe070196:
https://aspen-discovery.atlassian.net/browse/DIS-999


code/web/sys/Theming/Theme.php
```diff
@@ -3630,4 +3630,27 @@ public function updateStructureForEditingObject($structure) : array {
...
+	public function getOrSetDefaultTheme(): Theme {
+		if ($this->_defaultTheme != null) {
+			return $this->_defaultTheme;
+		}
+
+		$defaultTheme = new Theme;
+		$defaultTheme->themeName = 'default';
+
+		if (!$defaultTheme->find(true)) {
+			$defaultTheme->displayName = 'default';
```
mdnoble73: @Chloe070196 in this case if we don't find a theme named "default" I think we could fall back to just using the first theme in the system. 

Leo recently made a change to prevent deleting the theme named "default" so it should exist on most systems and if it doesn't the library would probably prefer the theme they have customized already to a generic default theme?

Chloe070196: I think this makes a lot of sense, thanks! I'll update the PR accordingly

mdnoble73 (approved):

PR #2672, opened 2025-06-26
DIS-657: Prevent Uploaded File Overwrite Part II

K-Alette:
- Update naming conventions for multiple areas where images are uploaded to prevent files being overwritten
  - Web builder images will now have the prefix "web_builder_image" followed by the object id
  - Web builder PDFs will now have the prefix "web_builder_PDF" followed by the object id
  - All 25 images uploaded in themes will each have their own special prefix followed by the theme id
  - Web resource logos will now have the prefix "web_resource_image" followed by the object id
  - Placard images will now have the prefix "placard_image" followed by the object id unless they are using a linked Web Resource for the image
  - Update existing images with a DB update to follow the new naming convention
- Updat PDFs.php and Images.php to account for local admins without a home library
- Update release notes

### DIFF
--- a/code/web/release_notes/25.07.00.MD
+++ b/code/web/release_notes/25.07.00.MD
@@ -66,6 +66,7 @@
 
 ### Web Builder Updates
 - Correct notice when deleting cells from a row. (DIS-494) (*MDN*)
+- Generate thumbnail for uploaded PDFs based on the first page of the PDF rather than the last page. (DIS-1029) (*MDN*)
 
 ### Other Updates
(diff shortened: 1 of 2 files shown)


mdnoble73 (changes requested):
Found a couple of little things so far that I wanted to drop in before I got too far:
XLarge images don't get moved for uploaded images since the path is x-large.
For uploaded files, we should move the thumbnail as well if it exists during the conversion process.
After uploading a file, the original file and thumbnail still exist, after the file and thumbnail based on ID are created.
When uploading an image, the full size gets renamed, but all of the generated images stay with the original name. The uploaded image in the full directory also doesn't get deleted.

PR #2675, opened 2025-06-26
DIS-1027: Fixed Location Display Order in Search Results Availability

LeoStoyanovByWater:
- Refined pickup location display order so that search results availability now correctly prioritizes pickup locations in the logical sequence: preferred pickup location, alternate pickup location 1, alternate pickup location 2, home library, and so on.
- Resolved an issue where the patron's home library would always be the top result when selecting at what location to place a hold.

Test Plan:
1. Log into a patron account and check your “My Preferences” page. Keep note of your home library and pickup location preferences; set any that might be missing.
2. Search the catalog and click “Place Hold” on an ILS record. Notice that it does not display the pickup location options in the following order: preferred pickup location, alternate location 1, alternate location 2, then home library, etc.
3. Apply the patch and repeat step 1. Notice that the pickup location options are now in the aforementioned order.


mdnoble73 (changes requested):
See comment in Jira

PR #2688, opened 2025-06-27
DIS-1022 Cloud Library fix redirect handling

LiYanjun19:
If libraries do not require a PIN in their Cloud Library, patrons usually only need to use a barcode to sign in the Cloud Library website. When patrons check out Cloud Library records on Aspen, and click “Open In CloudLibrary”, Aspen will redirect patrons to Cloud Library, but see an error` {"error":"Invalid Token"}`

According to Cloud Library support, Aspen should not be passing a password field to {$userInterfaceUrl}/login if the library does not require a PIN in Cloud Library. 

To recreate:
1. Set Cloud Library to do not require a PIN in Cloud Library website 
2. Check out CL records on Aspen
3. Go to “Check out Titles” > Cloud Library > click “Open In CloudLibrary” 
4. See the error {"error":"Invalid Token"}

For {$webPatronUrl}/login endpoint:
1. If Aspen submits a correct barcode and password, or only a correct barcode for the library does not require PIN, or
2. If Aspen includes a password for a library that does not require a PIN, or
3. If Aspen submits an incorrect barcode and password

In all three cases, the endpoint responds with a 200 status code and the same response structure. The only difference is that the session ID is invalid in cases 2 and 3. There doesn’t appear to be any clear indicator in the response that the session ID is invalid or not until we try to make a follow-up call to `webPatronUrl/ebooks/{recordId}?auth_cookie={sessionId}`
_Contacted Cloud Library to see if they can make any changes on their side._ 

Before Cloud Library could fix it or not, the current flow is:
try `/login` with password -> get sessionid -> try to call redirectUrl with sessionid -> gets 200 code, redirect patrons to the url -> if we get 400 code -> try `/login` **without** password -> get sessionid again -> try to call redirectUrl with new sessionid ->  gets 200 code, redirect patrons to the url -> if we get 400 again, redirect patron to the record detail page. 

To test:
1. Set Cloud Library to do not require a PIN in Cloud Library website 
2. Check out CL ebook record and mp3 record on Aspen
3. Go to “Check out Titles” > Cloud Library > click “Open In CloudLibrary” 
4. See the error {"error":"Invalid Token"}
5. Apply PR 
6. Click "Open in CloudLibrary" again for both ebook record and mp3 audiobook record on Aspen
7. Can access to the record without issue. See no obvious delay in between "clicking 'Open in CloudLibrary' button" and "a new tab opening" (excluding the time it takes for the page to fully render).

### DIFF
--- a/code/web/Drivers/CloudLibraryDriver.php
+++ b/code/web/Drivers/CloudLibraryDriver.php
@@ -986,35 +986,61 @@ class CloudLibraryDriver extends AbstractEContentDriver {
 	public function redirectToCloudLibrary(User $patron, CloudLibraryRecordDriver $recordDriver) {
 		$settings = $this->getSettings($patron);
 		$userInterfaceUrl = $settings->userInterfaceUrl;
+
+		//Setup the default redirection paths
+		$redirectDetailUrl = $userInterfaceUrl . '/detail/' . $recordDriver->getId();
+
+		//Generate redirect URL to Cloud Library with password
+		$redirectUrl = $this->generateRedirectUrl($patron, $recordDriver, true);
+
+		if (!$redirectUrl) {
+			//Some libraries do not require PIN to login cloud library
+			//Generate redirect URL to cloud library without password
+			$redirectUrl = $this->generateRedirectUrl($patron, $recordDriver, false);
+		}
+
+		if (!$redirectUrl) {
+			//Fall back to record detail page
+			$redirectUrl = $redirectDetailUrl;
+		}
+
+		header('Location:' . $redirectUrl);
+		die();
+
+	}
+
+	private function generateRedirectUrl(User $patron, CloudLibraryRecordDriver $recordDriver, $includePassword = false) {
+		$settings = $this->getSettings($patron);
+		$userInterfaceUrl = $settings->userInterfaceUrl;
 		if (substr($userInterfaceUrl, -1) == '/') {
 			$userInterfaceUrl = substr($userInterfaceUrl, 0, -1);
 		}
 
-		//Setup the default redirection paths
-		//These URLs no longer work
-//		if ($recordDriver->getPrimaryFormat() == 'MP3') {
-//			$redirectUrl = $userInterfaceUrl . '/AudioPlayer/' . $recordDriver->getId();
-//		} else {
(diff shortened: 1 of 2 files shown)


code/web/Drivers/CloudLibraryDriver.php
```diff
@@ -986,35 +986,61 @@ public function getItemStatus($itemId, $patron) {
...
+		//Setup the default redirection paths
+		$redirectDetailUrl = $userInterfaceUrl . '/detail/' . $recordDriver->getId();
+
+		//Generate redirect URL to Cloud Library with password
+		$redirectUrl = $this->generateRedirectUrl($patron, $recordDriver, true);
+
+		if (!$redirectUrl) {
+			//Some libraries do not require PIN to login cloud library
+			//Generate redirect URL to cloud library without password
+			$redirectUrl = $this->generateRedirectUrl($patron, $recordDriver, false);
```
mdnoble73, line 999: Do we need to make multiple calls to Cloud Library to determine this or is there a setting we can use to tell when a password is required or not? Maybe we can add something to the settings?

mdnoble73 (changes requested):

code/web/Drivers/CloudLibraryDriver.php
LiYanjun19: Thank you! I considered both approaches: 1) adding a setting, 2) adding auto fallback logic 
I still chose to add fallback logic because 
1. Adding a setting creates a dependency on libraries configuring it correctly. If a library misconfigures the setting initially or changes their CloudLibrary PIN policy without updating Aspen, patrons experience failed logins until we resolves the configuration mismatch. 
2. The extra `/login` API call only occurs for libraries that do not require the PIN. I believe most of the libraries do require the PIN.  Currently we make 2 or 4 API calls per "Open in CloudLibrary" action (depending whether libraries require PIN or not) as we also verify the sessionId before redirecting -- The `/login `endpoint doesn't provide a clear success indicator. 
3. I am communicating with CloudLibrary team to see if they can enhance the `/login` endpoint to provide clearer authentication status indicators. If they can make some changes, we can reduce this to 1 or 2 calls per action. 
4. No noticeable latency for user experience during my test.  

Please let me know what you think.

PR #2737, opened 2025-07-09
DIS-1019: Add toast notifications to Community Engagement campaign milestones progress

ammopt:
https://aspen-discovery.atlassian.net/jira/software/c/projects/DIS/boards/2?selectedIssue=DIS-1019

Test plan, using adb with ktd:

Enable Community Engagement module.
Create a reward. Create a milestone of type 'Rating'. Create a campaign that starts at a past date and ends at a future date. Available for all libraries and all patron categories. Add the previously created milestone to this campaign with a goal e.g. 10.
Masquerade as a real patron, I've used '23529000035676' on my testing (Henry Acevedo).
On the user menu, click 'Your campaigns' and enroll on the campaign created in 2)
Do a search e.g. 'music' and on the search results, pick one and click 'Add a review'. Submit the review with some rating and any text.
Notice a toast notification displaying a milestone progress message is shown. You may have to wait up to 10 seconds for this notification to show as that is the SSE heartbeat frequency. If you move to any other URL before the notification is shown, it'll display instantly (because it restarts the SSE connection on a new page visit).
The above is the baseline testing of the functionality. Additional testing:

Create other types of milestone types (checkouts, holds)
Create a manual milestone. Verify that the toast notifications show for manual progress
Complete a milestone. Verify that a 'Milestone completed' toast notification is shown.
Complete a campaign. Verify that a 'Campaign completed' toast notification is shown.
Create a milestone that allows progress beyond 100%. Verify the 'milestone progress' toast notification is also shown for every progress made beyond 100%.
Continuing from 3. Verify that the 'Campaign completed' is only shown when the campaign is completed, and not for every milestone progress beyond 100%.

### DIFF
--- a/code/web/services/MyAccount/AJAX.php
+++ b/code/web/services/MyAccount/AJAX.php
@@ -10055,6 +10055,12 @@ class MyAccount_AJAX extends JSON_Action {
 						$userCampaign->campaignId = $campaignMilestoneProgressEntry->ce_campaign_id;
 						$userCampaign->find(true);
 
+						$unwantedOverflowProgress = $campaignMilestoneUsersProgress->progress > $campaignMilestone->goal &&  !$milestone->progressBeyondOneHundredPercent;
+						$wantedOverflowProgress = $campaignMilestoneUsersProgress->progress > $campaignMilestone->goal &&  $milestone->progressBeyondOneHundredPercent;
+						if( $unwantedOverflowProgress ){
+							exit();
+						}
+
 						# Handle milestone progress notification
 						echo "event: ce_notification\n";
 						echo "data: " . json_encode(
@@ -10078,7 +10084,7 @@ class MyAccount_AJAX extends JSON_Action {
 						) . "\n\n";
 
 						# Handle milestone completion notification
-						if ($campaignMilestoneUsersProgress->progress >= $campaignMilestone->goal ) {
+						if ($campaignMilestoneUsersProgress->progress >= $campaignMilestone->goal && !$wantedOverflowProgress) {
 							echo "event: ce_notification\n";
 							echo "data: " . json_encode(
 								array(
@@ -10102,7 +10108,7 @@ class MyAccount_AJAX extends JSON_Action {
 						}
 
 						# Handle campaign completion notification
-						if ($userCampaign->completed) {
+						if ($userCampaign->completed && !$wantedOverflowProgress) {
 							echo "event: ce_notification\n";
 							echo "data: " . json_encode(
 								array(
--- a/code/web/services/MyAccount/AJAX.php
+++ b/code/web/services/MyAccount/AJAX.php
@@ -9986,6 +9986,160 @@ class MyAccount_AJAX extends JSON_Action {
 		}
 	}
 
+	/**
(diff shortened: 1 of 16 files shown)


code/web/services/MyAccount/AJAX.php
```diff
@@ -9986,6 +9986,166 @@ private function processCampaignMilestones($entity, $campaignId, $entityDate, $e
 		}
 	}
 
+	/**
+	 * Sends server-sent events (SSE) notifications about community engagement milestones and campaigns.
+	 */
+
+	public function CommunityEngagementSSE() {
+		$debug = false; // Set to true to enable debug mode. true for dev only.
```
mdnoble73, line 9994: It looks like this is permanently keeping a connection open with the server. Is there a reason to use this pattern rather than polling for events from the patron side?

mdnoble73 (changes requested):
Followed the test plan and never saw a toast appear. Looked at the network data and never saw data returned from the CommunityEngagementSSE request.

ammopt:
> Followed the test plan and never saw a toast appear. Looked at the network data and never saw data returned from the CommunityEngagementSSE request.

Hi Mark, thank you in advance for the time spent looking here.
I've attached a short video (1:40 min) of me going through the test plan on a fresh adb and it appears to be working as described on my end. Can you please add additional information so that I can try and help further?

Some things worth double-checking:
1) Was the patron correctly enrolled in the campaign?
2) Was the campaign active, i.e. started on a past date and ended on a future date?

https://github.com/user-attachments/assets/490db0b1-16f2-4f2c-90a2-62bd3ade8e47

code/web/services/MyAccount/AJAX.php
ammopt: Hi Mark, again, thank you for looking here and appreciate your comments. This is a fair observation, I'll try and explain my thought process.

> It looks like this is permanently keeping a connection open with the server.

For as long as the client's browser keeps the session active. Once the browser page is closed so is the connection.

I started with the SEE approach as the original plan was to only send updates to the client when a change in the database occurred (not on a loop). It became apparent that that wouldn't be possible without adding a message queue type of component (RabbitMQ or Redis) to the stack, so I decided to not pursue that. I couldn't find a good enough reason to drop the SSE based architecture in favor of polling as it's effectively:

Polling:
- Client sends a request to the server every 10 seconds
- Server responds with either 'no update' or 'update available'
- 6 requests per minute

SSE:
- Client establishes a single connection and listen for updates
- Server only pushes updates to the client when they're available
- Up to 6 responses per minute

I'll happily follow-up in changing the flow of the solution to polling instead of SSE if that is critical, but I hope the above at least provides context as to why it is how it is.

mdnoble73:
@ammopt I figured it out. I had 2 issues:
1) I have linked users setup and accidentally opted one of them in rather than the main user (confusing since they show above the main user. 
2) There is a bug checking for existing milestone progress for user ratings (https://aspen-discovery.atlassian.net/browse/DIS-1190). I'll put a fix in for that shortly.

code/web/services/MyAccount/AJAX.php
mdnoble73: Both seem to work fine, I'm good with this approach.

PR #2744, opened 2025-07-10
DIS-1087: "Share Holdings" Setting for Talpa Search

Lauren-at-LibraryThing:
This update is intended to give librarians the ability to manually trigger the talpaRecalculationCron in order to send their catalog/groupedWorkIDs to Talpa for immediate proccessing. This adds a "Share Holdings" section to Talpa Settings.

### DIFF
--- a/code/web/services/Talpa/TalpaSettings.php
+++ b/code/web/services/Talpa/TalpaSettings.php
@@ -3,6 +3,7 @@
 require_once ROOT_DIR . '/Action.php';
 require_once ROOT_DIR . '/sys/Talpa/TalpaSettings.php';
 require_once ROOT_DIR . '/services/Admin/ObjectEditor.php';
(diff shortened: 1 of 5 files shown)


code/web/services/Talpa/TalpaSettings.php
```diff
@@ -74,5 +75,168 @@ function getActiveAdminSection(): string {
 	function canView(): bool {
 		return UserAccount::userHasPermission('Administer Third Party Enrichment API Keys');
 	}
+
+	/**
+	 * Override the editObject method to trigger the talpa recalculation cron when the setting is enabled
+	 */
+	function editObject($objectAction, $structure) {
```
mdnoble73, line 82: Rather than duplicating editObject here you can override the insert and update methods within code/web/sys/Talpa/TalpaSettings.php or we can add an action button to the form to trigger sending the holdings.

mdnoble73 (changes requested):

PR #2781, opened 2025-07-18
DIS-1123: Document getListTitles from ListAPI in the API docs

skora-cpl:
Hi, 

This is a draft of creating a documentation for the ListAPI, specifically the getListTitles function, which enables developers to obtain a list of all works and events within a user created list. 

See the list API's getListTitles function in action at https://clevelandreads.com/recommended-reads/ 

Being my first PR and still learning about Aspen, 

I have several questions about OpenAPI formatting and the listAPI: 

**OpenApi formatting questions**

- how to cite items that items may sometimes be a empty string ? 
like  "length": "",
example https://gist.github.com/skora-cpl/76d9b8d0add53ed69d30dbff85a076e3#file-gistfile1-txt-L21

- ~~How to state if a parameter is required or optional ?~~ (see: https://github.com/Aspen-Discovery/aspen-discovery/blob/8defde7966e772e04e9b8ef9001f2f48e9dbaba5/code/web/openapi/aspen_openapi.json#L77) 

- What order should items be placed in the response ? Alphabetical order or in the order that they are in response ? 

For example, at https://github.com/Aspen-Discovery/aspen-discovery/blob/beb0861174e7c74eb059d38a87a8995d6a9423cc/code/web/openapi/aspen_openapi.json#L369 I alphabetized them based on the string name; but the raw api JSON returns them in a different order - e.g. within the titles array, the order appears as so https://gist.github.com/skora-cpl/716ad5396a60c16e4d3cff115db73b4d)

- What kind of type (in OpenAPI speak) are URLs; like at https://gist.github.com/skora-cpl/76d9b8d0add53ed69d30dbff85a076e3#file-gistfile1-txt-L187 

I represented it as a string at https://github.com/Aspen-Discovery/aspen-discovery/blob/beb0861174e7c74eb059d38a87a8995d6a9423cc/code/web/openapi/aspen_openapi.json#L381

**Questions specific to Aspen-Discovery about getListTitles**

- for numtitles parameter, is there a maximum value (i don't have any lists that have a high value) 
- When I added a particular edition to a list; the entire grouped_Work was included; is this intentional ? (see https://search.cpl.org/MyAccount/MyList/84147 when I added a particular edition like https://search.cpl.org/OverDrive/kindle:9816fff8-ff33-4c2f-be39-e87488f685e4/Home to a list; all of the editions were added) (this is more out of curiousity, i guess, i don't necessarily see this as a bug). 
- Are the "grouped_work" and "event" the only possible values for recordType ? 
- are the images in small_image always 100px height ? 
- are the images in image always 200px height  ?
(these names go back to getBookcoverUrl but I couldn't find if the height is always specified) 
- What's the difference between ID and shortID ? In all of my API responses, each title's id and shortID are identical; is that always true ? 

- How should I provide extra contextual information like a list needs to be made public for a successful request; (unless you specify the user's name and password ? otherwise the result will be  `{"success":false,"message":"The user was invalid.  A valid user must be provided for private lists."}}` 

**Other notes:** 

I found how to write for openAPI a bit daunting at first and the learning curve is not easy. 

Having some clearer guidelines would be really helpful and an additional doc would be helpful for future contributors. 

For example, including where the documentation file can be found (code/web/openapi/aspen_openapi.json); 
what the name of the spec is (openAPI 3.1); maybe a couple links to what the OpenAPI is and some examples, recommending editing the raw JSON file itself, @myr-onl also mentioned that they "made a tag for each (necessary for docs) php file in /code/web/services/API/ and was using the notes inside the php files to write things." 

Others can build upon what I've shared for writing other API documentation. 

I have been using and contributing to open source projects for about 15 years so I understand that incentives (monetary, personal) are usually not aligned to work on documentation and that creating (and maintaining) documentation requires considerable more effort and time than most people perceive and its "value" and more energy regarding docs has been towards the general docs around confluence and aspen discovery (Which has a larger audience and is extremely useful!) 

**Resources**

I personally learn best from examples and couldn't find many existing examples out there that are written in JSON (seems that many openAPI examples are in yaml) so I found the following very helpful

https://api.weather.gov/openapi.json
https://www.weather.gov/documentation/services-web-api#/

(NOTE: the weather.gov openAPI uses openAPI 3.0 spec; Aspen uses 3.1 - there are some differences betwen them https://beeceptor.com/docs/concepts/openapi-what-is-new-3.1.0/)

https://json-schema.org/blog/posts/validating-openapi-and-json-schema#what-is-a-json-schema-dialect

### DIFF
--- a/code/web/openapi/aspen_openapi.json
+++ b/code/web/openapi/aspen_openapi.json
@@ -378,7 +378,8 @@
         "properties": {
           "author": {
             "type": "string",
-            "description": "Author of the work"
+            "description": "Author of the work", 
+            "example": "Ellison, Ralph"
           },
           "description": {
             "type": "string",
@@ -389,7 +390,13 @@
             "description": "mediums (book, e-book, cassette) that the work is available",
             "items": {
               "type": "string"
-            }
+            }, 
+            "example": [
+                    "Book",
+                    "eAudiobook",
+                    "Audiobook CD",
+                    "eBook"
+                ]
           },
           "id": {
             "type": "string",
@@ -414,17 +421,22 @@
             "type": "string",
             "description": "publisher of the work"
           },
+          "recordType": {
+            "type": "string",
+            "description": "TO ADD, typically grouped_work"
+          },
           "small_image": {
             "type": "string",
             "description": "URL partial for an image of 100px height"
           },
           "title": {
             "type": "string",
-            "description": "Title of the work"
+            "description": "Title of the work",
+            "example": "Invisible Man"
           },
           "titleURL": {
             "type": "string",
-            "description": "Full URL to access the title in the catalog "
+            "description": "Full URL to access the title in the catalog"
           }
         }
       },
@@ -440,7 +452,8 @@
               },
               "listTitle": {
                 "type": "string",
-                "description": "Title of the user's list"
+                "description": "Title of the user's list", 
+                "example": "Recommended Reads of Summer Classics"
               },
               "listDescription": {
                 "type": "string",
--- a/code/web/openapi/aspen_openapi.json
+++ b/code/web/openapi/aspen_openapi.json
@@ -75,7 +75,7 @@
             },
             "description": "The ID of the list to return information about",
             "required": true, 
-            "example": "42"
+            "example": "44114"
           },
           {
             "in": "query",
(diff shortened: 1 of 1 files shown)


skora-cpl:
An example of the API visualized: 

image

PR #2800, opened 2025-07-24
DIS-131: Fixed Staff View & Reviews from GoodReads Accordions to Load When Set to Open by Default

LeoStoyanovByWater:
- "Staff View" and "Reviews from GoodReads" accordions now loads properly when set to open automatically in Grouped Work Display settings.
- Removed unused `Springshare/moredetails-accordion.tpl`.

Test Plan:
1. In Grouped Work Display settings, disable “Collapse by Default” for the "Staff View" and "Reviews from GoodReads".
2. Navigate to a grouped work’s page and see that the Staff View never loads and just displays “Loading…”; same is true for "Reviews from GoodReads".
3. Apply the patch and reload the page. Notice that both load.


code/web/RecordDrivers/GroupedWorkDriver.php
```diff
@@ -1571,11 +1571,7 @@ public function getMoreDetailsOptions() {
 		if ($interface->getVariable('showStaffView')) {
 			$moreDetailsOptions['staff'] = [
 				'label' => 'Staff View',
-				'onShow' => "AspenDiscovery.GroupedWork.getStaffView('{$this->getPermanentId()}');",
-				'body' => '<div id="staffViewPlaceHolder">' . translate([
-						'text' => 'Loading Staff View.',
-						'isPublicFacing' => true,
-					]) . '</div>',
+				'body' => $interface->fetch($this->getStaffView()),
```
mdnoble73: We do still want this loaded asynchronously since the payload for the staff view can be large and generating it can be an expensive process.

PR #2804, opened 2025-07-24
DIS-1153: Add "Allow Material Requests Branch Choice" for ILS Material Requests

LeoStoyanovByWater:
- Added "Allow Material Requests Branch Choice" setting for the ILS Request System, allowing patrons to select which branch should receive their materials request instead of defaulting to their home branch.

Test Plan:
1. Apply the patch and enable the “ILS Request System” as the Material Request System.
2. Log in or masquerade as a patron account that can make material requests. Notice that multiple branch locations can be chosen in the form.

### DIFF
--- a/code/web/Drivers/Koha.php
+++ b/code/web/Drivers/Koha.php
@@ -5164,9 +5164,10 @@ class Koha extends AbstractIlsDriver {
 		}
 
 		global $interface;
-		$allowPurchaseSuggestionBranchChoice = $this->getKohaSystemPreference('AllowPurchaseSuggestionBranchChoice');
+		global $library;
+		$allowMaterialRequestsBranchChoice = $library->allowMaterialRequestsBranchChoice;
 		$pickupLocations = [];
-		if ($allowPurchaseSuggestionBranchChoice == 1) {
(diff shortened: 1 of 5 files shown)


code/web/Drivers/Koha.php
```diff
@@ -5164,9 +5164,10 @@ function getNewMaterialsRequestForm(User $user) {
 		}
 
 		global $interface;
-		$allowPurchaseSuggestionBranchChoice = $this->getKohaSystemPreference('AllowPurchaseSuggestionBranchChoice');
+		global $library;
+		$allowMaterialRequestsBranchChoice = $library->allowMaterialRequestsBranchChoice;
```
mdnoble73, line 5177: Did this system preference get removed from Koha? Do we need to only allow this if the option is enabled in both Koha and the Library System?

mdnoble73 (changes requested):

code/web/Drivers/Koha.php
ghost: I mention in the Jira issue [DIS-1153](https://aspen-discovery.atlassian.net/browse/DIS-1153) that this System Preference was removed. Today, I added a reference to the Bugzilla in the Jira.

PR #2812, opened 2025-07-28
DIS-1076: Object Restoration Functionality

LeoStoyanovByWater:
- Object Restorations
  - The following objects can be restored after their deletion:
    - User Lists
      - Patrons may choose "Opt Out of Soft Deletion" to permanently delete them immediately.
    - Basic Pages 
    - Custom Pages (Portal Pages)
    - Custom Forms 
    - Web Resource Pages 
    - Web Resources 
    - Image Uploads (hard-deleting removes the file from the server)
    - File Uploads (hard-deleting removes the file from the server)
    - Placards
  - A daily process permanently deletes any items that have been soft-deleted for more than 30 days.
  - A new permission, "Administer Object Restoration," is required to access the Restore Deleted Objects page.

### DIFF
--- a/code/web/services/Admin/ObjectEditor.php
+++ b/code/web/services/Admin/ObjectEditor.php
@@ -869,13 +869,17 @@ abstract class ObjectEditor extends Admin_Admin {
 			foreach ($_REQUEST['selectedObject'] as $id => $value) {
 				if ($index == 1) {
 					$object1 = $this->getExistingObjectById($id);
-					$object1EditUrl = "/{$this->getModule()}/{$this->getToolName()}?objectAction=edit&id=$id";
-					$interface->assign('object1EditUrl', $object1EditUrl);
+					if ($this->showEditButtons()) {
+						$object1EditUrl = "/{$this->getModule()}/{$this->getToolName()}?objectAction=edit&id=$id";
+						$interface->assign('object1EditUrl', $object1EditUrl);
+					}
 					$index = 2;
 				} else {
 					$object2 = $this->getExistingObjectById($id);
-					$object2EditUrl = "/{$this->getModule()}/{$this->getToolName()}?objectAction=edit&id=$id";
-					$interface->assign('object2EditUrl', $object2EditUrl);
+					if ($this->showEditButtons()) {
+						$object2EditUrl = "/{$this->getModule()}/{$this->getToolName()}?objectAction=edit&id=$id";
+						$interface->assign('object2EditUrl', $object2EditUrl);
+					}
 				}
 			}
 			if ($object1 == null || $object2 == null) {
@@ -890,6 +894,10 @@ abstract class ObjectEditor extends Admin_Admin {
 			$interface->assign('error', 'Please select two objects to compare');
 		}
 
+		$interface->assign('showEditButtons', $this->showEditButtons());
+		$interface->assign('showReturnToList', $this->getToolName() === 'ObjectRestorations');
+		$interface->assign('module', $this->getModule());
+		$interface->assign('toolName', $this->getToolName());
 		$interface->setTemplate('../Admin/compareObjects.tpl');
 	}
 
@@ -1020,6 +1028,9 @@ abstract class ObjectEditor extends Admin_Admin {
 			//Work with an existing record
 			global $interface;
 			$curObject = $this->getExistingObjectById($id);
+			if (!$curObject) {
+				AspenError::raiseError('The object with ID ' . $id . ' does not exist.');
+			}
 			$interface->assign('curObject', $curObject);
 			$interface->assign('id', $id);
 			$displayNameColumn = $curObject->__displayNameColumn;
@@ -1049,6 +1060,9 @@ abstract class ObjectEditor extends Admin_Admin {
 				$objectHistory[] = clone $historyEntry;
 			}
 			$interface->assign('objectHistory', $objectHistory);
+			$interface->assign('showEditButtons', $this->showEditButtons());
+			$interface->assign('module', $this->getModule());
+			$interface->assign('toolName', $this->getToolName());
 			$this->display('../Admin/objectHistory.tpl', $title);
 			exit();
 		}
@@ -1373,6 +1387,15 @@ abstract class ObjectEditor extends Admin_Admin {
 	}
 
 	protected function showHistoryLinks() {
+		return true;
+	}
+
+	/**
+	 * Control whether edit buttons should be shown in history and compare views.
+	 * Purpose: Override it to hide edit functionality when appropriate.
+	 * @return bool True if edit buttons are enabled, false otherwise.
+	 */
+	protected function showEditButtons(): bool {
 		return true;
 	}
 
--- a/code/web/services/Admin/ObjectRestorations.php
+++ b/code/web/services/Admin/ObjectRestorations.php
@@ -1,0 +1,514 @@
+<?php
+
+use JetBrains\PhpStorm\NoReturn;
+
+require_once ROOT_DIR . '/services/Admin/ObjectEditor.php';
+require_once ROOT_DIR . '/sys/ObjectRestoration.php';
+
+/**
+ * Displays every soft-deleted row across a curated set of classes
+ * and lets privileged staff restore or permanently purge them.
+ *
+ * To add a new object for restoration:
+ * 1. Ensure the target DataObject implements `supportsSoftDelete()` returning
+ *    `true` and that its table contains `deleted` (TINYINT(1)), `dateDeleted` (INT),
+ *    and `deletedBy` (INT(11)) columns.  The base `DataObject` already handles
+ *    the logic.
+ * 2. Append an entry to `self::$managedClasses` below:
+ *        `'NewClass' => [
+ *            'titleColumn' => 'title',
+ *            'classFile'   => '/sys/Path/To/NewClass.php',
+ *        ],`
+ *    Use the path from `ROOT_DIR` so the class can be lazy-required.
+ * 3. Modify the target DataObject's `delete()` method to prevent deletions of
+ *    object dependencies unless `$useWhere` is `true`, which indicates a hard deletion.
+ *    If the target DataObject has no `delete()`, you may have to override `purgeExpired`
+ *    for proper hard-deletion cleanup (e.g., ImageUpload).
+ * 4. The batch-purge cron and UI will pick it up the next time the admin page is loaded.
+ */
+class Admin_ObjectRestorations extends ObjectEditor {
+	private static array $managedClasses = [
+		'UserList'              => ['titleColumn' => 'title', 'classFile' => '/sys/UserLists/UserList.php'],
+		'BasicPage'             => ['titleColumn' => 'title', 'classFile' => '/sys/WebBuilder/BasicPage.php'],
+		'PortalPage'            => ['titleColumn' => 'title', 'classFile' => '/sys/WebBuilder/PortalPage.php'],
+		'CustomForm'            => ['titleColumn' => 'title', 'classFile' => '/sys/WebBuilder/CustomForm.php'],
+		'CustomWebResourcePage' => ['titleColumn' => 'title', 'classFile' => '/sys/WebBuilder/CustomWebResourcePage.php'],
+		'WebResource'           => ['titleColumn' => 'name',  'classFile' => '/sys/WebBuilder/WebResource.php'],
+		'ImageUpload'           => ['titleColumn' => 'title', 'classFile' => '/sys/File/ImageUpload.php'],
+		'FileUpload'            => ['titleColumn' => 'title', 'classFile' => '/sys/File/FileUpload.php'],
+		'Placard'               => ['titleColumn' => 'title', 'classFile' => '/sys/LocalEnrichment/Placard.php'],
+	];
+	private array $cachedRows;
+
+	public function __construct() {
+		parent::__construct();
+		$this->cachedRows = $this->buildRows();
+	}
+
+	function getObjectType(): string { return 'ObjectRestoration'; }
+	function getToolName(): string  { return 'ObjectRestorations'; }
+	function getModule(): string    { return 'Admin'; }
+	function getPageTitle(): string { return 'Restore Deleted Objects'; }
+	function getDefaultSort(): string { return 'deletedOn desc'; }
+	function getPrimaryKeyColumn(): string { return 'compositeId'; }
+	function getIdKeyColumn(): string { return 'compositeId'; }
+	function canView(): bool { return UserAccount::userHasPermission('Administer Object Restoration'); }
+	function canAddNew(): bool { return false; }
+	function canDelete(): bool { return true; }
+	function canEdit(DataObject $object): bool { return false; }
+	function canCopy(): bool { return false; }
+	function canBatchDelete(): bool { return false; }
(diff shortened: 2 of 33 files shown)


code/web/sys/DB/DataObject.php
```diff
@@ -126,6 +128,19 @@ public function find($fetchFirst = false, $requireOneMatchToReturn = false): boo
 		}
 
 		global $aspen_db;
+		if (!isset($aspen_db)) {
+			return false;
+		}
+
+		// Soft-delete support: automatically hide rows where deleted = 1 unless caller opted-in.
+		if (method_exists($this, 'supportsSoftDelete') && $this->supportsSoftDelete()) {
```
mdnoble73: Why do we need to check if the method exists here? The method is defined below so it should always be present.

code/web/sys/DB/DataObject.php
```diff
@@ -614,6 +636,16 @@ public function count() {
 		if (!isset($aspen_db)) {
 			return false;
 		}
+
+		// Soft-delete support: automatically hide rows where deleted = 1 unless caller opted-in.
+		if (method_exists($this, 'supportsSoftDelete') && $this->supportsSoftDelete()) {
```
mdnoble73: Why do we need to check if the method exists here? The method is defined below so it should always be present.

code/web/sys/DB/DataObject.php
```diff
@@ -614,6 +636,16 @@ public function count() {
...
+			$includeDeleted = property_exists($this, '_includeDeleted') ? $this->_includeDeleted : false;
```
mdnoble73, line 642: Would it be good to have an abstract class or Interface for objects that support soft delete so we can guarantee these properties exists?

code/web/sys/DB/DataObject.php
```diff
@@ -536,13 +551,20 @@ public function get($columnName, $value = null): bool {
 		return $this->find(true);
 	}
 
-	public function delete($useWhere = false) : int {
+	public function delete($useWhere = false, $hardDelete = false) : int {
```
mdnoble73, line 554: Shouldn't hardDelete default to true to preserve previous behavior? Or is the intention to always have the objects soft delete if possible unless explicitly hard deleted?

code/web/sys/DBMaintenance/version_updates/25.08.00.php
```diff
@@ -99,6 +99,52 @@ function getUpdates25_08_00(): array {
 		//Yanjun Li - ByWater
 
 		// Leo Stoyanov - BWS
+		'object_restoration_permission' => [
+			'title' => 'Add Object Restoration Permission',
+			'description' => 'Add new permission to allow administrators to restore soft-deleted objects.',
+			'continueOnError' => true,
+			'sql' => [
+				"INSERT INTO permissions (sectionName, name, requiredModule, weight, description) VALUES ('System Administration', 'Administer Object Restoration', '', 13, 'Allows the user to view and restore soft-deleted objects (e.g., User Lists) within Aspen.')",
```
mdnoble73, line 123: We should enable this for opacAdmin by default

code/web/services/Admin/ObjectEditor.php
```diff
@@ -1393,6 +1407,15 @@ protected function showHistoryLinks() {
 		return true;
 	}
 
+	/**
+	 * Control whether edit buttons should be shown in history and compare views.
+	 * Purpose: Override it to hide edit functionality when appropriate.
+	 * @return bool True if edit buttons are enabled, false otherwise.
+	 */
+	protected function showEditButtons(): bool {
```
mdnoble73: This should probably be named more explicitly. To me it implies that edit buttons shouldn't be shown anywhere.

code/web/services/Admin/ObjectRestorations.php
```diff
@@ -0,0 +1,520 @@
+<?php
+
+use JetBrains\PhpStorm\NoReturn;
```
mdnoble73: We should talk about if we want to use these attributes. I like the clarity of them, but I know several people don't use PhpStorm.

code/web/services/Admin/ObjectRestorations.php
```diff
@@ -0,0 +1,520 @@
...
+			if (!empty($selectedObjects)) {
+				$id = $selectedObjects[0];
+				if (str_contains($id, '_')) {
+					[$objectType, $actualId] = explode('_', $id, 2);
+					if (class_exists($objectType)) {
+						$managedClasses = self::$managedClasses;
+						if (isset($managedClasses[$objectType])) {
+							require_once ROOT_DIR . self::$managedClasses[$objectType]['classFile'];
+							$tempObject = new $objectType();
+							$structure = $tempObject->getObjectStructure($context);
```
mdnoble73, line 96: Can we/should we use the context to add these extra properties rather than hardcoding them here?

code/web/services/Admin/ObjectRestorations.php
```diff
@@ -0,0 +1,520 @@
...
+			$sample->selectAdd();
+			$sample->selectAdd($sample->getPrimaryKey());
+			$sample->selectAdd($info['titleColumn']);
+			$sample->selectAdd('dateDeleted');
+			$sample->selectAdd('deletedBy');
+			if ($class === 'UserList') {
+				$sample->selectAdd('user_id');
+				$sample->selectAdd('deleteFromIndex');
+			}
+			if (!method_exists($sample, 'supportsSoftDelete') || !$sample->supportsSoftDelete()) continue;
```
mdnoble73: We know this method will always exist right?

code/web/services/Admin/ObjectRestorations.php
```diff
@@ -0,0 +1,520 @@
...
+				// Exclude lists marked for deletion from index.
+				$sample->whereAdd('(deleteFromIndex IS NULL OR deleteFromIndex = 0)');
+			}
+			$sample->find();
+			while ($sample->fetch()) {
+				$rows[$class . '_' . $sample->getPrimaryKeyValue()] = $this->createRestorationItem($sample, $class, $info['titleColumn']);
+			}
+		}
+
+		return $rows;
```
mdnoble73: This could potentially be a large number of rows? Do we need any form of filtering or pagination before loading the entire list? Could we run into memory or time constraints?

code/web/services/Admin/ObjectRestorations.php
```diff
@@ -0,0 +1,520 @@
...
+				$item->daysRemaining = 'Final Day!';
+			}
+		}
+
+		if (!empty($orig->deletedBy)) {
+			require_once ROOT_DIR . '/sys/Account/User.php';
+			$deletingUser = new User();
+			$deletingUser->id = $orig->deletedBy;
+			if ($deletingUser->find(true)) {
+				$item->deletedBy = trim($deletingUser->getBarcode() . ' - ' . $deletingUser->getDisplayName(), ' -');
```
mdnoble73: Users i.e. SSO users and Local admins may not have barcodes. Do we want additional information here in those cases?

code/web/services/Admin/ObjectRestorations.php
```diff
@@ -0,0 +1,520 @@
...
+	 * @noinspection PhpUnused
+	 */
+	#[NoReturn] public function hardDeleteSingle(): void {
+		$compositeId = $_REQUEST['id'] ?? '';
+		$user = UserAccount::getActiveUserObj();
+		if (str_contains($compositeId, '_')) {
+			[$class, $id] = explode('_', $compositeId, 2);
+			if (class_exists($class)) {
+				$deleteObj = new $class();
+				$pk = $deleteObj->getPrimaryKey();
```
mdnoble73, line 365: Should this be getPrimaryKeyColumn? Either getPrimaryKeyColumn or getPrimaryKey should work, but could be good to be explicit that it is the column?

code/web/services/Admin/ObjectRestorations.php
```diff
@@ -0,0 +1,520 @@
...
+	 * Restore a single object from the "recycle-bin" back to an active state.
+	 */
+	#[NoReturn] public function restore(): void {
+		$compositeId = $_REQUEST['id'] ?? '';
+		$user = UserAccount::getActiveUserObj();
+		if (str_contains($compositeId, '_')) {
+			[$class, $id] = explode('_', $compositeId, 2);
+			if (class_exists($class)) {
+				$obj = new $class();
+				$pk = $obj->getPrimaryKey();
```
mdnoble73, line 330: Should this be getPrimaryKeyColumn? Either getPrimaryKeyColumn or getPrimaryKey should work, but could be good to be explicit that it is the column?

mdnoble73 (changes requested):

code/web/sys/DB/DataObject.php
ghost: Leftover from an earlier iteration, so all have been removed.

code/web/sys/DB/DataObject.php
ghost: The current approach using `supportsSoftDelete()` method and `_includeDeleted` property is a simple pattern for `DataObject` subclasses, so adding an interface would introduce additional complexity without meaningful benefits.

code/web/sys/DB/DataObject.php
ghost: This is intentional because:
- If `$hardDelete = false` (default) AND object supports soft delete: Perform soft delete.
- If `$hardDelete = true` OR object doesn't support soft delete: Perform hard delete.

This means objects that don't support soft delete will always be hard deleted, and objects that support soft delete will be soft deleted by default.

code/web/sys/DBMaintenance/version_updates/25.08.00.php
ghost: Added DB update.

code/web/services/Admin/ObjectEditor.php
ghost: Renamed method signature to `showEditButtonsInCompareAndHistoryViews()` for clarity.

code/web/services/Admin/ObjectRestorations.php
ghost: Removed them in the meantime.

code/web/services/Admin/ObjectRestorations.php
ghost: If I understand your question, using the context would require editing the object structure of all of the subclasses that support soft deletion just to unnecessarily add the same two property definitions in each subclass.

code/web/services/Admin/ObjectRestorations.php
ghost: Added SQL pagination.

code/web/services/Admin/ObjectRestorations.php
ghost: Refactor: If barcode is empty, try to use username, and fall back to just display name if both are empty.

code/web/services/Admin/ObjectRestorations.php
ghost: Here, I am working with a DataObject instance and need the actual property name to set the primary key value, whereas `getPrimaryKeyColumn()` is a method for ObjectEditor instances.

ghost:
> It looks like this handles not indexing soft deleted lists, but it does not appear to address soft deleted web builder content. Is that addressed elsewhere?

Added consideration for soft-deleted Web Builder objects during indexing.

code/web/sys/DB/DataObject.php
mdnoble73: Are there cases where this could have unintended consequences where we are expecting something to be deleted, but it isn't because of the new behavior? Have all cases of delete with the objects that are now able to be soft deleted been audited?

code/web/sys/DB/DataObject.php
mdnoble73: That works for me, thanks

code/web/sys/LocalEnrichment/Placard.php
mdnoble73, line 1: When soft deleting a placard do we also need to soft delete the triggers so they can be restored as well or will the administrators need to rebuild the triggers after restoring the data? 

From what I can tell the triggers and links to libraries, locations, and languages aren't removed until the Placard is hard deleted. If that is the case, the placard may still be generated while soft deleted since I also didn't see logic to not display soft deleted placards in the results.

mdnoble73: We'll need to think about this logic for all things that can be soft deleted.

code/web/sys/DB/DataObject.php
mdnoble73: Since the _includeDeleted property is defined in this class, we don't need the property_exists check.

code/web/sys/WebBuilder/BasicPage.php
mdnoble73, line 1: When a page is restored, do we need to update the date updated so that it gets reindexed? This will apply to all indexed content that can be soft deleted.

code/web/sys/DB/DataObject.php
ghost: `supportsSoftDelete()` is false by default for all DataObjects; this means that there is no way for an object to be soft deleted that does not explicitly have the overridden version of the method that returns true, unless a developer adds that by mistake somehow.

mdnoble73: I was thinking more of if a piece of code triggers a list deletion and expects it to be fully deleted at that time, but it is soft deleted instead. We could wind up with more things being soft deleted than are intended to be deleted by administrators.

ghost: The general design is that these objects that explicitly support soft deletion via that method are always soft deleted, unless declared otherwise. And by declared otherwise, I mean by simply passing in `true` for `$hardDelete`. Currently, the way the objects that support soft deletion are built, there cannot be these situations where hard deletion _must_ happen; the intention is for this to apply to objects that provide "enrichment" and are uncritical to functionality. Considering the privacy of users, patrons can decide to hard delete user lists by selecting the checkbox in the modal.

code/web/sys/LocalEnrichment/Placard.php
ghost: Triggers and links to objects are not removed until hard deletion. The soft-deleted placards do not generate in search results because of the guard in the DataObject `find()` method that prevents querying of soft-deleted objects.

code/web/sys/WebBuilder/BasicPage.php
ghost: The `lastUpdate` column is not necessary for indexing other than for the fetch frequency of the pages. If set to fetch monthly, it will correspond to when it was actually last updated, not to when it was restored, so it will be reindexed a month after that original date (which won't matter if it is soft-deleted at that moment). I don't see this having any bad consequences. In terms of the concern of stale pages in the index, this is not possible because the index is wiped every time the process runs.

code/web/services/Admin/ObjectRestorations.php
mdnoble73: I'm ok with that. Definitely less duplication of code.

code/web/sys/WebBuilder/BasicPage.php
mdnoble73: Ope yep, it is indexing all of them every time. I was thinking we only did changed pages.

mdnoble73 (approved):

PR #2842, opened 2025-08-04
DIS-1188: Fixed Adding Administrators for SSO Accounts

LeoStoyanovByWater:
- Fixed an issue where new Administrators could not be added for SSO accounts.

### DIFF
--- a/code/web/services/Admin/Administrators.php
+++ b/code/web/services/Admin/Administrators.php
@@ -107,11 +107,27 @@ class Admin_Administrators extends ObjectEditor {
 					if($numResults == 0) {
 						$newAdmin = UserAccount::findNewUser('', $login);
 						if ($newAdmin === false) {
-							$newAdmin = false;
-							$errors[$login] = translate([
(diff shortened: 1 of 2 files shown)


code/web/services/Admin/Administrators.php
```diff
@@ -107,11 +107,27 @@ function processNewAdministrator() : void {
 					if($numResults == 0) {
 						$newAdmin = UserAccount::findNewUser('', $login);
 						if ($newAdmin === false) {
-							$newAdmin = false;
-							$errors[$login] = translate([
-								'text' => 'Could not find a user with that barcode or username.',
-								'isAdminFacing' => true,
-							]);
+							// Try searching by username field for admin_sso users.
+							$newAdmin = new User();
```
mdnoble73, line 111: @LeoStoyanovByWater this should check the source to be sure that we don't access a user from a different account profile

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2891, opened 2025-08-15
DIS-1217: Fixed Multi-Level Subdomain Detection for Library/Location Routing

LeoStoyanovByWater:
- Fixed subdomain detection to properly handle multi-level subdomains by extracting all possible subdomain combinations for more flexible library/location matching.

Test Plan:
1. Add a second-level subdomain to the Apache config.
2. Add the subdomain to the respective Locations setting (e.g., discover.kids).
3. Notice that when you travel to that site, the domain resolves to the original domain of the site, likely the consortium catalog.
4. Apply the patch and reload the page to see it resolve to the desired domain.

### DIFF
--- a/code/web/release_notes/25.09.00.MD
+++ b/code/web/release_notes/25.09.00.MD
@@ -13,6 +13,8 @@
 // imani
 
 // leo
+### Server Setup Updates
+- Fixed subdomain detection to properly handle multi-level subdomains by extracting all possible subdomain combinations for more flexible library/location matching. (DIS-1217) (*LS*)
 
 // yanjun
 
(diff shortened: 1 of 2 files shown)


NashvillePublicLibrary:
@LeoStoyanovByWater I hadn't even considered second-level subdomains as a possibility! I think I understand that the Aspen Location should look like `discover.kids` but I don't understand what the apache config should look like. Can you please offer a brief example?

ghost:
@NashvillePublicLibrary In your aspen site's `httpd-sitename.production.conf`, you would add an Apache mod_rewrite rule, such as the following: `RewriteCond %{SERVER_NAME} =discovery.kids.library.org [OR]`. This follows the standard way like all other subdomains.

PR #2908, opened 2025-08-19
DIS-1050 sorting date on order

lathomas64:
to test:
do a search and order by date purchased and filter to on order items. The items should be properly sorted by their date purchased now.


code/reindexer/src/org/aspen_discovery/reindexer/GroupedWorkSolr2.java
```diff
@@ -671,9 +671,37 @@ private Long loadScopedDaysAdded(ItemInfo curItem) {
...
+						daysSinceAdded = Long.MAX_VALUE;
+					}
+				} else {
+					daysSinceAdded = DateUtils.getDaysSinceAddedForDate(dateAdded);
+				}
+				//in order to make this appear before anything else we are going to shift it by -999
+				daysSinceAdded += -999L;
+				//clamping to -1 in case we get a value > 998 
+				//worst case scenario we are getting the previous behavior. 
+				daysSinceAdded = Math.clamp(daysSinceAdded, -999L, -1L);
```
mdnoble73: @lathomas64 Math.clamp is not supported in the minimum version of Java that we use. We currently need to support back to Java 11 for a few older systems. Can you update to use different functions.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2909, opened 2025-08-19
DIS-1225: Optimized `getHistoryPropertyName` Method

LeoStoyanovByWater:
- Optimized `getHistoryPropertyName` method to reduce memory consumption and decrease save times.
- Suppressed errors "Error uploading file" errors just because no file was uploaded to reduce `messages.log` clutter.

Test Plan: Ensure admin pages still save and their Object Histories are updated. In particular, saving Themes are expensive for creating Object History entries, so save before and after the patch to see the time difference.


code/web/sys/DataObjectUtil.php
```diff
@@ -457,8 +457,9 @@ static function processProperty(DataObject $object, $property, $fieldLocks) : vo
 				$object->setProperty($propertyName, $_REQUEST["{$propertyName}_existing"], $property);
 			} elseif (isset($_FILES[$propertyName])) {
 				if ($_FILES[$propertyName]["error"] > 0) {
-					//return an error to the browser
-					$logger->log("Error uploading file " . $_FILES[$propertyName]["error"], Logger::LOG_ERROR);
+					if ($_FILES[$propertyName]["error"] != UPLOAD_ERR_NO_FILE) {
```
mdnoble73: This should be split out into another pull request since it doesn't have anything to do with DIS-1225

mdnoble73 (approved):

PR #2917, opened 2025-08-20
DIS-1229: Add API Documentation to Aspen Administration

myr-onl:


myr-onl:
@mdnoble73 It looks like the spaces v. tabs failure is because of the `rapidoc-min.js` file that I added in and not one of the other files... I could swap back to CDN for it, but I figured it'd be good to have the file locally.

PR #2932, opened 2025-08-26
DIS-669: Sync Opting In/Out of Reading History Between Evergreen & Aspen

LeoStoyanovByWater:
- Under Library System settings, the "Opting In to Reading History Updates ILS Settings" and "Opting Out of Reading History Updates ILS Settings" now sync with the Evergreen to update patrons' reading history preferences.
  - If both settings are enabled for the patron's home library, the patron is automatically opted in/out to Reading History in Aspen depending on the patron's Reading History preferences in Evergreen.
- Added success message to display when patrons stop and start recording their reading history.

Tested on CW Mars test server.

Test Plan:
1. Ensure both "Opting In to Reading History Updates ILS Settings" and "Opting Out of Reading History Updates ILS Settings" are enabled for the test patron account's home Library System.
2. Log in as the patron.
3. Start and stop recording reading history as the patron. In the Evergreen OPAC >> My Account >>> Preferences >>> Search & History, notice that the "Keep history of checked out items?" setting updates accordingly.
4. If you change that setting in the Evergreen OPAC, log out and back in as the patron, the patron's reading history preferences about whether to record it update.


code/web/services/MyAccount/ReadingHistory.php
```diff
@@ -59,7 +59,11 @@ function launch() : void {
 				//Perform the requested action
 				$selectedTitles = $_REQUEST['selected'] ?? [];
 				$readingHistoryAction = $_REQUEST['readingHistoryAction'];
-				$patron->doReadingHistoryAction($readingHistoryAction, $selectedTitles);
+				$result = $patron->doReadingHistoryAction($readingHistoryAction, $selectedTitles);
+				if (isset($result['message'])) {
```
mdnoble73, line 63: Rather than storing this in the session, let's use the user->updateMessage which is the newer technique for showing update messages

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #2933, opened 2025-08-26
DIS-1225: Added Sorting by Frozen Holds

LeoStoyanovByWater:
- Add the option to sort pending, frozen holds by "Reactivation Date".

Test Plan:
1. Log into a patron account.
2. Freeze a couple of Pending Holds.
3. Sort them by "Reactivation Date". Notice that frozen holds that will be reactivated with the earliest reactivation dates are at the top.


code/web/services/MyAccount/AJAX.php
```diff
@@ -4084,6 +4084,7 @@ public function getHolds(): array {
 				if ($library->showHoldCancelDate) {
 					$unavailableHoldSortOptions['cancelDate'] = 'Hold Cancellation Date';
 				}
+				$unavailableHoldSortOptions['reactivate'] = 'Reactivation Date';
```
mdnoble73: Not all ILSs and eContent support reactivation dates (Sierra, Evolve, maybe others). We should check to see if reactivation dates are supported prior to displaying this option.

mdnoble73 (changes requested):
Not all ILSs and eContent support reactivation dates (Sierra, Evolve, maybe others). We should check to see if reactivation dates are supported prior to displaying this option.

This option should also be added to LiDA and the APIs if it has not been already.

PR #2947, opened 2025-08-29
DIS-1227: Check the maximum age for self registration

lucasmontoya13:
Now Aspen does honor to the Koha system preference 'PatronSelfRegistrationAgeRestriction' when submitting a new self registration form.

Before the patch: 

1 - Enable the system preference 'PatronSelfRegistrationAgeRestriction' in Koha
2 - Enable the ILS Based Self Registration
3 - Submit a self registration form with an age bigger than the set in 'PatronSelfRegistrationAgeRestriction'
4 - Aspen approves the form

After the patch:

- Repeat steps 1,2,3
- Aspen throws a message about the invalidation of the entered date of birth and submit nothing.

### DIFF
--- a/code/web/sys/Utils/SystemUtils.php
+++ b/code/web/sys/Utils/SystemUtils.php
@@ -54,7 +54,7 @@ class SystemUtils {
 		}
 	}
 
-	static function validateAge($minimumAge, $dob): bool {
+	static function validateAge($minimumAge, $dob, $maximumAge = null): bool {
 		$today = date("d-m-Y");//today's date
 
 		$dob = new DateTime($dob);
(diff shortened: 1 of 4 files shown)


code/web/sys/Utils/SystemUtils.php
```diff
@@ -65,6 +65,9 @@ static function validateAge($minimumAge, $dob): bool {
 		$age = $interval->y;
 
 		if ($age >= $minimumAge) {
+			if(isset($maximumAge) && $maximumAge < $age){
```
mdnoble73: In Koha it looks like the age if a maximum age of 100 is entered that an age of 100 is allowable. So $maximumAge should be <= $age rather than <

mdnoble73 (changes requested):

code/web/sys/Utils/SystemUtils.php
lucasmontoya13: Done!

mdnoble73 (approved):

PR #2961, opened 2025-09-01
DIS-1282: Add print options dialog for lists

catsoup11789:
- added print interface option that uses an alternate layout template to reduce header/footer elements
- update print options button in My List to open a modal to prompt the user what elements they'd like to print: library name, library logo, list author, list description, list entry covers/series/formats/descriptions/notes/holdings/rating
- when using the print interface, always hide: Show Editions button, availability info, waitlist info, toolbars

### DIFF
--- a/code/web/Action.php
+++ b/code/web/Action.php
@@ -35,7 +35,7 @@ abstract class Action
 		$minimalInterface = $_REQUEST['minimalInterface'] ?? false;
 		$interface->assign('minimalInterface', $minimalInterface);
 
-		$printInterface = isset($_REQUEST['print']) ? (bool)$_REQUEST['print'] : false;
+		$printInterface = isset($_REQUEST['print']) ? filter_var($_REQUEST['print'], FILTER_VALIDATE_BOOLEAN) : false;
 		$interface->assign('printInterface', $printInterface);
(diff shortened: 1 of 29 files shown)


code/web/Action.php
```diff
@@ -35,6 +35,13 @@ function display($mainContentTemplate, $pageTitle, $sidebarTemplate = 'Search/ho
 		$minimalInterface = $_REQUEST['minimalInterface'] ?? false;
 		$interface->assign('minimalInterface', $minimalInterface);
 
+		$printInterface = isset($_REQUEST['print']) ? (bool)$_REQUEST['print'] : false;
```
mdnoble73: Rather than using the cast to bool here, we need to use filter_var($_REQUEST['print'], ]FILTER_VALIDATE_BOOLEAN) to handle a value of "false".

code/web/interface/themes/responsive/GroupedWork/multipleVariationManifestion.tpl
mdnoble73, line 1: We'll probably need to reconcile this with the new format horizontal display style?

code/web/interface/themes/responsive/GroupedWork/singleVariationManifestion.tpl
mdnoble73, line 1: We'll probably need to reconcile this with the new format horizontal display style?

PR #2981, opened 2025-09-05
DIS-1152: Get Unsuppressed Volume Data for All Variations

LeoStoyanovByWater:
- Fixed an issue where only volumes for the related record would display as options for holds; all variations are now considered.


mdnoble73 (changes requested):
@LeoStoyanovByWater it looks like this should be a new Jira rather than re-using DIS-1152. Can you explain the issue you are trying to correct in the new Jira?

PR #2987, opened 2025-09-06
DIS-1301: Fixed Search URL for Accessible Browse Categories

LeoStoyanovByWater:
- With "Use Accessible Layout" enabled for a Theme, browse categories on the home page now generate proper search URLs instead of placeholder links.

Test Plan:
1. Enable Accessible Layout in the Themes settings for a catalog.
2. Navigate to the home page to view browse categories.
3. Click on the main browse category search link that has subcategories. Notice that it does not navigate you anywhere.
4. Apply the patch, reload the page, and click the same search link. Notice that it navigates you to the appropriate results.

### DIFF
--- a/code/web/services/Search/Home.php
+++ b/code/web/services/Search/Home.php
@@ -214,10 +214,15 @@ class Search_Home extends Action {
 				if ($browseCategory->isValidForDisplay()) {
 					$textId = $browseCategory->textId;
 					$subCatResult = $searchAPI->getSubCategories($textId);
+					require_once ROOT_DIR . '/services/Browse/AJAX.php';
+					$browseAJAX = new Browse_AJAX();
+					$browseAJAX->textId = $textId;
+					$categoryResults = $browseAJAX->getBrowseCategoryResults();
+					$searchUrl = $categoryResults['searchUrl'] ?? '#';
 					$browseCategories[] = [
 						'textId' => $textId,
 						'label' => $browseCategory->label,
-						'searchUrl' => '#',
+						'searchUrl' => $searchUrl,
 						'hasSubcategories' => !empty($subCatResult['subCategories']),
 					];
 				}
--- a/code/web/services/Search/Home.php
+++ b/code/web/services/Search/Home.php
@@ -214,16 +214,17 @@ class Search_Home extends Action {
 				if ($browseCategory->isValidForDisplay()) {
 					$textId = $browseCategory->textId;
 					$subCatResult = $searchAPI->getSubCategories($textId);
-					require_once ROOT_DIR . '/services/Browse/AJAX.php';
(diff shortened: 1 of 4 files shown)


code/web/services/Search/Home.php
```diff
@@ -214,10 +214,15 @@ private function getInitialBrowseCategoryFeed(array $localBrowseCategories = nul
...
+					require_once ROOT_DIR . '/services/Browse/AJAX.php';
+					$browseAJAX = new Browse_AJAX();
+					$browseAJAX->textId = $textId;
+					$categoryResults = $browseAJAX->getBrowseCategoryResults();
+					$searchUrl = $categoryResults['searchUrl'] ?? '#';
 					$browseCategories[] = [
 						'textId' => $textId,
 						'label' => $browseCategory->label,
-						'searchUrl' => '#',
+						'searchUrl' => $searchUrl,
```
mdnoble73, line 226: @LeoStoyanovByWater I think this change needs to be reverted or we would need to do additional validation here.

ghost: Why would this need to be changed if basic validation is done in the template to prevent hyperlinking the label?

mdnoble73: Browse_AJAX:getBrowseCategoryResults isn't doing validation that the searchUrl is valid. For example, it still creates a url for list based browse categories even if the list isn't valid. With the regular view the parent browse category is assumed to not be valid so we can use that same convention here.

code/web/services/Search/Home.php
```diff
@@ -214,10 +214,15 @@ private function getInitialBrowseCategoryFeed(array $localBrowseCategories = nul
 				if ($browseCategory->isValidForDisplay()) {
 					$textId = $browseCategory->textId;
 					$subCatResult = $searchAPI->getSubCategories($textId);
+					require_once ROOT_DIR . '/services/Browse/AJAX.php';
+					$browseAJAX = new Browse_AJAX();
+					$browseAJAX->textId = $textId;
+					$categoryResults = $browseAJAX->getBrowseCategoryResults();
```
mdnoble73: This is an expensive call since we generate the full results with all the records just to get the search url and then everything else gets thrown away.

mdnoble73 (changes requested):

mdnoble73:
Sorry missed completing my review yesterday

code/web/services/Search/Home.php
ghost: We still need a search URL for browse categories that are only one sub-category. The newest changes are visible on Alpha (https://alpha.aspendiscovery.org/), where solo browse categories still have search URLs, while parent categories with subcategories do not.

mdnoble73 (approved):

PR #3046, opened 2025-09-18
DIS-1345: Koha reserve fee message

Chloe070196:
WIP for [DIS-1345](https://aspen-discovery.atlassian.net/browse/DIS-1345)

[DIS-1345]: https://aspen-discovery.atlassian.net/browse/DIS-1345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Note: the method generating the fee messages are separate (pre and post) despite being very similar as, in the future, they will be rewritten to query separate Koha API endpoints. In fact, the post submission may later become redundant and need to be deprecated. This version of the patch matches the current Koha main branch.

BLOCKED BY: Koha Bugzilla Bug 40817

### DIFF
--- a/code/web/Drivers/AbstractIlsDriver.php
+++ b/code/web/Drivers/AbstractIlsDriver.php
@@ -1079,4 +1079,8 @@ abstract class AbstractIlsDriver extends AbstractDriver {
 	public function getPatronHoldGroups($patronId): ?array {
 		return null;
 	}
+
+	public function hasHoldFeeMessage(): bool {
+		return false;
+	}
 }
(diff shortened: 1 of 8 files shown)


mdnoble73:
@Chloe070196 if this is ready for inclusion in 26.05 can you take it out of draft status and let me know.

Chloe070196:
@mdnoble73 unfortunately not - it requires a Koha circulation rule introduced as a recent enhancement that hasn't made it into main yet. I don't think this should get into core aspen without a Koha version check, but we don't yet know what koha version Aspen will need to be checking for. The dependency is why I've kept this as a draft (DIS-1938 is in a similar spot).

mdnoble73 (approved):

PR #3052, opened 2025-09-18
DIS-1161: More User List Sort Options & Format Filtering

LeoStoyanovByWater:
- Added the following sorting options:
  - Publication Date (Oldest First) and Publication Date (Newest First): Uses 4-digit years and entries with invalid years are sorted to the end.
  - Call Number: Sort ILS records by call number then shelf location and place other record types at bottom.
  - Availability (Available First) and Availability (Unavailable First): Sort by total number of available or unavailable copies, respectively.
  - Number of Copies (Most First) and Number of Copies (Least First): Sort by total number of available copies across all items within a Grouped Work.
- Added Format filtering, such that when a format is chosen, only record manifestations corresponding to that format display in the list.
  - The new sort options respect Format filtering, thus only sorting by the formats chosen.

Test Plan:
1. Apply the patch.
2. Navigate to a User List with multiple formats and list entry types (e.g., Events, Grouped Works, etc.).
3. Sort and filter by formats, trying different variations and inspecting the results.


mdnoble73 (changes requested):
Some questions within the Jira
Also noticed that changing format filters do not reset to page 1. The active page may not be valid.

code/web/sys/UserLists/UserList.php
```diff
@@ -235,12 +293,220 @@ function getListEntries($sort = null, $forLiDA = false, $appVersion = 0, $start
...
 					$listEntry->orderBy("CASE WHEN ItemAuthor IS NULL THEN 1 ELSE 0 END ASC, ItemAuthor");
-				} else {
-					$listEntry->orderBy($sortOptions[$sort]);
+					// Publication date sort: Extracts clean 4-digit years from messy publication date strings.
+					// Joins through grouped_work -> grouped_work_records -> indexed_publication_date tables.
+					// Uses regex to find first valid year (0001-2999), filters out obviously invalid entries.
+					// Records without valid years are sorted to the end.
+				} elseif ($sort == "publication_date" || $sort == "publication_date_desc") {
+					require_once ROOT_DIR . '/sys/Grouping/GroupedWork.php';
+					$groupedWorkInfo = new GroupedWork();
```
mdnoble73, line 302: Within Solr we store the publication date for sorting we may need to add that information to the database, but it would provide a single value for sorting. I'd be ok dealing with the performance in a future update if libraries could choose which of the advanced sort options they wanted to enable.

code/web/sys/UserLists/UserList.php
```diff
@@ -235,12 +293,220 @@ function getListEntries($sort = null, $forLiDA = false, $appVersion = 0, $start
...
+					$listEntry->orderBy("CASE WHEN NormalizedYear IS NULL THEN 1 ELSE 0 END ASC, NormalizedYear $order");
+
+					// Call number sort: Sorts ILS records by call number then shelf location; places other record types at bottom.
+					// Joins through grouped_work -> grouped_work_records -> grouped_work_record_items -> indexed_call_number + indexed_shelf_location.
+					// Uses grouped_work_variation.eContentSourceId to distinguish ILS (-1,NULL,0) from e-content (>0).
+					// Sort order: ILS records grouped by shelf location, then by call number within each location -> e-content/non-GroupedWork.
+					// Groups by list entry ID to prevent duplicates when multiple items have different call numbers/locations.
+				} elseif ($sort == "call_number") {
+					require_once ROOT_DIR . '/sys/Grouping/GroupedWork.php';
+					$groupedWorkInfo = new GroupedWork();
```
mdnoble73, line 340: We also have a single sortable call number in solr stored per grouped work, but not necessarily by format. We can deal with the performance and discrepancies in sorting in a future release?

code/web/sys/UserLists/UserList.php
```diff
@@ -235,12 +293,220 @@ function getListEntries($sort = null, $forLiDA = false, $appVersion = 0, $start
...
+
+					require_once ROOT_DIR . '/sys/Grouping/GroupedWorkRecord.php';
+					$gwRecords = new GroupedWorkRecord();
+					$listEntry->joinAdd($gwRecords, "LEFT", 'gwRecords', 'groupedWork.id', 'groupedWorkId');
+
+					require_once ROOT_DIR . '/sys/Grouping/GroupedWorkItem.php';
+					$gwItems = new GroupedWorkItem();
+					$listEntry->joinAdd($gwItems, "LEFT", 'gwItems', 'gwRecords.id', 'groupedWorkRecordId');
+
+					$indexedFormat = new class extends DataObject {
```
mdnoble73: There is already a class for IndexedFormat why redeclare it here?

mdnoble73 (changes requested):
We need to look at the performance and where this data is stored

code/web/sys/UserLists/UserList.php
ghost: Opted to use the class.

mdnoble73 (approved):

PR #3064, opened 2025-09-22
DIS-1226: Split backend service and introduce php-fpm (Docker)

lucasmontoya13:
It is possible to start using php-fpm with Apache instead of the classic mode (mod_php) in Aspen.
There are a lot of benefits about it : 
- Apache only loads PHP when necessary
- Apache processes are lighter for static content
- PHP-FPM can automatically recycle processes instead of creating a new one each time
- Independent tuning of Apache vs PHP process numbers


lucasmontoya13:
Done!

PR #3072, opened 2025-09-23
DIS-1368: Added Lazy Loading of Circulation Data

LeoStoyanovByWater:
### Account Updates
- When "Show Checkouts and Holds in Results" is enabled in Preferences, circulation data is no longer fetched before page load, thus decreasing response times when navigating the catalog. (DIS-1368) (*LS*)
  - An improved 5-minute cache is maintained with circulation data, so the circulation buttons load faster.
  - Circulation buttons (e.g., "Place Hold") automatically refresh when the cache becomes outdated and display a subtle shimmer effect during refreshes.
  - This lazy loading works in dynamically loaded content, including modals and AJAX responses.

### Other Updates
- Fixed miscellaneous PHP warnings related to ShareIt and InnReach. (DIS-1368) (*LS*)

Test Plan:
1. Log into a patron account and enable "Show Checkouts and Holds in Results" in Preferences.
2. Check out and place holds on various titles from various sources.
3. After a set of checkouts and holds, you will notice a shimmer effect on buttons as you navigate the catalog. This indicates that the action buttons may be updated based upon new circulation data.
4. After a couple of seconds, depending on the number of sources from which to fetch, the buttons for the records you checked out/placed a hold on will be updated to reflect your actions.
5. Return some checked out items and cancel some holds. Notice that the action buttons are updating once again while still displaying the cached circulation data.
6. After 5 minutes of no circulation actions and while searching the catalog, the action buttons will shimmer because cached data is being updated in case any circulation action did occur.
   - This is particularly the case for titles checked out from the ILS and outside of Aspen (e.g., Libby app).


code/web/Drivers/PalaceProjectDriver.php
```diff
@@ -542,7 +543,7 @@ function cancelHold($patron, $recordId, $cancelId = null, $isIll = false): array
 			}
 			if (!$cancelWorked) {
 				$result['message'] = translate([
-					'text' => "Could not cancel Palace Project hold, " . (string)$status->statusMessage,
+					'text' => 'Could not cancel Palace Project hold.',
```
mdnoble73, line 546: How do these changes relate to the Jira?

code/web/RecordDrivers/HooplaRecordDriver.php
```diff
@@ -494,7 +512,7 @@ function getActions() {
 				'isPublicFacing' => true,
 			]);
 			$actions[] = [
-				'onclick' => "return AspenDiscovery.Hoopla.getCheckOutPrompts('{$this->id}')",
+				'onclick' => "return AspenDiscovery.Hoopla.getCheckOutPrompts('{$this->id}', 'Instant', this)",
```
mdnoble73: Why is this hard coded to Instant? Shouldn't it use $this->getHooplaType()?

code/web/interface/themes/responsive/css/responsive_base.less
```diff
@@ -238,6 +238,39 @@ body {
   }
 }
 
+.btn[data-needs-refresh="true"], .btn.shimmer-fade-out {
```
mdnoble73: The shimmer should be a theme setting for libraries that prefer to not have the buttons shimmer during load.

code/web/services/CloudLibrary/AJAX.php
```diff
@@ -14,15 +14,23 @@ function placeHold() {
 			} else {
 				return [
 					'result' => false,
+					'title' => translate([
+						'text' => "Error Placing Hold on CloudLibrary Title",
+						'isPublicFacing' => true,
+					]),
 					'message' => translate([
-						'text' => 'Sorry, it looks like you don\'t have permissions to place holds for that user.',
+						'text' => 'Sorry, it looks like you don\'t have permissions to place holds as that user.',
```
mdnoble73: I don't think the terminology change is appropriate here. In this case, we are placing a hold for a particular user. That user will still be using it .

code/web/services/MyAccount/AJAX.php
```diff
@@ -16,6 +16,13 @@ function launch($method = null) : void {
 				break;
 			case 'getUserHolds':
 				$method = 'getUserHolds';
+				break;
+			case 'refreshUserCirculationCache':
```
mdnoble73: There is no need to change the method name to the same value

code/web/sys/Account/User.php
```diff
@@ -2199,6 +2168,213 @@ public function getHolds($includeLinkedUsers = true, $unavailableSort = 'sortTit
...
+	}
+
+	/**
+	 * Save current circulation cache to session storage.
+	 * This allows the UI to continue displaying accurate circulation status immediately
+	 * after checkout/hold actions, while still enabling the lazy-loading system to refresh
+	 * buttons with updated data in the background.
+	 */
+	private function saveCirculationCacheToSession(): void {
+		if (!empty($this->_circulationStatusCache)) {
```
mdnoble73: Why put this in the session table rather than leaving it in the existing cache tables?

code/web/Drivers/PalaceProjectDriver.php
ghost: While working on it all, I noticed IntelliSense detect that this variable is never defined or initialized, so it was a trivial enough code cleanup to include it.
image

code/web/sys/Account/User.php
ghost: When a patron checks out a title or puts a title on hold, `forceReloadOfCheckouts()` and `forceReloadOfHolds()` are invoked, respectively. These clear DB entries, so upon page reload, no circulation buttons would be returned to their original states. The session cache persists this circulation data between requests, so even if there is one outdated button because the title was checked out, for example, all other buttons' circulation displays persist.

code/web/services/MyAccount/AJAX.php
ghost: Reverted

code/web/services/CloudLibrary/AJAX.php
ghost: Reverted

code/web/interface/themes/responsive/css/responsive_base.less
ghost: Added `showButtonShimmer` to Themes to make it optional.

code/web/RecordDrivers/HooplaRecordDriver.php
ghost: Added `$this->getHooplaType()`.

code/web/Drivers/PalaceProjectDriver.php
mdnoble73: Sounds good thanks for the explanation

code/web/interface/themes/responsive/css/responsive_base.less
mdnoble73: thanks!

code/web/sys/Account/User.php
mdnoble73: We could update to not clear the database, but instead set that the data needs to be reloaded. That way the data is still stored efficiently and across all sessions for the user. Then when the holds/checkouts are actually reloaded the old values can be cleared.

mdnoble73 (approved):

PR #3083, opened 2025-09-24
DIS-1160: Added Confirauble

LeoStoyanovByWater:
### Grouped Work Facets Updates
- Added "Num Total Facets" option to set how many facets display when patrons click "More..." to expand facet groups while searching.
    - The Available At, Format, and Branch (i.e., Owning Location) facets continue to be dynamically calculated rather than configurable.
    - Renamed "Num Entries" column label to "Num Default Entries" to clarify its purpose.

#### New Settings
- Catalog/Grouped Works > Grouped Work Facets > "Num Total Entries" column under the Facets table

### Other Updates
- Improved `oneToMany.tpl` to support per-instance readonly field conditions and automatic max/min constraint application for integer fields, allowing individual items to have different readonly states with dynamic tooltips populated through `readOnlyReason` attributes across form element types.

Test Plan:
1. Apply the patch.
2. Navigate to Catalog/Grouped Works > Grouped Work Facets > "Num Total Entries" column under the Facets table.
3. Enter various numbers for each row, where allowed.
4. Search the catalog and browse through the facet groups' settings you modified to confirm total facet counts after clicking "More...".

### DIFF
--- a/code/web/interface/themes/responsive/DataObjectUtil/oneToMany.tpl
+++ b/code/web/interface/themes/responsive/DataObjectUtil/oneToMany.tpl
@@ -20,6 +20,11 @@
 			<tbody>
 			{foreach from=$propValue item=subObject}
 				{assign var=subObjectId value=$subObject->getPrimaryKeyValue()}
+				{* Get instance-specific structure once per object *}
+				{assign var=instanceStructure value=null}
+				{if method_exists($subObject, 'updateStructureForEditingObject')}
+					{assign var=instanceStructure value=$subObject->updateStructureForEditingObject($property.structure)}
+				{/if}
 				<tr id="{$propName}{$subObject->id}" class="{$propName}Row" data-id="{$subObject->id}">
 					<input type="hidden" id="{$propName}Id_{$subObject->id}" name="{$propName}Id[{$subObject->id}]" value="{$subObject->id}"/>
 					{if !empty($property.sortable)}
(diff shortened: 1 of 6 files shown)


code/web/interface/themes/responsive/DataObjectUtil/oneToMany.tpl
```diff
@@ -20,6 +20,11 @@
 			<tbody>
 			{foreach from=$propValue item=subObject}
 				{assign var=subObjectId value=$subObject->getPrimaryKeyValue()}
+				{* Get instance-specific structure once per object *}
+				{assign var=instanceStructure value=null}
+				{if method_exists($subObject, 'updateStructureForEditingObject')}
```
mdnoble73: This should be done outside of the template if possible

code/web/sys/Grouping/GroupedWorkFacet.php
```diff
@@ -180,13 +182,48 @@ static function getObjectStructure(string $context = ''): array {
...
 				'default' => '5',
 			],
+			'numTotalEntriesToShowInMore' => [
+				'property' => 'numTotalEntriesToShowInMore',
+				'type' => 'integer',
+				'label' => 'Num Total Entries',
+				'description' => 'The number of values to show in the displayed after clicking &quot;More...&quot;. This setting is ignored for Format, Available At, and Owning Location facets, which use calculated limits.',
+				'default' => '30',
+				'max' => 100,
+				'min' => 1,
```
mdnoble73: This should be higher than 1 I think. If there is only 1 value there is no point showing the pop and patrons would have to search for every value. In fact the popup won't even show if you have at least 5 values. A minimum of 30 would be more appropriate so we'd get a list that can still be scanned and would fill the popup.

code/web/interface/themes/responsive/DataObjectUtil/oneToMany.tpl
ghost: Logic moved outside of the template.

mdnoble73 (changes requested):
I pulled the latest code and tried to change language to 100 and got errors because I have over 100 locations in my system so the changes failed to save because that number is over the max. When a field is read-only the max/min should be disabled.

ghost:
> I pulled the latest code and tried to change language to 100 and got errors because I have over 100 locations in my system so the changes failed to save because that number is over the max. When a field is read-only the max/min should be disabled.

Added check to prevent validation for read-only text fields.

PR #3096, opened 2025-09-26
DIS-1396: Standardized Image Sizes & Saved URLs to Text Files When "Use Original Cover URLs" Is Enabled

LeoStoyanov:
- When "Use Original Cover URLs" is enabled, standardized cover images dimensions, prevented relatively small images from displaying, and modified how they are stored.

Test Plan:
1. Apply the patch.
2. Enable System Variables > Use Original Cover URLs. Clear your browser’s image cache.
3. Browse the catalog to see cover images have more consistent dimensions, and no images are too small or big relative to other cover images.


code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -2240,17 +2235,27 @@ private function validateCoverUrl(string $url, string $source, bool $forceValida
...
 		$imageInfo = @getimagesizefromstring($image);
 		if ($imageInfo === false || ($imageInfo[0] < 2 && $imageInfo[1] < 2)) {
-			$this->log("Cover URL $url returned an image that is too small; falling back.", Logger::LOG_DEBUG);
+			$this->log("Cover URL $url returned an image that is too small; falling back.");
 			return false;
 		}
 
+		// Enforce minimum dimensions for display quality, except for small size (used in editions/lists).
+		if ($this->size !== 'small') {
+			$minWidth = 100;
```
mdnoble73: These dimensions seem overly restrictive. The small image we display in scrollers (i.e. for more like this and also in this series) is only 125px high so several images that we would happily display could be removed.

ghost: Reverted changes on how original cover URLs are stored.
These restrictions only apply to images that do not have `size` set to small. In More Like This, the medium images fetched have an actual height of 200px, and there must be some CSS that shrinks their height to about 125px. They are not stored at that smaller height. This check prevents images fetched that are too small from being chosen; it does not prevent any other modifications that shrink them.

PR #3149, opened 2025-10-09
DIS-1437: Add toolbar to bottom of list page

catsoup11789:


code/web/interface/themes/responsive/MyAccount/list.tpl
mdnoble73, line 1: Could we pull the toolbar code out into another template to make it easier to update?

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #3160, opened 2025-10-14
DIS-1392: Added Functionality to Update Home Library in Preferences for Symphony

LeoStoyanov:
- Added the ability for patrons to update their Home Library on their Preferences page.

Test Plan:
1. Apply the patch.
2. Log into a patron account and navigate to the Preferences page.
3. Update your Home Library and notice that the change is reflected after saving.


code/web/Drivers/SirsiDynixROA.php
```diff
@@ -2636,24 +2636,65 @@ function updatePatronInfo($patron, $canUpdateContactInfo, $fromMasquerade): arra
...
+			'success' => false,
+			'messages' => [],
+		];
 
-			default:
-				return $code;
+		$sessionToken = $this->getStaffSessionToken();
+		if ($sessionToken) {
+			$webServiceURL = $this->getWebServiceURL();
+			if ($userID = $patron->unique_ils_id) {
```
mdnoble73: Should this be a double equal or are you checking to see if the unique_ils_id is empty?

mdnoble73 (changes requested):

code/web/Drivers/SirsiDynixROA.php
LeoStoyanov: It's an assignment and check for truthiness. I changed it to check emptiness for clarity and robustness.

mdnoble73 (approved):

PR #3164, opened 2025-10-15
DIS-1438 community engagement milestone sync

lathomas64:
to test:
without the code
1. make a campaign with a checkout milestone
2. enroll a user in the campaign
3. checkout out an item to the user matching the milestone
4. wait for indexing to sync the item with Aspen 
5. Do not visit the profile page
6. Check the item back in
7. wait for indexing again
8. check the user's campaign milestone progress and the milestone should not have updated

with the code
1. repeat steps 1-4 above
2. manually run updateCommunityEngagementMilestone.php on the server
3. repeat steps 5-7 above.
4. the campaign milestone should have progressed this time.

the same should hold true for hold milestones as well. 

The cron should do nothing if community engagement is not turned on, or there are no users enrolled in active campaigns.


lathomas64:
Accidentally have a bit from the 25.10 release notes in here but I assume that will be making it into 25.11 eventually anyways.

mdnoble73 (changes requested):
This should also go in the default crontab_settings.txt files so new installs will get the updates.

mdnoble73 (approved):

PR #3198, opened 2025-10-21
DIS-1440: Added Library-Specific URL Support for Web Resources

LeoStoyanov:
### Web Resources Updates
- Added support for library-specific URLs in Web Resources, enabling different libraries to link to distinct versions of the same resource.
    - When patrons access a Web Resource, Aspen automatically uses the URL configured for their active library, falling back to the base resource URL if no library-specific one is set.
- All display pages have been updated to respect library-specific URLs, including: individual Web Resource pages, A–Z Resource lists, Search results, and Portal Cells.
- Fixed bug where staff without full permissions would accidentally clear library selections and "Allow Access to Patrons of These Home Libraries" selections when saving a Web Resource.
- Added accessibility improvements with proper aria-labels that conditionally indicate when resources open in new tabs, and displays the external link icon only when the "Open In New Tab" setting is  enabled. 
  - Fixed issue where no browser preview link would display when hovering over an "Open Resource" button.

### Other Updates
- Introduced a new framework component: `checkboxWithOptions` for multi-select fields.
  - Enables administrators to select multiple contexts via checkboxes with optional configuration for each selected context (e.g., a custom URL).

Test Plan:
1. Apply the patch.
2. Navigate to a Web Resource or create a new one.
3. Modify an applied library or a new one to input its “Library-Specific URL”.
4. Navigate to that library catalog where the Web Resource lives.
5. Click “Open Resource” and notice that it navigates to the custom URL rather than the base URL defined in the Web Resource setting.


code/web/sys/WebBuilder/WebResource.php
```diff
@@ -186,10 +188,19 @@ static function getObjectStructure(string $context = ''): array {
 			'libraries' => [
 				'property' => 'libraries',
 				'type' => 'multiSelect',
-				'listStyle' => 'checkboxSimple',
+				'listStyle' => 'checkboxWithOptions',
```
mdnoble73, line 191: Why create a new admin type here rather than using the the typical oneToMany structure?

mdnoble73 (changes requested):

code/web/sys/WebBuilder/WebResource.php
LeoStoyanov: The typical oneToMany structure is cumbersome here to add many individual libraries for large consortia. It is more intuitive to keep the checkbox selection structure that is typical for library selection, but then expanding its functionality to allow for library-specific behavior here (and for future cases).

PR #3239, opened 2025-10-30
DIS-1518: Ensured Display of All Indexed Series Volumes for Record Details

LeoStoyanov:
- Fixed an issue where only one series volume was displayed when MARC records contain multiple 800 fields with different volumes.

Test Plan:
1. Find a MARC record that has multiple 800 fields with different volume numbers. 
2. Navigate to the full record page and view the Series section. Note how many series volumes are displayed.
3. Perform a search that returns the same record in search results. View the Series information shown in the search result listing. Note whether all volumes are displayed.
4. Apply the patch.
5. Repeat step 1. Verify that all series volumes from the MARC 800 fields are now displayed, with the first volume shown as the primary series and additional volumes listed below it.
6. Repeat step 2. Verify that all series volumes are displayed in the search results listing.


code/web/RecordDrivers/GroupedWorkDriver.php
```diff
@@ -2402,13 +2402,38 @@ public function getSeries($allowReload = true, ?int $seriesId = null): ?array {
 				} else {
 					$seriesFromIndex = $this->getIndexedSeries();
 					if ($seriesFromIndex != null && count($seriesFromIndex) > 0) {
+						// Sort series entries by volume for consistent display order.
+						usort($seriesFromIndex, function($a, $b) {
+							$volA = $a['volume'] ?? '';
```
mdnoble73, line 2428: We likely will want anything without a volume to sort after series with a volume and likely should sort by series name prior to sorting by volume since the series name displays before the volume. Unless we can determine which series is used most (which I don't think we can just from the data in the index). If we can determine which series is used most that should display first. The goal would be to have the actual series displayed before things like Thorndike Large Print, Penguin Classics, etc.

mdnoble73 (changes requested):

code/web/RecordDrivers/GroupedWorkDriver.php
LeoStoyanov: I've added your recommendations to the sorting logic.

mdnoble73 (approved):

PR #3246, opened 2025-10-31
DIS-1522 label templates

lathomas64:
No functional changes here, just adding comments to templates to identify which files rendered which markup while inspecting the markup on any given page. 

To test:
Go to any url in aspen with ?debug=true appended to the url (or &debug=true if you already have query parameters in the url)
You should see comments in the code indicating the start and end of templates including files and paths.

### DIFF
--- a/code/web/release_notes/25.12.00.MD
+++ b/code/web/release_notes/25.12.00.MD
@@ -1,0 +1,44 @@
+## Aspen Discovery Updates
+// mark
+
+// kirstien
+
+// kodi
+
+// leo
+
+// imani
+
+// yanjun
+
+//alexander
+
+//chloe
+
+//lucas
+
+## This release includes code contributions from
+### ByWater Solutions
+- Leo Stoyanov (LS)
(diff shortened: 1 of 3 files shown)


mdnoble73 (changes requested):
Rather than changing every single template we should be able to update the smarty processer to output this data

lathomas64:
> Rather than changing every single template we should be able to update the smarty processer to output this data

Any suggestions on how to go about doing so? I see pre filters which would apply only to the top level here: https://www.smarty.net/docs/en/advanced.features.prefilters.tpl
And do not seem to actually process the template functions when placed here but nothing to apply to each include. 

Modifying smarty_internal_compile_include.php seems very deep level and doesn't seem to actually apply any changes either.

lathomas64:
ah nevermind the compiled templates were cached. i can modify things in the latter place perhaps.

mdnoble73:
You can modify the fetch method in /sys/Interface.php. You may need to override display as well

PR #3249, opened 2025-10-31
DIS-1525: Check CircControl for renewal remaining

LiYanjun19:
Koha has a system preference CircControl, to decide 

Use the calendar and circulation rules of for 
1) the library you are logged in at
2) the library the patron is from
3) the library the item is from

When Aspen is checking the circulation rules for renewals in getCheckouts, getHolds, renewCheckouts, we should check the CircControl to know which library’s circulation rule should be used. 

To Test:
1. Check out books on Aspen or Koha. 
2. Go to Koha system preference > CircControl, select different options and save
3. Go to Patrons > check out, see the renew column information
4. Log in to Aspen, view Checked Out Titles > Renewed and Renewals Remaining, see it matches the koha renew column.


code/web/Drivers/Koha.php
```diff
@@ -504,6 +504,16 @@ public function getCheckouts(User $patron): array {
 		require_once ROOT_DIR . '/sys/Indexing/IlsRecord.php';
 		IlsRecord::preloadIlsRecords($this->getIndexingProfile()->name, $allBibNumbers);
 
+		/** @noinspection SqlResolve */
+		$circControlSql = "SELECT value FROM systempreferences WHERE variable = 'CircControl'";
```
mdnoble73: This should use the getKohaSystemPreference function to get the System preference.

code/web/Drivers/Koha.php
```diff
@@ -2337,6 +2353,17 @@ public function getHolds($patron, $page = 1, $recordsPerPage = -1, $sortOption =
...
```
mdnoble73: This should use the getKohaSystemPreference function to get the System preference.

code/web/Drivers/Koha.php
```diff
@@ -2970,21 +3003,38 @@ public function renewCheckout($patron, $recordId, $itemId = null, $itemIndex = n
 			$renewResponse = $this->getXMLWebServiceResponse($renewURL);
 			ExternalRequestLogEntry::logRequest('koha.renewCheckout', 'GET', $renewURL, $this->curlWrapper->getHeaders(), '', $this->curlWrapper->getResponseCode(), $renewResponse, []);
 
+			/** @noinspection SqlResolve */
+			$circControlSql = "SELECT value FROM systempreferences WHERE variable = 'CircControl'";
```
mdnoble73: This should use the getKohaSystemPreference function to get the System preference.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #3254, opened 2025-11-02
DIS-1531: Add manual clear cached values option

LiYanjun19:
Add a “Clear Cached Values” checkbox in System Variables so Admins can manually purge the cached_values table. Admins can see a success or error message after they check the box and save the page. 

This idea came from when an Evergreen server was updated, Aspen needed to clear the cached values, especially for IDL entries. Giving library admins a checkbox on the interface lets them handle the cleanup without waiting for support companies to respond. And the same control can potentially help with other cases when cached values need a quick purge. 

To test:
1. Log in as an admin with access to System Variables
2. Check DB and see cached_values table is not empty 
3. Go to System Variables > Check Clear Cached Values
4. Click “Save Changes and Stay Here” or “Save Changes and Return”, and see the success or error message 
5. Check the cached_values table again and confirm whether entries are cleaned or not

### DIFF
--- a/code/web/release_notes/25.11.00.MD
+++ b/code/web/release_notes/25.11.00.MD
@@ -188,6 +188,14 @@
 //yanjun
 ### Other Updates
 - Add a Clear Cached Values control under System Variables so admins can purge cached values manually. (DIS-1531) (*YL*)
+- Add Clear Cached Values Action to every upgrade. (DIS-1531) (*YL*)
+
+<div markdown="1" class="settings">
+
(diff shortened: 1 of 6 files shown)


mdnoble73 (changes requested):
This doesn't need to be a column. It can just be an additional action that can be clicked. 

We also want to automatically clear the cache when performing upgrades (at the end of DB maintenance)

mdnoble73 (approved):

PR #3283, opened 2025-11-05
DIS-1542 delete list return check

lathomas64:
to test:
* create a list for a user
* In postman or your favorite API testing tool make a request to /API/ListAPI?method=deleteList with the id of your list, and the username and password of your user added
* if the deletion is successful you should get the normal list deleted successfully message.
* if there are any issues, like for example you don't have an active user since you just hit the api directly without a session from logging in, when the delete fails you should get a message saying such instead of a false positive saying a delete was successfull when it wasn't.

### DIFF
--- a/code/web/services/API/ListAPI.php
+++ b/code/web/services/API/ListAPI.php
@@ -1325,12 +1325,32 @@ class ListAPI extends AbstractAPI {
 				$userCanEdit = $user->canEditList($list);
 				if ($userCanEdit) {
 					$optOutOfSoftDeletion = !empty($_REQUEST['optOutOfSoftDeletion']) && filter_var($_REQUEST['optOutOfSoftDeletion'], FILTER_VALIDATE_BOOLEAN);
-					$list->delete(false, $optOutOfSoftDeletion);
-					return [
-						'success' => true,
-						'title' => translate(['text' => 'Success', 'isPublicFacing' => true]),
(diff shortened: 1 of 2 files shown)


code/web/services/API/ListAPI.php
```diff
@@ -1325,12 +1325,32 @@ function deleteList() {
...
+							'success' => false,
+							'title' => translate(['text' => 'Error', 'isPublicFacing' => true]),
+							'message' => translate(['text' => 'We attempted to delete your list but it looks like no changes occurred', 'isPublicFacing' => true]),
+						];
+					} else { // some kind of DB error happened and we almost certainly did not successfully delete the list.
+						return [
+							'success' => false,
+							'title' => translate(['text' => 'Error', 'isPublicFacing' => true]),
+							'result' => $result,
+							'message' => translate(['text' => $list->getLastError(), 'isPublicFacing' => true]),
```
mdnoble73: We shouldn't push this back since it would contain the raw SQL of what is wrong.

mdnoble73 (changes requested):

code/web/services/API/ListAPI.php
lathomas64: fair point, logging the error and sending back a more generic message instead.

mdnoble73 (approved):

PR #3296, opened 2025-11-07
Dis 1485 external request log request type

lathomas64:
to test:
* go to greenhouse
* go to External Request Settings
* create a new setting for an api route in the form API.method (example: SystemAPI.getLibraryInfo)
* through LiDA or your favorite means of API testing access the route (API access should not be granted for your IP because that route often skips the externalRequestlogentry)
* check the external request logs to see an entry for the route 
* repeat with a setting for just the api instead of the full path
* you can comment out the minimum for expireDate in ExternalRequestSettings.php and set the expire date in the past to confirm it does not log if the expire date is already passed.
* if enabled is false it should also not log the entry.


lathomas64:
just realized the window for 25.11 is already past and I need to redo this to point at 25.12

lathomas64:
closing this since i don't see a straightforward way to switch which branch its targeting

PR #3304, opened 2025-11-09
DIS-1569 Use the Schema API instead of downloading the schema file

tomascohen:
- Use /schema/fields instead of /admin/file?file=schema.xml
- Compatible with modern Solr versions and SolrCloud
- Maintains same error handling for offline Solr detection

Test Plan:
1. SolrCloud: Deploy cluster, verify schema fields load in admin interface
2. Single Node: Test catalog browsing and search functionality
3. Error Handling: Stop Solr, verify 'Solr is not currently running' error
4. Cache: Test ?reload=1 parameter refreshes schema from API
5. Manual: curl http://localhost:8983/solr/schema/fields | jq '.fields[0]'

### DIFF
--- a/code/web/release_notes/25.12.00.MD
+++ b/code/web/release_notes/25.12.00.MD
@@ -141,6 +141,10 @@
 - Add support for ChiliFresh as a cover image source. ([DIS-1602](https://aspen-discovery.atlassian.net/browse/DIS-1602)) (*GMC*)
 
 
+//tomas
+### Indexing Updates
+- Add SolrCloud support by using the Schema API instead of downloading the schema file. (DIS-1569) (*TC*)
+
 ## This release includes code contributions from
 ### ByWater Solutions
 - Leo Stoyanov (LS)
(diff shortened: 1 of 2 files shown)


tomascohen:
This needs testing, please.

tomascohen:
After extensive testing, I noticed I was getting the URL paths wrong. The new patchset changes 'host' for 'baseurl' to better understand what's going on, and includes some comments in the code too.

mdnoble73:
@tomascohen before this gets merged it will need release notes as well

mdnoble73:
@tomascohen it looks like this is using spaces rather than tabs for indents. Could you correct?

mdnoble73 (changes requested):
The readme file should not have had spaces converted to tabs, just Solr.php. Can you revert those changes?

tomascohen:
I've done, but I cannot find the problem with tabs vs. spaces in Solr.php

mdnoble73 (approved):

PR #3347, opened 2025-11-17
DIS-1596:  Remove requirement of setting LOCAL_USER_ID environment variable

lucasmontoya13:
In the context of containers, this type of variable should not be mandatory and should have an alternative value instead of closing.
Also, following Docker good practices, it shouldn’t modify permissions of anything of the host side since it could be a risk in terms of security.

This patch removes the mandatority of setting the LOCAL_USER_ID environment variable and removes a chown command which modifies host permissions.

### DIFF
--- a/code/web/release_notes/25.12.00.MD
+++ b/code/web/release_notes/25.12.00.MD
@@ -118,6 +118,9 @@
 
 
 //lucas
+### Docker Updates
+- Removed requirement of setting LOCAL_USER_ID environment variable. (DIS-1596) (*LM*)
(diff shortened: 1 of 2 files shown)


tomascohen:
Thanks Lucas. This is important and we need it in a maintenance 25.11 release.

mdnoble73 (approved):

mdnoble73:
@lucasmontoya13 can you correct conflicts please

mdnoble73 (approved):

mdnoble73 (approved):

PR #3399, opened 2025-11-24
DIS-1602: Add support for ChiliFresh cover images

gmcharlt:


gmcharlt:
Test plan (assumes you have permission or a subscription from ChiliFresh)

- Enable the integration under ChiliFresh Settings under Third Party Enrichment
- Verify that cover images are displaying as expected
- If you have a code from ChiliFresh for a fallback / generic art image specific to your subscription, update Aspen's ChiliFresh settings to use it, then verify that a record that does not have a recognized ISBN/ISSN/UPC displays the ChiliFresh fallback image rather than the one generated by Aspen.

mdnoble73 (approved):

PR #3506, opened 2025-12-18
DIS-1716:  Remove koha configuration from aspen initialization

lucasmontoya13:
This patch removes the automatic Koha ILS configuration that was previously done via Docker environment variables during container startup.

  1. ILS-specific scripts don't belong in Docker initialization: Having a dedicated script specifically for Koha creates an inconsistency—no other ILS drivers (Evergreen, Polaris, Sierra, Symphony, etc.) have initialization scripts. This approach doesn't scale and creates maintenance burden.

  2. Admin interface is the proper place: Koha connections should be configured through the Aspen Administration interface (System Administration → Account Profiles), just like every other ILS integration. This provides:
    - Validation and error handling
    - Consistent user experience
    - No need for container restarts to change ILS settings
    
  3. Production deployments: In production environments, ILS configurations are typically set once and rarely change. Having them in environment variables adds complexity without benefit, and risks exposing sensitive credentials in Docker Compose files or environment variable dumps.
  
  4. The existence of ADB, and the fact that it is prepared to create a development environment for testing Aspen and Koha, is another reason that allows us to remove the internal Koha initialization script.


mdnoble73 (approved):

mdnoble73 (changes requested):
@lucasmontoya13 could you resolve conflicts on this please

PR #3517, opened 2025-12-20
DIS-1733: Show Cancelled Holds in "My Account" (Polaris)

K-Alette:
Add setting in primary configuration to "Show Cancelled Holds in My Account" for Polaris libraries 
Setting is set to default as on
Placing a hold from the cancelled list in "My Account" places a hold on the same item that was cancelled but does not remove it from the cancelled holds list in "My Account" or Polaris

Update release notes


code/web/Drivers/Polaris.php
```diff
@@ -853,7 +861,7 @@ function placeItemHold($patron, $recordId, $itemId, $pickupBranch, $cancelDate =
 		}
 
 		//Need to set the Workstation
-		$body->WorkstationID = $this->getWorkstationID($patron);
+		$body->WorkstationID = "1";
```
mdnoble73: @K-Alette we need this to not be hard coded.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #3530, opened 2025-12-31
DIS-1766: FIX: Sideloads Usage Graphs

Chloe070196:


code/web/services/SideLoads/UsageGraphs.php
```diff
@@ -39,7 +39,7 @@ function canView(): bool {
 	*/
 	private function getSideloadIdBySideLoadName($name): int {
 		$sideload = new SideLoad();
-		$sideload->whereAdd('name = "' . $name .'"');
+		$sideload->whereAdd('name = ' . $sideload->escape($name));
```
mdnoble73: @Chloe070196 this can be done by simply doing $sideload->name = $name which will handle the escaping automatically.

mdnoble73 (changes requested):

code/web/services/SideLoads/UsageGraphs.php
Chloe070196: That would indeed be more straightforward, thanks for pointing it out!

mdnoble73 (approved):

PR #3541, opened 2026-01-02
DIS-1689: customisable usage period

Chloe070196:
Depends on [4346749](https://github.com/Aspen-Discovery/aspen-discovery/commit/43467499b8465164d6d2e70e9edbb92a25df997a)

### DIFF
--- a/code/web/sys/AbstractUsage.php
+++ b/code/web/sys/AbstractUsage.php
@@ -13,4 +13,18 @@ abstract class AbstractUsage extends DataObject {
 		}
 		return "{$this->month}-{$this->year}"; // monthly is the default
 	}
+	public function buildCustomPeriodQuery(array $custom): void {
+		$escapedPeriodDuration = $this->escape($custom['customUsagePeriodDuration']);
+		$escapedPeriodStart = $this->escape($custom['customUsagePeriodStart']);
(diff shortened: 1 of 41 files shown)


code/web/sys/AbstractUsage.php
```diff
@@ -0,0 +1,42 @@
+<?php /** @noinspection PhpMissingFieldTypeInspection */
+
+abstract class AbstractUsage extends DataObject {
+	public function getCurPeriod($timeframes) {
+		if (in_array('day', $timeframes)) {
+			return "{$this->day}-{$this->month}-{$this->year}";
```
mdnoble73: This should be internationalized based on local. In the U.S. Month - Day - Year would be used.

mdnoble73 (changes requested):
See the comment on internationalization. Also, how do we want to show things if grouping by day prior to data being recorded by day? It currently shows the day as 0 which could be confusing. I wonder if we need a note someplace saying when daily tracking started?

code/web/sys/AbstractUsage.php
Chloe070196: As discussed over Slack - absolutely agree on internationalization being a priority. Changes will need to await a separate Aspen submission that makes a new smarty plugin available which effectively  allows for date formats to be set based on locale.

As for the '0', I think adding a tooltip that explains its meaning is certainly doable, and only fair. That will be added along with the internationalization related changes

However, I would see tracking and displaying the date at which daily tracking started as a further enhancement beyond the scope of this development.

mdnoble73:
@Chloe070196 when this is ready, please update to the latest release and pull it out of draft mode

Chloe070196:
@mdnoble73 waiting on a dependency on that one. I will update this once it has been submitted into core.

Chloe070196:
With DIS-2605 submitted, I'm working to get this ready for 26.07.00

code/web/sys/AbstractUsage.php
Chloe070196: Sorted! Date format internationalisation implemented using the utils introduced by DIS-2605

mdnoble73 (approved):

PR #3565, opened 2026-01-05
DIS-1796 Refactor getDefaultColor to use hashtable

JonahCWilson:
I noticed that getDefaultColor is used in both aspen.js as well as in aspen/admin.js.  I checked to ensure that the function bodies are identical.

I pulled the properties into a hash table with propName:hexValue, then abstracted out the document.getElementById call to a separate function.

There is not currently a work-item for this in Jira, it's just something I noticed when looking through the code.

### DIFF
--- a/code/web/interface/themes/responsive/js/aspen.js
+++ b/code/web/interface/themes/responsive/js/aspen.js
@@ -5320,6 +5320,125 @@ AspenDiscovery.Account = (function () {
 	};
 }(AspenDiscovery.Account || {}));
 AspenDiscovery.Admin = (function () {
+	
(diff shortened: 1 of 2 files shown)


mdnoble73:
@JonahCWilson aspen.js is built from the other javascript files so they will always match. That let's us maintain the javascript more easily with fewer conflicts while still delivering a single file to the end user. Can you create a Jira for this and resubmit for 26.02?

JonahCWilson:
I noticed after the fact that the merge_javascript file is pulling everything together, so that makes sense.  And yes, I'll update Jira and resubmit for 26.02.  Thanks!

PR #3573, opened 2026-01-06
DIS-1320: display and update primary contact method

Chloe070196:
This is blocked by [Koha bug 28701](https://bugs.koha-community.org/bugzilla3/show_bug.cgi?id=28701). Raising as draft for now.


gmcharlt:
Noting that the Koha enhancement is available in 25.11 and later.

Chloe070196:
Thank you for the note @gmcharlt , I'd lost track of that one. I'll rebase and submit to 26.08.00 since it is a bug fix

Chloe070196:
I'm having this target 26.08.00 in case the fix is wanted in that version, I know this has been held off longer than it should have (my bad!)

However, I also do realise we're quite late in the testing period, and I'll happily target 26.09.00 instead if that is better for testers. Whichever works best!

Chloe070196:
Rebased and now submitting to 26.09.00

mdnoble73 (approved):

PR #3585, opened 2026-01-08
DIS-1404 progressive web application

lathomas64:
Phase one of the PWA
- Updating Notifications to handle expo or firebase Tokens
- putting in first steps towards the PWA allowing installation as an App with basic Notification support. 

The biggest thing here is how we are touching the notifications. This should leave LiDA and expo notifications unaffected as far as I can tell.

to test:
* run database maintenance to add module and settings

    enable the Aspen Progressive Web App module

  enable all permissions related to the module for your user

    add Aspen PWA settings

    for firebase settings you can get settings from [console.firebase.google.com](http://console.firebase.google.com/) and make a project

    open your project

    open the app within your project

    on the general tab under SDK setup and configuration click config to get the vapid key

    cloud messaging > web configuration > web push certificates generate key pair or use the existing public key if one is already visisble (for example if you’re using an existing firebase app)

    Visit the page from a phone and allow notifications

    the aspen progressive web application testing tool should send notifications that reach your device

    LiDA notifications should be unaffected by these changes whether or not you enabled Aspen Progressive Web App


code/web/interface/themes/responsive/AspenPWA/README.md
mdnoble73, line 1: How is this readme intended to be used/viewed? This isn't consistent with how other documentation in Aspen works.

mdnoble73 (changes requested):
@lathomas64 is this ready for review? It needs testing notes in Jira as well as additional documentation of what the Jira is intended to do before I can review it fully.

mdnoble73 (changes requested):
This needs testing notes in Jira before merge

install/aspen.sql
mdnoble73, line 1: This needs an update using a database maintenance script to update existing installs. Updating aspen.sql will only update a new install.

code/web/release_notes/26.02.00.MD
```diff
@@ -15,6 +15,8 @@
 // yanjun
 
 // imani
+- Updated Notifications to be able to handle multile token types (DIS-1404)(*IT*)
```
mdnoble73: This needs additional information about the Progressive Web App functionality

lathomas64:
added testing notes in Jira as well as notes mentioning the smaller scope of that ticket after the conversation in Triage this week about bigger projects. will remove the subtask tickets.

code/web/interface/themes/responsive/AspenPWA/README.md
lathomas64: Its quick reminder notes for setup I plan to have more robust documentation in confluence.

code/web/sys/Account/User.php
```diff
@@ -4921,7 +4907,11 @@ public function getAdminActions() : array {
 		$sections['support']->addAction(new AdminAction('Help Center', 'View the Help Center for Aspen Discovery.', 'https://help.aspendiscovery.org'), true);
 		$sections['support']->addAction(new AdminAction('API Documentation', 'View available OpenAPI specifications for Aspen Discovery APIs.', '/API/Documentation'), true);
 		$sections['support']->addAction(new AdminAction('Release Notes', 'View release notes for Aspen Discovery which contain information about new functionality and fixes for each release.', '/Admin/ReleaseNotes'), true);
-
+		
+		if (array_key_exists('Aspen Mobile', $enabledModules)){
```
mdnoble73: This does not follow currently established style. Happy to discuss if we want to make a change, but we shouldn't have this be a special case.

code/web/services/AspenMobile/Manifest.php
```diff
@@ -0,0 +1,64 @@
...
+			'id' => $setting->manifestID,
+			'name' => 'Aspen Mobile',
+			'short_name' => 'Aspen Mobile',
+			'theme_color' => '#000000',
+			'start_url' => $setting->startURL,
+			'description' => 'testing this',
+			'icons' => [
+				[
+					//'src' => '/API/SystemAPI?method=getLogoFile&themeId=1&type=appIcon&slug='.$setting->slug,
+					'src' => 'https://placehold.co/512x512?text='.$setting->slug,
```
mdnoble73: Should this be returning an actual icon?

code/web/services/AspenMobile/NotificationTestingTool.php
```diff
@@ -0,0 +1,71 @@
...
+		$formattedContents = print_r($var, true);
+		if ($formattedContents !== false) {
+			$contents .= $formattedContents;
+		}
+		$contents .= '</pre>';
+		return $contents;
+	}
+
+	function getInstructions(): string {
+		return 'https://docs.expo.dev/push-notifications/sending-notifications/#errors';
```
mdnoble73: These instructions go to expo is that correct?

code/web/sys/Account/User.php
```diff
@@ -5783,6 +5774,7 @@ public function getNotificationPushTokenByDevice(): array {
 		$tokens = [];
 		$obj = new UserNotificationToken();
 		$obj->userId = $this->id;
+		$obj->tokenType = $_REQUEST['tokenType'] ?? "expo";
```
mdnoble73, line 5874: Does this mean a user can only use expo or firebase notifications?

mdnoble73 (changes requested):
Technical Committee would like this moved to 26.03 due to the number of issues still outstanding and need for additional testing with a pull request of this magnitude. Technical Committee would also like to see things named AspenMobile changed to AspenPWA or similar to avoid confusion with LiDA and responsive design.

code/web/services/API/SystemAPI.php
```diff
@@ -1221,6 +1245,16 @@ function getMaterialsRequestForm() : array {
 		];
 	}
 
+<<<<<<< DIS-1404-progressive-web-application
+	function getFirebaseSettings() {
```
catsoup11789: Why do we need an API endpoint to expose Firebase settings? If the wrong person got a hold of this (its public) it could expose Firebase tokens that could be misused.

code/web/services/AspenMobile/NotificationTestingTool.php
```diff
@@ -0,0 +1,71 @@
...
+		return $breadcrumbs;
+	}
+
+	function getActiveAdminSection(): string {
+		return 'aspen_mobile';
+	}
+
+	function canView(): bool {
+		return UserAccount::userHasPermission([
+			'Send Notifications to All Libraries',
```
catsoup11789: These are permissions for the Aspen LiDA module, and require that module to be active to assign them. The PWA should have its own independent permissions, maybe based off a PWA module.

code/web/services/AspenMobile/Settings.php
```diff
@@ -0,0 +1,70 @@
...
+		return $breadcrumbs;
+	}
+
+	function getActiveAdminSection(): string {
+		return 'aspenMobile';
+	}
+
+	function canView(): bool {
+		// TODO should we change this to a Aspen Mobile specific Permission?
+		return UserAccount::userHasPermission('Administer Aspen LiDA Settings');
```
catsoup11789: Should have separate permissions not dependent on Aspen LiDA

code/web/sys/AspenPWA/Setting.php
```diff
@@ -0,0 +1,180 @@
...
+				'description' => 'description here',
+				'maxLength' => 100,
+				'required' => true,
+
+			],
+			'serviceAccount' => [
+				'property' => 'serviceAccount',
+				'type' => 'textarea',
+				'label' => 'Service Account',
+				'description' => 'Contents of your Service Account json file from firebase. Service Accounts > Generate new private key > copy contents of the downloaded file into this field.',
```
catsoup11789: service account json files contain private keys that should be kept secure, do we want them exposed here? @mdnoble73 any suggestions on this one?

code/web/interface/themes/responsive/js/aspen/serviceWorker.js
```diff
@@ -0,0 +1,109 @@
...
+		const messaging = firebase.messaging();
+
+		console.log(messaging);
+		messaging.onMessage(function(payload) {
+			console.log('message recieved');
+		});
+		messaging.onBackgroundMessage(function(payload) {
+			console.log('[firebase-messaging-sw.js] Received background message ', payload);
+			// Customize notification here
+			const notificationTitle = 'Background Message Title';
```
catsoup11789: should this contain actual data for the notification?

code/web/interface/themes/responsive/js/aspen/serviceWorker.js
catsoup11789, line 1: should remove or comment out console.log calls

code/web/sys/DBMaintenance/aspen_mobile_updates.php
catsoup11789, line 1: this should be included in the applicable release db update file

code/web/interface/themes/responsive/layout.tpl
```diff
@@ -52,6 +52,11 @@
 		{/if}
 		<link rel="search" type="application/opensearchdescription+xml" title="{$site.title|escape} Catalog Search" href="/Search/OpenSearch?method=describe">
 		{include file="cssAndJsIncludes.tpl"}
+		{if (array_key_exists('Aspen Mobile', $enabledModules))}
+			{* manifest.json produced from code/web/services/AspenMobile/manifest.php *}
+			<link rel="manifest" href="/manifest.json"/>
+			<script type="module" src="/interface/themes/responsive/js/aspen/initFCM.js"></script>
```
catsoup11789: is this just prompting the user to allow notifications on any page load? this should probably be a setting in My Account and allow the user to select which types of notifications they want (similar to LiDA's notification permission screen). These settings are very likely going to be required for various digital privacy laws/regulations.

code/web/sys/Account/User.php
catsoup11789: @mdnoble73 the user notification table lets us store as many tokens for a patron as they want (i.e. original use case was if a patron had more than 1 device with LiDA installed, they would need separate tokens), this should be ok. 

****LiDA will likely need updated though to check the token type in the case the user has both the app and PWA on the same device.****

code/web/services/AspenMobile/Manifest.php
```diff
@@ -0,0 +1,64 @@
...
+		// if we have no settings
+		if(!$setting->find(true))
+		{
+			return ['success' => false,'message'=>'settings not found'];
+		}
+
+		return [
+
+			'id' => $setting->manifestID,
+			'name' => 'Aspen Mobile',
```
catsoup11789: This should the library name (at the very least) or something custom the library can set in the PWA settings.

code/web/interface/themes/responsive/layout.tpl
catsoup11789: I'm thinking we should only be including these header files if the library has actually setup the PWA and not just enabled the module.

code/web/sys/AspenPWA/Setting.php
catsoup11789, line 1: These settings should allow libraries to select which libraries and/or locations the PWA settings apply to.

code/web/sys/Account/UserNotificationToken.php
```diff
@@ -10,6 +10,7 @@ class UserNotificationToken extends DataObject {
 	public $notifyCustom;
 	public $notifyAccount;
 	public $onboardAppNotifications;
+	public $tokenType; //Expo or Firebase
```
catsoup11789, line 13: LiDA will also need updated to look for the "Expo" tokenType when it fetches tokens for determining if the user has enabled notifications or not (specifically in the case the user has both LiDA and the PWA enabled and the deviceModel is the same).

code/web/sys/Account/User.php
lathomas64: LiDA shouldn't need to be updated here. LiDA should only receive messages sent to LiDA tokens.

code/web/services/API/SystemAPI.php
lathomas64: this is just returning what the frontend needs for the initializeApp method to set up messaging. I should maybe rename this to getFirebaseMessagingConfig. Would that work better?

code/web/sys/AspenPWA/Setting.php
lathomas64: serviceAccount is added to getEncryptedFieldNames so it is encrypted in the database.

I will modify this to not show current value on the page as well.

code/web/sys/Account/User.php
catsoup11789: > LiDA shouldn't need to be updated here. LiDA should only receive messages sent to LiDA tokens.

It's not an issue of LiDA receiving PWA notifications, but LiDA checks the list of tokens for a user and only expects one token for the device when determining what permissions have been set for the Permission screens. If the user has both PWA and LiDA installed and notifications enabled on the same device, there will be two tokens available for LiDA to find. LiDA will need be updated to look at the tokenType to know which one to pick in this case.

lathomas64: Couldn't they could already have multiple tokens if they logged into LiDA from multiple devices?

catsoup11789: > Couldn't they could already have multiple tokens if they logged into LiDA from multiple devices?

Yes but it is checking the device name to determine which one to select for that device. The issue here is if the PWA and Expo token have the same device name assigned (i.e. iPhone 16)

lathomas64:
I've address all the big issues and am ready to try and sneak into 26.03 if you'll have this.

mdnoble73 (changes requested):
@lathomas64 this needs the module renamed from AspenMobile to AspenPWA or similar to avoid confusion with Aspen LiDA and the base responsive functionality as previously requested. 

Could you also record a demo showing the functionality working so we can see how it's intended to function.

lathomas64:
Some folks were very keen on the Aspen Mobile name but I've got the greenlight to change that I'm just making sure I didn't break anything renaming things and then I'll push that up.

mdnoble73:
@lathomas64 when this is ready, please remove from draft status and submit to the current release

lathomas64:
> @lathomas64 when this is ready, please remove from draft status and submit to the current release

by current release did you mean 26.03 or 26.04? I assumed the latter but i wanted to make sure I was understanding you correctly.

code/web/sys/DBMaintenance/version_updates/26.03.00.php
```diff
@@ -93,7 +93,78 @@ function getUpdates26_03_00(): array {
...
+				"INSERT IGNORE into `permissions` (name, sectionName, requiredModule, weight, description) VALUES ('Send Aspen Progressive Web Application(PWA) Notifications to Home Location','Aspen Progressive Web Application(PWA)', 'Aspen Progressive Web Application(PWA)', 6, 'Controls if the user can send notifications to Aspen Progressive Web Application(PWA) users from their home location.');",
+				"INSERT IGNORE into `permissions` (name, sectionName, requiredModule, weight, description) VALUES ('Send Aspen Progressive Web Application(PWA) Notifications to Home Library Locations','Aspen Progressive Web Application(PWA)', 'Aspen Progressive Web Application(PWA)', 6, 'Controls if the user can send notifications to Aspen Progressive Web Application(PWA) users for all locations that are part of their home library.');",
```
catsoup11789: The name on this last permission gets truncated due to the column being VARCHAR(75)

code/web/sys/Account/User.php
```diff
@@ -4952,7 +4938,12 @@ public function getAdminActions() : array {
 		$sections['support']->addAction(new AdminAction('Help Center', 'View the Help Center for Aspen Discovery.', 'https://help.aspendiscovery.org'), true);
 		$sections['support']->addAction(new AdminAction('API Documentation', 'View available OpenAPI specifications for Aspen Discovery APIs.', '/API/Documentation'), true);
 		$sections['support']->addAction(new AdminAction('Release Notes', 'View release notes for Aspen Discovery which contain information about new functionality and fixes for each release.', '/Admin/ReleaseNotes'), true);
-
+		
+		if (array_key_exists('Aspen Progressive Web Application(PWA)', $enabledModules)){
+			$sections['Aspen Progressive Web Application(PWA)'] = new AdminSection('Aspen Progressive Web Application(PWA)');
+			$sections['Aspen Progressive Web Application(PWA)']->addAction(new AdminAction('Notification Test Tool', 'Aspen Progressive Web Application(PWA) notification test tool', '/AspenPWA/NotificationTestingTool'), [true]);
```
catsoup11789: the last parameter here should be a string of the permission that is required to access each section action; as-is, even if a user has all permissions related to PWA, they still can't see any of the Admin menu options...

catsoup11789: sorry, I wasn't completely accurate on this one, but it should still have the permissions required instead of true so that the items won't show if the user doesn't actually have permission to view them. BUT, the "(PWA)" in the section name is breaking the javascript on the accordion toggle, preventing the opening/closing of the section and accessing the settings without knowing the URL when not on the Admin landing page.

code/web/services/AspenPWA/NotificationTestingTool.php
```diff
@@ -0,0 +1,71 @@
...
+		return 'aspen_mobile';
+	}
+
+	function canView(): bool {
+		return UserAccount::userHasPermission([
+			'Send Aspen Progressive Web Application(PWA) Notifications to All Libraries',
+			'Send Aspen Progressive Web Application(PWA) Notifications to All Locations',
+			'Send Aspen Progressive Web Application(PWA) Notifications to Home Library',
+			'Send Aspen Progressive Web Application(PWA) Notifications to Home Location',
+			'Send Aspen Progressive Web Application(PWA) Notifications to Home Library Locations',
```
catsoup11789: This permission won't work since the name is truncated in the database

code/web/services/API/SystemAPI.php
```diff
@@ -50,6 +53,7 @@ function launch() : void {
 					'getCatalogStatus',
 					'getLocations',
 					'getMaterialsRequestForm',
+					'getFirebaseMessagingConfig',
```
catsoup11789: This check only works for LiDA and is part of how it authenticates API requests, so you can probably safely remove this method from being returned here.

code/web/services/AspenPWA/Manifest.php
```diff
@@ -0,0 +1,72 @@
...
+		//TODO we should return an error code instead of 200
+		// if we have no settings
+		if(!$setting->find(true))
+		{
+			http_response_code(404);
+			return ['success' => false,'message'=>'settings not found'];
+		}
+		$theme = new Theme();
+		$theme->id = $setting->themeId;
+		$theme = $theme->find(true);
```
catsoup11789: This will make $theme just equal true or false and always use the fallback color, you should wrap $theme->find(true) in an if statement, and then call the theme object within that

code/web/services/AspenPWA/Icon.php
```diff
@@ -0,0 +1,35 @@
...
+ * /pwa-icon.png to this function
+ */
+class AspenPWA_Icon extends Action {
+
+	function launch()
+	{
+		$success = true;
+		//TODO we should return an error code instead of 200
+		// if we have no settings
+		if($setting->find(true))
```
catsoup11789: You need to define $setting here first, else it returns an error: Call to a member function find() on null

code/web/services/AspenPWA/Icon.php
```diff
@@ -0,0 +1,35 @@
...
+		{
+			$_REQUEST['themeId'] = $setting->themeId;
+		} else {
+			$_REQUEST['themeId'] = 1;
+		}
+
+		$api = new SystemAPI('internal');
+		$_REQUEST['type'] = "logoApp";
+		$api->getLogoFile();
```
catsoup11789, line 30: You should update the description of the logoApp property in Theme since right now it only says it's used for Aspen LiDA

code/web/sys/Account/User.php
catsoup11789: This still needs to be accounted for in LiDA. You could also go the route of adding "Firebase" or "PWA" at the end of the device name if you didn't want to modify LiDA.

code/web/interface/themes/responsive/MyAccount/myNotificationPreferences.tpl
```diff
@@ -0,0 +1,135 @@
...
+					<div class="alert alert-success">{$msg}</div>
+				{/foreach}
+			{/if}
+			{if !empty($offline)}
+				<div class="alert alert-warning"><strong>{translate text=$offlineMessage isPublicFacing=true}</strong></div>
+			{else}
+				<div class="grant-notification-permissions">
+					<label for="allowNotifications" class="control-label">{translate text='Allow Notifications' isPublicFacing=true}</label>&nbsp;
+					{if $edit == true}
+						<input type="checkbox" class="form-control" name="allowNotifications" id="allowNotifications" {if $profile->noPromptForUserReviews==1}checked='checked'{/if} data-switch="">
```
catsoup11789: $profile->noPromptForUserReviews is tied to whether or not libraries allow users to leave reviews... not sure why its being called on this page?

code/web/interface/themes/responsive/MyAccount/myNotificationPreferences.tpl
```diff
@@ -0,0 +1,135 @@
...
+					{else}
+						{if $profile->noPromptForUserReviews==0} {translate text='No' isPublicFacing=true}{else} {translate text='Yes' isPublicFacing=true}{/if}
+					{/if}
+				</div>
+				<script type="module" type="text/javascript">
+					import { initialize, appToken } from '/interface/themes/responsive/js/aspen/initFCM.js';
+					{* Initiate any checkbox with a data attribute set to data-switch=""  as a bootstrap switch *}
+					{literal}
+					$(function(){ $('input[type="checkbox"][data-switch]').bootstrapSwitch()});
+					$("#usernameHelpButton").on('click', function() {
```
catsoup11789: why is this javascript here?

code/web/interface/themes/responsive/MyAccount/myNotificationPreferences.tpl
```diff
@@ -0,0 +1,135 @@
...
+						return false;
+					})
+					function handleAllowNotifications() {
+						var allow = $("#allowNotifications").is(":checked");
+						if(allow)
+						{
+							let token = initialize();
+							console.log(appToken);
+							console.log("Token was: "+token);
+							if(Notification.permission === "granted")
```
catsoup11789: This does weird things when I clear Notification settings for the website (so it should prompt me again, it is neither in Deny or Allow state) in Safari, I do not get prompted again to enable and the toggle is forever stuck in an "ON" state. Toggling it "OFF" only stays until I refresh the page. I noticed my user never populated a row in user_notification_tokens even when my browser did initially prompt me.

code/web/sys/AspenPWA/Setting.php
```diff
@@ -0,0 +1,294 @@
...
+				'type' => 'enum',
+				'label' => 'Theme',
+				'values' => $availableThemes,
+				'description' => 'The theme which should be used for the application',
+			],
+			'manifestID' => [
+				'property' => 'manifestID',
+				'type' => 'text',
+				'label' => 'Manifest ID',
+				'description' => 'A Unique identifier for your web application: https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/id',
```
catsoup11789: Make all these "note" property types so they are just always displaying. It was very hard to read the tooltips, and you can't click links or copy text from it.

I also found a lot of these very hard to find.

code/web/sys/AspenPWA/Setting.php
```diff
@@ -0,0 +1,294 @@
...
+			],
+			'slug' => [
+				'property' => 'slug',
+				'type' => 'text',
+				'label' => 'Slug',
+				'description' => 'slug to identify the application',
+				'maxLength' => 50,
+				'required' => true,
+			],
+			'sha256CertFingerprint' => [
```
catsoup11789, line 120: Is it required to have a published app in the Google Play store to have a PWA? Need some clarification on this property.

code/web/sys/AspenPWA/Setting.php
```diff
@@ -0,0 +1,294 @@
...
+			'firebaseAppID' => [
+				'property' => 'firebaseAppID',
+				'type' => 'text',
+				'label' => 'Firebase application ID',
+				'description' => 'description here',
+				'maxLength' => 50,
+				'required' => true,
+
+			],
+			'firebaseMeasurementID' => [
```
catsoup11789: The "Firebase Measurement ID" is tied to Google Analytics being enabled on the project, but you didn't clarify that Google Analytics is a requirement for having a PWA. Is that true? If not, you should make this field not required. If it is, you should clarify and also update the description to make it useful.

code/web/sys/AspenPWA/Setting.php
```diff
@@ -0,0 +1,294 @@
...
+				'property' => 'firebaseMeasurementID',
+				'type' => 'text',
+				'label' => 'Firebase Measurement ID',
+				'description' => 'description here',
+				'maxLength' => 50,
+				'required' => true,
+
+			],
+			'vapidKey' => [
```
catsoup11789, line 181: I had to Google what this was. My Firebase just called them "Web Push certificates" and never once used the term Vapid.

code/web/services/MyAccount/MyNotificationPreferences.php
catsoup11789, line 1: This page would probably load a lot faster if it wasn't trying to get all this irrelevant data.

code/web/interface/themes/responsive/MyAccount/myNotificationPreferences.tpl
catsoup11789: This toggle state is saved by the browser, so when I logged into a new user without notifications enabled, it already was toggled (though I couldn't see it because of another issue), and all the notification type toggles were displaying already, and on the "ON" state, even though they were off in the database. I had to toggle them twice before they synced up.

code/web/interface/themes/responsive/MyAccount/myNotificationPreferences.tpl
catsoup11789: Wanted to add I do get prompted again for permissions when in Chrome after resetting.

code/web/interface/themes/responsive/MyAccount/myNotificationPreferences.tpl
catsoup11789: Once the notification toggle types display, it hides the "Enable Notifications" toggle forever until I reset the permissions to get prompted again (Except in Safari), so you can only turn it off in your browser settings for the site. This toggle should not go away, and when it's toggled off, we should delete the token from the database.

code/web/services/API/SystemAPI.php
```diff
@@ -1217,6 +1233,28 @@ function getMaterialsRequestForm() : array {
 		];
 	}
 
+	//only returns the portion of settings needed by
+	//the front end to get a token
+	function getFirebaseMessagingConfig() {
```
catsoup11789, line 1243: Since libraries are allowed to have different settings, you should be making sure you're fetching the correct one instead of whatever is first.

code/web/interface/themes/responsive/MyAccount/account-sidebar.tpl
catsoup11789, line 1: you will also need to add a link in account-menu.tpl, unfortunately the current setup doesn't include this template in mobile, so when you're in mobile you can't access the notifications page.

code/web/interface/themes/responsive/MyAccount/myNotificationPreferences.tpl
catsoup11789: Finally after completely closing down Safari, I got the notification permission to prompt again. But still having issues. I allowed notifications, but it did not create a new token for me (it says undefined in console.log, and does not exist in the db). Toggling notifications "ON" did turn on the display of the notification types, but no matter what I do, they don't create (or update) a token for Safari so I can never actually receive notifications (and its never an option in the testing tool either). I do see an error that I did not see in Chrome when initFCM.js runs: "Notification prompting can only be done from a user gesture." on Notification.requestPermission().then((permission).....

code/web/sys/Account/User.php
catsoup11789: After some testing, I noticed on my Mac it just puts the device model as "Macintosh" which got really confusing when I was testing multiple browsers and I had no way to distinguish tokens between them (yes likely just a testing scenario to have multiple, but still should be accounted for)... maybe at least include the OS version and name of the browser? so for my Mac, it'd be "macOS 10.15.7 - Chrome"

code/web/interface/themes/responsive/layout.tpl
```diff
@@ -52,6 +52,10 @@
 		{/if}
 		<link rel="search" type="application/opensearchdescription+xml" title="{$site.title|escape} Catalog Search" href="/Search/OpenSearch?method=describe">
 		{include file="cssAndJsIncludes.tpl"}
+		{if $AspenPWAEnabled}
+			{* manifest.json produced from code/web/services/AspenPWA/manifest.php *}
+			<link rel="manifest" href="/manifest.json"/>
```
catsoup11789, line 57: This needs to load in the correct manifest depending on the active library

code/web/sys/Account/UserNotificationToken.php
catsoup11789: Resolving this one since the same discussion is happening over here on User.php https://github.com/Aspen-Discovery/aspen-discovery/pull/3585#discussion_r2764716604

code/web/services/API/SystemAPI.php
```diff
@@ -1299,7 +1337,7 @@ public function getHomeScreenLinks(): array {
 		} else {
 			return [
 				'success' => false,
-				'message' => 'Must provide either a Location Id or Library Id',
+				'error' => 'no settings found'
```
catsoup11789: I don't think you meant to change the return of getHomeScreenLinks here

code/web/sys/Account/User.php
lathomas64: pushed a change to add more info to the field and format it a little better when we are putting together multiple fields. And add PWA to the end whether or not we could get model data.

code/web/sys/Account/User.php
lathomas64: spaces in the sections were a problem too. pushing a change for adding permission sand cleaning up the section.

code/web/services/AspenPWA/Manifest.php
lathomas64: ah the field called fetchFirst doesn't make it return the first item, fun. thanks.

code/web/services/AspenPWA/Icon.php
lathomas64: ah strange I had that there but it got accidentally removed in a later commit, thanks.

code/web/sys/Account/User.php
catsoup11789: The new is so much better. Thank you! 
`Apple Macintosh macOS 10.15.7 Chrome PWA`

code/web/index.php
```diff
@@ -1269,6 +1269,42 @@ function loadModuleActionId() {
 	}catch (Exception $e) {
 		//This happens if web builder is not fully installed, ignore the error.
 	}
+	try {
+		if ($library->aspenPWASettingId != -1)
```
LiYanjun19: Should be `AspenPWASettingId` to match DB and property

code/web/sys/AspenPWA/Setting.php
```diff
@@ -0,0 +1,294 @@
...
+class AspenPWASetting extends DataObject {
+	public $__table = 'aspen_mobile_settings';
+	public $id;
+	public $name;
+	public $shortName;
+	public $description;
+	public $themeId;
+	public $autoRotateCard;
+	public $enableSelfRegistration;
+	public $showMoreInfoBtn;
```
LiYanjun19: Should we remove these three since they are not being used for AspenPWA anywhere?

code/web/index.php
```diff
@@ -1269,6 +1269,42 @@ function loadModuleActionId() {
...
+			{
+				$_GET['module'] = "AspenPWA";
+				$_GET['action'] = "Firebase";
+				$_REQUEST['module'] = "AspenPWA";
+				$_REQUEST['action'] = "Firebase";
+			}
+			else if($requestURI == "/pwa-icon.png")
+			{
+				global $logger;
+				$logger->log("processing pwa-icon.png", Logger::LOG_ERROR);
```
LiYanjun19: Should this log get removed?

code/web/services/API/SystemAPI.php
LiYanjun19: Agree with Kirstien. We should either limit the PWA to one setting or make sure we’re fetching the correct one. Right now, a few places in the code just grab the first setting, so those would need to be updated if we change this.

code/web/interface/themes/responsive/account-menu.tpl
```diff
@@ -105,6 +105,7 @@
 				</div>
 			{/if}
 			{if !empty($showUserPreferences)}<div class="header-menu-option" ><a href="/MyAccount/MyPreferences">{translate text='Your Preferences' isPublicFacing=true}</a></div>{/if}
+			{if $AspenPWAEnabled }<div class="myAccountLink"><a href="/MyAccount/MyNotificationPreferences">{translate text="Your Notification Preferences" isPublicFacing=true}</a></div>{/if}
```
LiYanjun19: Should use `class="header-menu-option"`

code/web/services/AspenPWA/Manifest.php
```diff
@@ -0,0 +1,70 @@
...
+require_once ROOT_DIR . '/sys/AspenPWA/Setting.php';
+
+class AspenPWA_Manifest extends Action {
+
+	function launch() {
+		header('Content-type: application/json');
+		header('Cache-Control: no-cache, must-revalidate'); // HTTP/1.1
+		header('Expires: Mon, 26 Jul 1997 05:00:00 GMT'); // Date in the past
+		$output = json_encode($this->build_manifest());
+		http_response_code(200);
```
LiYanjun19: http_response_code(200) overrides the 404 code from build_manifest()

code/web/services/AspenPWA/Icon.php
```diff
@@ -0,0 +1,36 @@
...
+	{
+		$setting = new AspenPWASetting();
+		$success = true;
+		//TODO we should return an error code instead of 200
+		// if we have no settings
+		if($setting->find(true))
+		{
+			$_REQUEST['themeId'] = $setting->themeId;
+		} else {
+			$_REQUEST['themeId'] = 1;
```
LiYanjun19: It would be better if we return 404 when the setting isn't found.  themeId = 1 is not a guaranteed theme; it could be deleted.

code/web/interface/themes/responsive/js/aspen/initFCM.js
```diff
@@ -0,0 +1,67 @@
+import { initializeApp } from 'https://www.gstatic.com/firebasejs/12.1.0/firebase-app.js';
+import { getMessaging } from "https://www.gstatic.com/firebasejs/12.1.0/firebase-messaging-sw.js";
+import { getToken } from "https://www.gstatic.com/firebasejs/12.1.0/firebase-messaging.js";
```
LiYanjun19: firebase-messaging-sw.js is for service worker. We can do 
`import { getMessaging, getToken } from "https://www.gstatic.com/firebasejs/12.1.0/firebase-messaging.js"; ` here

code/web/sys/Account/User.php
```diff
@@ -6644,6 +6643,53 @@ function updateSortPreferences(): void {
...
+				$expoNotification = new ExpoNotification();
+				$expoNotification->sendExpoPushNotification($body, $notificationToken->pushToken, $this->id, $notificationType);
+				$expoNotification = null;
+				$count++;
+			} else if (strcasecmp($notificationToken->tokenType, "firebase") == 0)
+			{
+				require_once ROOT_DIR . '/sys/Notifications/FirebaseNotification.php';
+				$fireBaseNotification = new FirebaseNotification();
+				$fireBaseNotification->sendPushNotification($body, $notificationToken->pushToken, $this->id, $notificationType);
+				$firebaseNotification = null;
```
LiYanjun19, line 6740: it should be `$fireBaseNotification = null;`

code/web/interface/themes/responsive/js/aspen/initFCM.js
```diff
@@ -0,0 +1,67 @@
...
+			const firebaseConfig = data.result.settings;
+			// Initialize Firebase
+			const app = initializeApp(firebaseConfig);
+
+			// Initialize Firebase Cloud Messaging and get a reference to the service
+			const messaging = getMessaging(app);
+			getToken(messaging, { vapidKey: firebaseConfig['vapidKey'] }).then((currentToken) => {
+				if (currentToken) {
+					appToken = currentToken;
+					Notification.requestPermission().then((permission) => {
```
LiYanjun19: Should we only call getToken() after permission is granted?

LiYanjun19 (changes requested):

code/web/sys/Account/User.php
lathomas64: I think these all should be lowercase b but they should definitely be matching.

code/web/sys/AspenPWA/Setting.php
lathomas64: Looks like it was dropped since i started with this.  going to drop this and measurementId which are both not required.

code/web/interface/themes/responsive/js/aspen/initFCM.js
lathomas64: moved the permission request outside of initialize entirely and only calling it if we already have permission so removing that unnecessary check here.

code/web/interface/themes/responsive/js/aspen/initFCM.js
lathomas64: artifact from when this file was trying to do both. fixed.

code/web/services/AspenPWA/Icon.php
lathomas64: agreed, fixed.

code/web/sys/AspenPWA/Setting.php
lathomas64: no, but if you publish to google play store then you need to include the value that google gives you here to include it in your assetlinks.json. I'll update the language and mark it as not required.

Required only for Google Play Store publishing. If you haven't published yet, leave this blank. You can retrieve this value from the Google Play Console after your initial submission. (https://support.google.com/googleplay/android-developer/answer/16641489?hl=en)

lathomas64:
the vapidKey was necessary for notifications they stopped working when I pulled that out but i updated the label to mention web push certificate as well hopefully that is more clear.

code/web/interface/themes/responsive/js/aspen/initFCM.js
```diff
@@ -0,0 +1,164 @@
...
+			{
+				//permission should already be granted before
+				//initialize is called
+				return;
+			}
+			appToken = currentToken;
+			let parser = new UAParser(window.navigator.userAgent);
+			let result = parser.getResult();
+			let modelName = result.device.vendor ? result.device.vendor + " " : "";
+			modelName += result.device.model ? result.device.model + " " : "";
```
JonahCWilson: Use your getModelName() function.

code/web/interface/themes/responsive/js/aspen/serviceWorker.js
```diff
@@ -0,0 +1,64 @@
...
+		// match the request to our cache
+		const cachedResponse = await cache.match(event.request);
+
+		// check if we got a valid response
+		if (cachedResponse !== undefined) {
+			// Cache hit, return the resource
+			return cachedResponse;
+		} else {
+			// Otherwise, go to the network
+			return fetch(event.request)
```
JonahCWilson: Nit: Unnecessary else block

code/web/services/API/SystemAPI.php
```diff
@@ -1217,6 +1232,28 @@ function getMaterialsRequestForm() : array {
...
+		require_once ROOT_DIR . '/sys/AspenPWA/Setting.php';
+		$settings = AspenPWASetting::getSettingsForCurrentLibrary();
+		if($settings)
+		{
+			return [
+				'success' => true,
+				'settings' => $settings->getFirebaseSettings(),
+			];
+		}
+		else
```
JonahCWilson: Nit: Unnecessary else block.

code/web/sys/AspenPWA/Setting.php
```diff
@@ -0,0 +1,296 @@
...
+
+	static function getSettingsForCurrentLibrary(): AspenPWASetting | null {
+		$settings = new AspenPWASetting();
+		$library = Library::getActiveLibrary();
+		$settings->id = $library->AspenPWASettingId;
+		if($settings->find(true))
+		{
+			return $settings;
+		}
+		else
```
JonahCWilson: Nit: Unnecessary else block

code/web/index.php
```diff
@@ -1269,6 +1269,40 @@ function loadModuleActionId() {
 	}catch (Exception $e) {
 		//This happens if web builder is not fully installed, ignore the error.
 	}
+	try {
+		if ($library->AspenPWASettingId != -1)
+		{
```
JonahCWilson, line 1308: `$_GET['module']` and `$_REQUEST['module']` are `AspenPWA` in every instance, can pull those out and set them once in outer scope.

lathomas64: i only want that set if it matches one of the URIs we are looking for. If I pull that out we would unintentionally overwrite the module value for other requests if PWA was turned on.

lathomas64: never mind previous comment I made an array of the routes and checked if the requestURI was a key in that array.

lathomas64:
@kylemhall I apologize for the ping I am testing something and needed to @ someone who wasn't already listed as a reviewer on this PR? please feel free to ignore.

code/web/index.php
JonahCWilson: This looks a lot nicer!

mdnoble73:
Let me know when this is ready for additional review. Thanks.

lathomas64:
> Let me know when this is ready for additional review. Thanks.

should be good now. Adressed previous comments and Jonah looked over and I've addressed his additional things.

LiYanjun19 (approved):
I tested the notifications tool, and it worked for me.

PR #3602, opened 2026-01-12
DIS-1827 index interval

lathomas64:
Aspen implements a soft-commit mechanism to make records searchable quickly, without putting too much pressure on I/O. This requires a hard commit request to make the changes persist on the hard drive. The current batch size for this hard commit actions is hardcoded.
The current value is good enough for most cases, but some sites might require different commit strategies depending on their usage.
We propose to make this configurable so each library/vendor can fine tune it to meet their needs.

to test:
Modify the indexCommitInterval variable in system variables to something other then the default of 10000, i.e. 20000 and run indexing. Indexing should run as normal.


mdnoble73 (changes requested):
@lathomas64 these changes look good, but you need a description and testing notes within Jira.

mdnoble73 (changes requested):
@lathomas64 please resolve conflicts on this pull request, thanks

mdnoble73 (approved):

PR #3617, opened 2026-01-14
DIS-1456: Added the ability for libraries and patrons to freeze holds immediately as they are created.

attheshow:
DIS-1456: Added the ability for libraries and patrons to freeze holds immediately as they are created.


attheshow:
Hmmm... This doesn't look right. Looks like it includes other commits. Will resubmit.

PR #3701, opened 2026-02-04
DIS-1888 Show system messages on all MyAccount pages

stephenrwicks:
Ensures system messages for "All Account Pages" show up on 
/MyAccount/Lists; 
/MyAccount/MyCampaigns; 
/Search/History
/MyAccount/MyEvents/

### DIFF
--- a/code/web/services/Search/History.php
+++ b/code/web/services/Search/History.php
@@ -18,6 +18,19 @@ class History extends Action {
             try {
                 $customAccountMessages = new SystemMessage();
                 $now = time();
+                global $action;
+                if ($action == 'CheckedOut') {
+                    $customAccountMessages->whereAdd("showOn = 1 OR showOn = 2");
+                } elseif ($action == 'Holds') {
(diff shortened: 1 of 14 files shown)


code/web/services/Search/History.php
```diff
@@ -9,6 +9,44 @@ class History extends Action {
...
+        parent::__construct($isStandalonePage);
+
+        //Load system messages
+        if (UserAccount::isLoggedIn()) {
+            $accountMessages = [];
+            try {
+                $customAccountMessages = new SystemMessage();
+                $now = time();
+                global $action;
+                if ($action == 'CheckedOut') {
```
mdnoble73: @stephenrwicks since we know we are on the History page, we can always set $customAccountMessages->showOn to 1 and skip the rest of the checks

mdnoble73 (changes requested):

code/web/services/Search/History.php
stephenrwicks: Got it

PR #3720, opened 2026-02-09
DIS-1923: Add logging for nightly reindexing

lucasmontoya13:
1. This patch adds debug logging to help track nightly reindexing processes by logging:
  - Possible triggers
  - Current values of relevant variables, such as runNightlyFullIndex, nightlyIndexTrigger, hasDbTriggers, and isManualCli
2. Debug logging can be enabled via the Log4j configuration.
3. Process name and PID are included in log messages for improved traceability.

Before applying the Patch

Run nightly reindex and observe the current log format:
cd /usr/local/aspen-discovery/code/reindexer
java -jar reindexer.jar <sitename> nightly
tail -20 /var/log/aspen-discovery/<sitename>/logs/grouped_reindex.log

Current behavior:
- Log format: [main] INFO 02/04 13:00:00,000 - message
- No process identification in logs
- No debug messages for nightly trigger detection

After applying the patch:

Configure logging:
- If config exists: Edit sites/<sitename>/conf/log4j.grouped_reindex.properties and update the pattern to [grouped_reindex:%pid]
- If config doesn't exist: It will auto-generate with the new format on first run

Enable debug logging (to see trigger messages):
appender.rolling.filter.threshold.level = debug
rootLogger.level = debug

Run nightly reindex:
java -jar reindexer.jar <sitename> nightly

Verify output:
tail -30 /var/log/aspen-discovery/<sitename>/logs/grouped_reindex.log


mdnoble73 (commented):
@lucasmontoya13 why is this logging to the .log file rather than putting it in the database log?

lucasmontoya13:
I added the logging messages at debug level in the file log to avoid noise. I could add them to the database log if you consider it possible

mdnoble73 (approved):

PR #3734, opened 2026-02-10
DIS-1551: Add Barcode to ce Admin View

AlexanderBlanchardAC:
https://aspen-discovery.atlassian.net/browse/DIS-1551


mdnoble73 (changes requested):
@AlexanderBlanchardAC this should be refactored to pass the search term to AJAX and then return the matches rather than doing the filtering within javascript. It also should ensure the user is eligible for community engagement (exclude local admins and admin users, probably anything without an ILS connection?)

AlexanderBlanchardAC:
> @AlexanderBlanchardAC this should be refactored to pass the search term to AJAX and then return the matches rather than doing the filtering within javascript. It also should ensure the user is eligible for community engagement (exclude local admins and admin users, probably anything without an ILS connection?)

I have changed this to search using ajax. Admin, including local admin are given campaigns during testing so I have not filtered them out for now.

code/web/services/CommunityEngagement/AJAX.php
```diff
@@ -1539,4 +1540,39 @@ function extraCreditRewardGivenUpdate() {
 		}
 		exit;
 	}
+
+	public function searchUsers() {
```
mdnoble73, line 1471: @AlexanderBlanchardAC this needs to be restricted by permission

code/web/interface/themes/responsive/CommunityEngagement/adminView.tpl
```diff
@@ -23,7 +23,7 @@
 						<select id="user_id" class="form-control-sm" style="margin-bottom: 3px;" onchange="AspenDiscovery.CommunityEngagement.filterDropdownOptions('user')">
 							<option value="">-</option>
 							{foreach from=$users item=$user}
-									<option value="{$user.id}">{$user.displayName}</option>
+								<option value="{$user.id}">{$user.displayName}{if $user.ils_barcode} - {$user.ils_barcode}{/if}</option>
```
mdnoble73, line 26: @AlexanderBlanchardAC I'm not seeing a way to search from this. It looks like it's still just displaying a list of all users.

mdnoble73 (changes requested):

code/web/interface/themes/responsive/CommunityEngagement/adminView.tpl
Chloe070196: Tested this and initially found the same. The answer is in the test plan on Jira, it seems (which I also only noted later)

_If you want to swap between the search bar and the dropdown for testing purposes, you can do this via Admin-> Primary Configuration-> Library Systems-> Community Engagement → Admin View User Select._ 

Having done so, I was presented with a search bar instead of a dropdown and able to find users by their ILS barcode.

mdnoble73:
Closing because this does not merge cleanly. Will replicate and merge manually.

PR #3735, opened 2026-02-10
DIS-1570: Allow Removal of Campaigns

AlexanderBlanchardAC:
https://aspen-discovery.atlassian.net/browse/DIS-1570


code/web/services/CommunityEngagement/AJAX.php
```diff
@@ -369,7 +374,89 @@ function filterCampaigns() {
...
+		require_once ROOT_DIR . '/sys/CommunityEngagement/UserRemovedCampaign.php';
+
+		$removed = new UserRemovedCampaign();
+		$removed->campaignId = $campaignId;
+		$removed->userId = $userId;
+		
+		return $removed->find(true);
+	}
 	
+	public function restoreCampaignForUser() {
```
mdnoble73, line 246: This should check permissions. Also should not need to override how the json is built and formatted.

code/web/services/CommunityEngagement/AJAX.php
```diff
@@ -148,6 +148,7 @@ function filterCampaigns() {
 					$html .= "<button class='btn btn-primary btn-sm' onclick='AspenDiscovery.CommunityEngagement.refreshCurrentUserStats($userId); return false;' style='margin: 5px 0;'>Refresh Campaign Progress</button>";
 
 					foreach ($allEligibleCampaigns as $campaign) {
+						$isRemoved = $this->userRemovedCampaignCheck($campaign->id, $userId);
```
mdnoble73, line 174: This HTML should be built with a template and should be translatable

code/web/services/CommunityEngagement/AJAX.php
```diff
@@ -369,7 +374,89 @@ function filterCampaigns() {
...
 	}
+
+	private function userRemovedCampaignCheck($campaignId, $userId) {
+		require_once ROOT_DIR . '/sys/CommunityEngagement/UserRemovedCampaign.php';
+
+		$removed = new UserRemovedCampaign();
+		$removed->campaignId = $campaignId;
+		$removed->userId = $userId;
+		
+		return $removed->find(true);
```
mdnoble73, line 243: This could use ->count() > 0 which would be slightly faster (not a big deal either way though). Should also prevent this from being called directly from URL.

code/web/services/MyAccount/AJAX.php
```diff
@@ -11836,4 +11836,113 @@ function deleteListGroup(): array {
...
+			'modalBody' => $interface->fetch('MyAccount/remove-campaign-modal.tpl'),
+			'modalButtons' => "<button class='tool btn btn-primary' onclick='AspenDiscovery.Account.removeCampaignFromUI($campaignId, $userId)'>" . translate([
+				'text' => 'Remove',
+				'isAdminFacing' => 'true',
+			]) . "</button>",
+		];
+	}
+
+	public function removeCampaignFromUI() {
+		$userId = $_REQUEST['userId'] ?? null;
```
mdnoble73, line 11956: We should ensure this is not called directly by URL, probably better to move it to the User Object

code/web/services/MyAccount/AJAX.php
```diff
@@ -11836,4 +11836,113 @@ function deleteListGroup(): array {
 		return $result;
 
 	}
+
+		public function removeCampaignModal() {
+		$userId = $_REQUEST['userId'] ?? null;
```
mdnoble73: This should ensure the passed user id is for the active user or use the active user id.

mdnoble73 (changes requested):

code/web/services/CommunityEngagement/AJAX.php
AlexanderBlanchardAC: Should be okay as a private function?

code/web/services/CommunityEngagement/AJAX.php
AlexanderBlanchardAC: Used template

code/web/services/MyAccount/AJAX.php
AlexanderBlanchardAC: Have left this as an ajax endpoint but added permission checks for safety.

code/web/services/MyAccount/AJAX.php
AlexanderBlanchardAC: added check

code/web/services/CommunityEngagement/AJAX.php
mdnoble73: Private functions can still be called from AJAX by specifying the method.

Chloe070196: Tested Alexander's suggestion: the AJAX does indeed `return success: false` on direct URL access attempt due to the private scope.

Is the change still required then?

mdnoble73 (approved):

mdnoble73:
Merged manually in https://github.com/Aspen-Discovery/aspen-discovery/pull/3889 because this had conflicts that could not be easily merged

PR #3746, opened 2026-02-11
DIS-1430 Implement Aspen Plugin System

Jacobomara901:
Obviously, this is not ready but this is where I am up to with implementing the plugin system currently.
@mdnoble73, Alex said you were interested in taking a look so I thought it best to throw this in a draft PR for now. It's a bit of a passion project but I've not been able to work on it as much as I would like recently! As Alex said, I should be able to get back to it properly after the events module updates have been pushed out of the door!

Definitely things I want to change here but your input is, as always, valued.
I want to rewrite the permissions, I don't like how they currently work. I've recently removed the plugin_methods table from the db and had php query has_method instead as at the scale we'll be using it, to have the methods in the db would be over-engineering I think. I've tried to implement a nice class system for adding hooks in the future, though currently there are not many implemented. 
I've implemented plugin configuration storage in the db, so you can configure plugins once installed, they follow a .plugzip extension for install currently (just a zip file) If you would like an example plugin to install I've pushed one here: https://github.com/Jacobomara901/aspen-example-plugin

It's turned on as a module but then you'll have to add the new plugin permissions for opacadmin (as I haven't added them by default yet)
since you're not using adb you'll also need to add the apache change for plugin file serving.

You'll notice a whole host of messages for warnings/alerts that "plugins can run code" and the like on the plugin page/ install page. I'm not sure which I liked and wanted to keep, so I've got them all there for now until I decide which makes the most sense!

Let me know what you think. As I said, I know it's still a little rough around the edges which is why you didn't yet have a draft PR! But the main functionality is there.

Just tagging a couple of people that might want to keep an eye on the above:
@AlexanderBlanchardAC, @Chloe070196 and I think you seemed interested at AspenCon: @tnlibrarian


mdnoble73:
Closing this post Plugin Meeting. New plugin starting point is available now.

PR #3747, opened 2026-02-11
DIS-1439 Aspen Api Improvements

Jacobomara901:
@mdnoble73 Another big one from me, I'm afraid. Apologies!

Some of this may change slightly but the general structure will remain the same (might swap to guard clauses etc)

First off, this is fully backwards compatible and will NOT break any existing API endpoints. I would like to eventually migrate the old api endpoint to use this new structure but it's important everything works in the meantime!

First set of patches adds oauth keys to Aspen. These are permission locked for the legacy endpoints and require you to have the superuser Administer all APIs permission for it to work. other than that, the structure is the same for legacy endpoints.
You can also disable the ability to create oauth tokens on a per role basis.

The second half of the patchset is the permissions work for apis. This rewrites how api's are structured and will instead use a different launch method. it has different scopes that define how the api endpoint can be authenticated (be it via greenhouse keys, ip, oauth) and these can be stored in an OR array in the openapi.json, so that becomes the source of truth.
You can then add permissions in there as well which will link by name to permissions that exist in the aspen db, so that a user who has the permission to do something in the UI with then have access in the api as well.
I've added a test endpoint there for you to have a play around with, it's in a do not push commit so it doesn't get accidentally merged!

Here is a md file with the general workflow:
[api-workflow.md](https://github.com/user-attachments/files/25233215/api-workflow.md)

The reasoning for this work is we need to be able to share an endpoint with LibraryOn for the events work and currently there was no secure way of doing this. It's part of a funded piece of work so we will be implementing this on our customer servers for testing within the next week or so, prior to the upstream here. That should ensure there are no edge cases or "nasties" that I've missed. Ideally, I'd like to extend this to properly use bearer tokens in the future and have expiration periods on the generated oauth tokens but this is a good foundational start that should be able to be extended later.

Again, tagging a few people that might want to stay updated:
@Chloe070196, @AlexanderBlanchardAC, @mrenvoize

### DIFF
--- a/code/web/services/API/AbstractAPI.php
+++ b/code/web/services/API/AbstractAPI.php
@@ -176,4 +176,26 @@ abstract class AbstractAPI extends Action{
 			}
 		}
 	}
+
+	protected function getAuthenticatedUserForOpenAPI() {
+		global $oAuthUser;
+		if (isset($oAuthUser) && $oAuthUser !== false) {
+			if ($oAuthUser->source == 'admin') {
+				return false;
+			}
+			return $oAuthUser;
+		}
+		
+		if (isset($_SERVER['PHP_AUTH_USER'])) {
+			if ($this->grantTokenAccess()) {
+				global $oAuthUser;
+				if ($oAuthUser->source == 'admin') {
+					return false;
+				}
+				return $oAuthUser;
+			}
+		}
+		
+		return false;
+	}
 }
--- a/code/web/services/API/AbstractAPI.php
+++ b/code/web/services/API/AbstractAPI.php
@@ -12,6 +12,9 @@ abstract class AbstractAPI extends Action{
 		if ($this->checkIfLiDA()) {
 			$this->context = 'lida';
 		}
+		if (empty($this->apiName)) {
+			$this->apiName = (new ReflectionClass($this))->getShortName();
+		}
 	}
 
 	function checkIfLiDA(): bool {
--- a/code/web/services/API/AbstractAPI.php
+++ b/code/web/services/API/AbstractAPI.php
@@ -121,9 +121,7 @@ abstract class AbstractAPI extends Action{
 
 		global $oAuthUser;
 		if (isset($oAuthUser) && $oAuthUser !== false) {
-			if ($oAuthUser->source == 'admin') {
-				return false;
-			}
+			// Note: 'Use All API Endpoints' permission is checked in grantTokenAccess()
 
 			if (empty($_REQUEST['language'])) {
 				global $activeLanguage;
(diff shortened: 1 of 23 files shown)


mdnoble73:
@Jacobomara901 what is the reason for attaching the oAuth keys to users? I think it would be much better to manage all of the keys centrally. That way there isn't a risk of having an oAuth key be "lost" when staff leave.

mdnoble73:
Jira ticket needs to be updated with a description as well as a test plan. What APIs is LibraryOn intending to call?

sites/default/httpd-default.conf
```diff
@@ -12,6 +12,7 @@
       Order allow,deny
       allow from all
       AllowOverride All
+      CGIPassAuth On
```
mdnoble73, line 15: @Jacobomara901 what is this needed for?

Jacobomara901:
Hiya! Sorry for no reply last week, I was abroad!
Q1: The idea was to be able to get the authentication obtained from the auth step to eventually allow you to act upon endpoints as your own user. I.E. If someone wanted to view their own Events for example through the api, they could do that. It's also honestly more similar to the api systems I've worked with previously though I'm open to discussion if you disagree with the thoughts here. My inclination was that if the permission system is fleshed out well, we wouldn't have an issue with a patron interacting with their own data.
In terms of having api keys hanging if a staff member leaves etc, I agree, that would be bad! My vision for the end goal here would be to have keys that expire after a period anyways, additionally I think we should be adding a removal for whenever staff are deleted from the db. Either through adding in php or by implementing some foreign key constraints here.

Q2: No worries, in that case, I shall assign the ticket to myself today and add some testing plans etc there. I was only not doing so as this was still in draft and I did not want to step on Kirstiens toes. I'll reach out to her again on the Jira ticket as I'm more than happy to collaborate on the above!
In terms of the api they will be interacting with, they are harvesting a new events api as part of the Registration work, it just gives information of upcoming events and their instances. It's worth noting that LibraryOn have been using the above work for around a month now with no issues (as the go live was a hard deadline)

Q3: Apache doesn't pass auth headers through to PHP-FPM by default. Since both the production docker and the dev box use PHP-FPM, CGIPassAuth is required! Shouldn't make a difference for other installs though, let me know if this is not the case in your testing.

code/web/services/API/AbstractAPI.php
```diff
@@ -72,6 +79,18 @@ function getLiDAUserAgent() {
 		return false;
 	}
 
+	protected function extractBasicAuthCredentials(): void {
+		if (!isset($_SERVER['PHP_AUTH_USER']) && (isset($_SERVER['HTTP_AUTHORIZATION']) || isset($_SERVER['REDIRECT_HTTP_AUTHORIZATION']))) {
```
JonahCWilson, line 83: Nit:

Assign the result of this boolean to a variable and use that instead.
Improves readability for the next person to just see
`if ($hasServerAuthorization)` or similar.

Jacobomara901: You are 100% completely correct here. There's also a couple places I want to swap to guard clauses etc. Will be amended for when this is no longer in draft. Thanks for taking a look, please do continue to point out areas like this that need fixing. I appreciate the input.

mdnoble73:
Closing as we jointly work on DIS-2214

PR #3777, opened 2026-02-17
DIS-1997 Update GitHub actions for pull requests to verify all commits start with the correct commit message

kylemhall:


mdnoble73 (changes requested):
This needs to be submitted against 26.03 and needs release notes.

mdnoble73:
The Technical Committee has decided to wait on this and evaluate how much of a problem remains after telling developers to add the DIS number to commits. The feeling was that many developers currently ignore github action warnings already and that there is not an easy way to change a commit name after it has already been submitted. We're also trying to keep the barriers to having code accepted as low as possible. We'll re-evaluate in a month or two to see if we'd like to include this. Thanks for submitting.

mdnoble73:
Closing per discussion with Yanjun

PR #3779, opened 2026-02-18
DIS-1998: Add jump to functionality to results page

JonahCWilson:
Partner explained they'll have multiple people working on a project, and they need functionality to go to page 26/113 without continuously clicking next.

They were given a workaround to update the link, but this functionality is a one-line change, so easy enough to implement.


JonahCWilson:
Realized this was more than I wanted, will update and point towards 26.03. 

Marking as draft in the interim.

mdnoble73:
@JonahCWilson can you resolve conflicts on this? There is also a checkbox to allow reviewers to make changes. If you check that I can correct these simple issues for you.

LiYanjun19:
@JonahCWilson It might be better if we apply this for any set of results like events, lists, etc, for consistency

mdnoble73 (changes requested):
Agree with Yanjun, let's do this for all results pages, not just Catalog Searches.

JonahCWilson:
Okay, I changed the default Pager to use canJumpToPage = true and deleted the call from everywhere it's used.

mdnoble73 (approved):

PR #3817, opened 2026-02-23
DIS-1652 allow indefinite hold freezes

lathomas64:
to test
-  go to /API/SystemAPI?method=getLibraryInfo&id=1 (replace the id with a valid library id on your system)
-  you should see allowIndefiniteHoldFreezes with a value of 0
-  run the database maintenance scripts attached to the Pull Request
-  In the general settings for lida that include the library you used before check the box for Allow hold freezes in LiDA without an end date
-  go to the getLibraryInfo api call again and this time allowIndefiniteHoldFreezes should have a value of 1


mdnoble73 (changes requested):
This shouldn't be a special setting for LiDA. We should utilize information already available within library information.

lathomas64:
was not aware of reactivateDateNotRequired, switching to use that instead. Thank you for the redirect!

mdnoble73 (approved):

mdnoble73:
@lathomas64 can you resolve conflicts on this?

mdnoble73:
Needs conflicts resolved. The new parameters also don't match the expected value in LiDA now.

lathomas64:
I modified the LiDA pull request too to match the changes I made here.

lathomas64:
okay looks like i made the changes but they didn't get pushed and I've got to do those over again thanks.

code/web/release_notes/26.03.00.MD
```diff
@@ -106,6 +106,11 @@
 ### Polaris Updates
 - Fix error when running a full update from Polaris. ([DIS-2053](https://aspen-discovery.atlassian.net/browse/DIS-2053)) (*MDN*)
 
+### Community Engagement Updates
```
mdnoble73: @lathomas64 it looks like these were added in error, can you remove and resubmit?

mdnoble73 (changes requested):

code/web/release_notes/26.03.00.MD
lathomas64: sure thing!

mdnoble73 (approved):

PR #3830, opened 2026-02-24
DIS-2027 - ItemAPI refactor

JonahCWilson:
Minor refactoring to bring marcFields into one place.  Easier to maintain should the logic change in the future.


JonahCWilson:
I'll update with release notes once another PR goes in.

mdnoble73 (changes requested):
The getItem and getBasicItem APIs intentionally return different information and there is a performance cost to loading additional fields. The fields specified are not actually MARC fields either. 

This likely just needs to be closed.

PR #3850, opened 2026-02-27
DIS-2045: InvoiceCloud cc service fee

LiYanjun19:
Currently in InvoiceCloud settings, Credit Card Service Fee can only be set as a fixed amount (e.g. $1.50).
Aspen should also allow the Credit Card Service Fee to be configured as a percentage (e.g. 1.5%). 

To test:
1. Apply PR.
2. In InvoiceCloud Settings, set Credit Card Service Fee to 1.5% and enable Force Debugging Logs 
3. Attempt to pay the fine
4. Check the External Request Log in Greenhouse, and see that CCServiceFee and ACHServiceFee are 1.5% of the total payment. 
5. Set Credit Card Service Fee to a fixed amount 1.50.
6. Attempt to pay the fine, check the log again and see CCServiceFee and ACHServiceFee are $1.50

### DIFF
--- a/code/web/release_notes/26.03.00.MD
+++ b/code/web/release_notes/26.03.00.MD
@@ -54,6 +54,9 @@
 ### Other Updates
 - Fix "Not Holdable" label missing for non-holdable items in Copies table. ([DIS-1973](https://aspen-discovery.atlassian.net/browse/DIS-1973)) (*YL*)
 
+### eCommerce Updates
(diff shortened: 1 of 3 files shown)


mdnoble73 (changes requested):
This could use some additional documentation inline as a note to explain how to apply the percentage and also clarify if they should add a $ symbol when applying a fixed fee service fee.

mdnoble73 (approved):

PR #3851, opened 2026-02-27
DIS-2046: Allow null callnumbers in getSortableDate

kidclamp:
TO test:
1 - Find a Continuing resource -  The Entomologist's monthly magazine in KTD
2 - Edit the record and remove all the callnumber fields (052, 092, etc. anything that looks like a callnumber) 
3 - Add some items of type continuing resource with no callnumber 
4 - In Aspen/ADB set the format for Continuing Resource to ‘Magazine' 
5 - Do a full reindex
6 - Load the record and see the error
7 - Apply patch
8 - Reload
9 - No error!


mdnoble73 (changes requested):
The remaining code in getSortableDate will not work with a null so there either needs to be a catch to see if $str is null or the calling function should check if $str is null before calling getSortableDate

mdnoble73 (changes requested):
@kidclamp I did some more testing on this and your change looks ok. Can you resolve changes and I'll merge into 26.03.

mdnoble73 (approved):

PR #3853, opened 2026-02-28
DIS-1986: Fix batch all hold actions for linked accounts

catsoup11789:
Hold actions now need to be explicitly called by the linked account user instead of the primary user.


code/web/services/MyAccount/AJAX.php
```diff
@@ -915,6 +915,11 @@ function cancelAllHolds() {
 				$cancelId = $hold->cancelId;
 				$holdType = $hold->source;
 				$isIll = $hold->isIll;
+				if ($hold->userId != $user->id) {
+					if ($user->getUserReferredTo($hold->userId)) {
+						$user = $user->getUserReferredTo($hold->userId);
+					}
+				}
```
JonahCWilson:
```suggestion
			if ($hold->userId != $user->id) {
					$user = $user->getUserReferredTo($hold->userId) ?? $user;
			}
```

catsoup11789: we don't need to reapply the user to user? it's already set to that?

JonahCWilson: True, but it's also doesn't change the behavior and costs nothing.

The logic, as-written, calls getUserReferredTo twice. We call it once to check if it exists, and if so, we assign that value to user.  If not, user's value doesn't change.

If that function call is computationally expensive, it can add up over time.

The suggested rewrite calls it once, and uses the ?? to determine if it's null. If not, we use the value, if so, we keep user as user, so the behavior doesn't change.

catsoup11789: I fully understand how the nullish coalescing operator works, thanks.

code/web/services/MyAccount/AJAX.php
```diff
@@ -915,6 +915,11 @@ function cancelAllHolds() {
 				$cancelId = $hold->cancelId;
 				$holdType = $hold->source;
 				$isIll = $hold->isIll;
+				if ($hold->userId != $user->id) {
+					if ($user->getUserReferredTo($hold->userId)) {
+						$user = $user->getUserReferredTo($hold->userId);
```
mdnoble73: I'm not overly worried about the extra call. We should use a temporary variable though rather than $user since overwriting that will result in problems if there are more than 1 linked users or if the the linked user doesn't link back to the original user.

mdnoble73 (approved):

PR #3865, opened 2026-03-02
DIS-2055: In-place refactoring

JonahCWilson:
While working on a bug, I encountered this function, this is an in-place rewrite to make it easier to follow.

Ordering is done by function, can easily be reordered, extended, or reduced as-needed.

Pulled out extra logic into helper functions.


mdnoble73 (changes requested):
@JonahCWilson this needs testing notes in Jira to be accepted. 

The technical committee strongly recommends not doing refactoring for the sake of refactoring. Although there are definitely reasons to make a code more efficient, there is a cost in terms code review and testing that are non-negligable. 

If desired, please resubmit this to 26.04 and remove from draft status.

PR #3866, opened 2026-03-02
DIS-1968: Configure Symphony Bill Reason Display

K-Alette:
Add translation map for symphony bill reasons

Update release notes

Tested on my local instance of aspen

### DIFF
--- a/code/web/Drivers/SirsiDynixROA.php
+++ b/code/web/Drivers/SirsiDynixROA.php
@@ -2247,6 +2247,14 @@ class SirsiDynixROA extends AbstractIlsDriver {
 			$totalFinesOwed = 0;
 
 			if (!empty($blockList->fields->blockList)) {
+				$indexingProfileId = $this->getIndexingProfile()->id;
+				require_once ROOT_DIR . '/sys/Indexing/TranslationMap.php';
+				$billReasonTranslation = new TranslationMap();
+				$billReasonTranslation->name = "bill_reason";
+				$billReasonTranslation->indexingProfileId = $indexingProfileId;
+				$billReasonTranslation->find();
+				$billReasonTranslation->fetch();
+
 				foreach ($blockList->fields->blockList as $block) {
 					$fine = $block->fields;
 					$title = '';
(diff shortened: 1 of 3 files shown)


code/web/Drivers/SirsiDynixROA.php
```diff
@@ -2247,6 +2247,14 @@ public function getFines($patron, $includeMessages = false): array {
 			$totalFinesOwed = 0;
 
 			if (!empty($blockList->fields->blockList)) {
+				$indexingProfileId = $this->getIndexingProfile()->id;
+				require_once ROOT_DIR . '/sys/Indexing/TranslationMap.php';
+				$billReasonTranslation = new TranslationMap();
+				$billReasonTranslation->name = "bill_reason";
+				$billReasonTranslation->indexingProfileId = $indexingProfileId;
+				$billReasonTranslation->find();
```
mdnoble73: @K-Alette this can be simplified to `$billReasonTranslation->find(true);` rather than having the find and fetch on separate lines. We should also store if the translation map was found.

code/web/Drivers/SirsiDynixROA.php
```diff
@@ -2261,11 +2269,12 @@ public function getFines($patron, $includeMessages = false): array {
 							$barcode = $fine->item->fields->barcode;
 						}
 					}
+					$translatedValue = $billReasonTranslation->translate($fine->block->key);
```
mdnoble73: @K-Alette let's check here that the translation map is found (just in case someone deletes it) and if it wasn't found, we can just use the original value.

code/web/sys/DBMaintenance/version_updates/26.03.00.php
```diff
@@ -104,3 +112,23 @@ function getUpdates26_03_00(): array {
...
 	];
 }
+function addBillReasonTranslationMap(&$update): void {
+	$ils = null;
+	require_once ROOT_DIR . '/sys/Account/AccountProfile.php';
+	$accountProfiles = new AccountProfile();
+	$accountProfiles->find();
+	while ($accountProfiles->fetch()) {
+		if ($accountProfiles->ils != 'na') {
+			$ils = $accountProfiles->ils;
```
mdnoble73: Let's check here if the ils is symphony and then we can use 
$accountProfiles->getIndexingProfile()
to load the indexing profile.

mdnoble73 (changes requested):
See inline comments

mdnoble73 (approved):

PR #3887, opened 2026-03-04
DIS-1960 Update reading history cronjob should access local API locally

kylemhall:
UpdateReadingHistoryTask makes an API call to updatePatronReadingHistory via the Aspen URL. In cases where a server is not exposed directly to the Internet, this causes the request to leave the network and return which is less efficient and makes the request subject to any WAF that may be in front of the server. The request should access localhost and set the host name in the request accordingly so apache will handle it correctly. This will prevent the request from leaving the server itself!

Test Plan

1) Apply this patch
2) Rebuild the cron jar
3) Run the UpdateReadingHistory cron job against a server with reading history users
4) Confirm reading history updates complete successfully (check cron log for updated counts, no new errors)
5) Confirm the requests no longer leave the server: each Aspen site has its own Apache access log (/var/log/aspen-discovery/{sitename}/access.log). Tail the log for the site being updated and note the updatePatronReadingHistory requests now come from 127.0.0.1 (or ::1) with status 200
6) On a multi-tenant server, run the cron for each tenant and confirm the requests appear in the access log of the *correct* site (not the default/first virtual host), and that reading history is updated for the right users on each tenant
7) Confirm reading history counts match a run from before the patch (no users newly failing)

### DIFF
--- a/code/cron/src/com/turning_leaf_technologies/cron/Cron.java
+++ b/code/cron/src/com/turning_leaf_technologies/cron/Cron.java
@@ -23,6 +23,12 @@ public class Cron {
 	 * @param args command line parameters passed
 	 */
 	public static void main(String[] args) {
+		// Java refuses to send the Host header set via setRequestProperty unless restricted headers are allowed.
+		// This must be set before the first HttpURLConnection is loaded, so it must happen here rather than in
+		// the process that needs it. UpdateReadingHistoryTask uses the Host header to route requests made via
+		// localhost to the correct virtual host.
+		System.setProperty("sun.net.http.allowRestrictedHeaders", "true");
+
 		String serverName;
 		if (args.length == 0) {
 			serverName = AspenStringUtils.getInputFromCommandLine("Please enter the server name");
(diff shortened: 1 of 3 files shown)


lathomas64:
ran into some difficulty trying to build the jar myself:
Unable to find image 'openjdk:11' locally
docker: Error response from daemon: manifest for openjdk:11 not found: manifest unknown: manifest unknown

will look at this again once i get past that hurdle.

lathomas64:
everything still successfully updates reading history for me with this. looks good.

mdnoble73:
Needs to be targeted to 26.05. Thanks

mdnoble73 (changes requested):
This does not work for my local configuration (Windows running XAMPP in a multi-tenant setup). The Host parameter is not being read by the webserver. 

Needs Jira updated properly with testing notes, status, etc. Testing should include multi-tenant systems and consortial systems with multiple URLs.

kylemhall:
> This does not work for my local configuration (Windows running XAMPP in a multi-tenant setup). The Host parameter is not being read by the webserver.

Ack! Can you provide a copy of your apache configs and the contents of:
aspenUrl
new URL(aspenUrl).getHost();
 
> Needs Jira updated properly with testing notes, status, etc. Testing should include multi-tenant systems and consortial systems with multiple URLs.

Can do! Let me know what you get from the output above and I'll see what I can do!

mdnoble73 (approved):

PR #3909, opened 2026-03-08
DIS-2055 related records refactor

JonahCWilson:


JonahCWilson:
As discussed, rebased off 26.04, updated Jira to include testing instructions.[Jira-2055](https://aspen-discovery.atlassian.net/browse/DIS-2055)

code/web/RecordDrivers/GroupedWorkDriver.php
```diff
@@ -300,106 +300,70 @@ static function compareLocalItemsForRecords(Grouping_Record $a, Grouping_Record
...
 	 * @return int
 	 */
-	function compareRelatedRecords(Grouping_Record $a, Grouping_Record $b) : int {
-		//Get literary form to determine if we should compare editions
-		$literaryForm = '';
-		if (isset($this->fields['literary_form'])) {
-			if (is_array($this->fields['literary_form'])) {
-				$literaryForm = reset($this->fields['literary_form']);
-			} else {
-				$literaryForm = $this->fields['literary_form'];
```
JonahCWilson, line 310: Just commenting here for my own edification to make sure I understand the code. Please let me know if I'm making a mistake and where.

$literaryForm is set to the empty string.

If it NOT the case that 'literary_form' is set in $this->fields, then $literaryForm remains the empty string.

If it IS the case that it is set, we then check:
- If it IS the case that $this->fields['literary_form'] is an array, then we set $literaryForm to the value of reset($this->fields['literary_form'). 
- If it is NOT the case that $this->fields['literary_form'] is an array, then we set $literaryForm = $this->fields['literary_form'].

In summary, the variable $literaryForm, at the end of this block of code, can be one of the following three values:
- ''
- reset($this->fields['literary_form'])
- $this->fields['literary_form']

----

The proposed refactor:
```
private function getPrimaryLiteraryForm(): string {
        $value = $this->fields['literary_form'];
		if (!isset($this->fields['literary_form'])) {
			return '';
		}
		return is_array($value) ? reset($value) : $value;
}
```
If it is NOT the case that $this->fields['literary_form'] is set, then we return the empty string.

If that if block is not hit, we know it is set, and we ask "Is this an array?"  If so, we return reset($value) else just the value itself.

At the end of this block of code, one of three values will be returned:
- ''
- reset($this->fields['literary_form'])
- $this->fields['literary_form']

My understanding is that this code and the refactor are functionally identical. Please correct me if I'm wrong.

code/web/RecordDrivers/GroupedWorkDriver.php
```diff
@@ -300,106 +300,70 @@ static function compareLocalItemsForRecords(Grouping_Record $a, Grouping_Record
...
-				return 1;
-			}
+
+		return 0;
+	}
+
+	private function getPrimaryLiteraryForm(): string {
+		if (!isset($this->fields['literary_form'])) {
+			return '';
 		}
```
JonahCWilson, line 324: My understanding is that we assign the values for $format1, $format2, and $formatComparison.

If $formatComparison is equal to 0, we skip the block and start doing other checks.  If it's not equal to 0, we do 1 of three things:
- Return -1
- Return 1
- Skip and do nothing if neither format is equal to 'Book'
   - In this case, and the ==0 case, $formatComparison retains its value.


My proposed refactor:
```
private function compareFormats(string $format1, string $format2): int {
		$comparison = strcasecmp($format1, $format2);

		if ($comparison !== 0) {
			if ($format1 === 'Book') return -1;
			if ($format2 === 'Book') return 1;
		}

		return $comparison;
	}
```

Receives the $format1 and $format2 parameters. It assigns the $formatComparison function. If the comparison is equal to 0, return the $comparison value.
If not, check whether each is a book and do one of the following:
- return -1
- return 1
- skip the block and return the unchanged $comparison value.

It is my understanding that these two pieces of code are functionally identical.  Please let me know if/where I'm wrong.

code/web/RecordDrivers/GroupedWorkDriver.php
```diff
@@ -300,106 +300,70 @@ static function compareLocalItemsForRecords(Grouping_Record $a, Grouping_Record
...
 
-		global $library;
-		$groupedWorkDisplaySettings = $library->getGroupedWorkDisplaySettings();
```
JonahCWilson, line 327: My understanding of this code is that we assign values to some function call, and for each case, we check if it's 0.  Another way to potentially write this is:

```
$val = functionCall()
if(!$val == 0) {
   return $val;
}

$val = nextFunctionCall()
if(!$val == 0){
   return $val
}
```

We could do that for every check, and at the end it would return 0 if none of them return early.

Since the above code is repetitive, we could abstract out the conditions into a list, and iterate over each one.

It is my understanding that the proposed refactor is functionally identical.

Please let me know if/where I'm wrong so I can correct it.

LiYanjun19 (approved):

code/web/RecordDrivers/GroupedWorkDriver.php
mdnoble73: In your proposed refactor, the first line will throw an error if $this->fields['literary_form'] is not set. 
If rewriting you could do: 
private function getPrimaryLiteraryForm(): string {
		if (!isset($this->fields['literary_form'])) {
			return '';
		}
		return is_array($this->fields['literary_form']) ? reset($this->fields['literary_form']) : $this->fields['literary_form'];
}

or 
private function getPrimaryLiteraryForm(): string {
		if (!isset($this->fields['literary_form'])) {
			return '';
		}
                $value = $this->fields['literary_form'];
		return is_array($value) ? reset($value) : $value;
}

code/web/RecordDrivers/GroupedWorkDriver.php
mdnoble73: They are the same, I don't see one or the other being especially easier or more difficult to read

code/web/RecordDrivers/GroupedWorkDriver.php
mdnoble73: if(!$val == 0){

does not feel correct since we don't really want to test the negation of $val. 
if ($val !== 0) {

would be correct. 

With the values provided both would ultimately do the same, but the first is not semantically correct

code/web/RecordDrivers/GroupedWorkDriver.php
JonahCWilson: My apologies, when free-writing this I actually didn't even copy + paste my own code.  The refactor, as-written, is the same as your second suggestion.

```
private function getPrimaryLiteraryForm(): string {
	if (!isset($this->fields['literary_form'])) {
		return '';
	}

	$value = $this->fields['literary_form'];

	return is_array($value) ? reset($value) : $value;
}
```

code/web/RecordDrivers/GroupedWorkDriver.php
JonahCWilson: The above is just pseudocode to illustrate another way to write the nested-ifs.  The actual code as written does use the !== check in the for-loop.

JonahCWilson:
@mdnoble73 I went ahead and merged 26.04 + added in your newest checks for statusRanking in the list + updated comments.  This should be good to go!

mdnoble73 (approved):

PR #3911, opened 2026-03-09
DIS-2146: fix: add login and permission checks

Chloe070196:


code/web/services/Events/AJAX.php
```diff
@@ -487,6 +526,26 @@ function checkEventsForType() : array {
 
 	/** @noinspection PhpUnused */
 	function saveEventsForType() : array {
+		if(!UserAccount::isLoggedIn()) {
+			return[
+				'success' => false,
+				'message' => translate([
+					'text' => "Access unauthorized.",
+					'isAdminFacing' => true,
+				])
```
JonahCWilson, line 535: All of these only differ by the error message.

Perhaps have a helper function similar to

```
private function getEventsAuthorizationError($msg) {
     return [
				'success' => false,
				'message' => translate([
					'text' => $msg,
					'isAdminFacing' => true,
				])
			];
}
```

catsoup11789: > All of these only differ by the error message.
> 
> Perhaps have a helper function similar to
> 
> ```
> private function getEventsAuthorizationError($msg) {
>      return [
> 				'success' => false,
> 				'message' => translate([
> 					'text' => $msg,
> 					'isAdminFacing' => true,
> 				])
> 			];
> }
> ```

It might make more sense to put a function like that in `JSON_Action.php` so its accessible to all AJAX files and only needs modified in one place.

mdnoble73 (approved):

PR #3917, opened 2026-03-11
DIS-281: Sierra x Aspen Holidays (QA Update) 

K-Alette:
Update batchUpdateHolidays() to handle trailing * on imported library code information, matching any library/location codes starting with the number preceding the *

### DIFF
--- a/code/web/services/Admin/AJAX.php
+++ b/code/web/services/Admin/AJAX.php
@@ -2004,8 +2004,8 @@ class Admin_AJAX extends JSON_Action {
 							$locationClosure = new LocationHours();
 							$locationClosure->locationId = $locId;
 							$locationClosure->day = $dayMap[$holidayDate];
-							$locationClosure->closed = 1;
 							if (!$locationClosure->find(true)) {
+								$locationClosure->closed = 1;
 								$locationClosure->open = '00:00:00';
 								$locationClosure->close = '00:00:00';
 								if ($locationClosure->insert()) {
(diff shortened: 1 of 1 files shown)


code/web/services/Admin/AJAX.php
```diff
@@ -2028,8 +2041,8 @@ public function batchUpdateHolidays() {
 						}
 					}
 				} if ((($scope == 'location' || $scope == 'all')) && strlen($holidayDate) == 3) {
-					$location->code = $libraryLocationCode;
-					if ($location->find(true)) {
+					$locations = preg_grep("/^" . $libraryLocationCode . "/", $locationCodes);
+					foreach ($locations as $loc){
```
mdnoble73: @K-Alette It looks like the block below this should use $loc rather than $location. Currently it's only updating one matching location rather than all of them. 

We should also look for the existence of an existing entry for that day. I'm getting duplicate Sunday closures in the database.

code/web/services/Admin/AJAX.php
```diff
@@ -1948,14 +1948,29 @@ public function batchUpdateHolidays() {
...
+
 		//go through each row and update according to user permissions and scope
 		foreach ($rows as $row) {
 			$libraryLocationCode = $row[0];
 			$holidayDate = $row[1];
-			$library = new Library();
-			$location = new Location();
 			$holidayDates = [];
 
+			$libraryLocationCodeTrimmed = rtrim($libraryLocationCode, '*');
```
mdnoble73: @K-Alette I think I gave you some bad advice. Rather than trimming off the "*" from the end, let's replace it with ".*" That should match the logic in Sierra. 

That way for location 83 we will get a match if the input data is 8*, but won't get a match if it's just 8.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #3919, opened 2026-03-11
DIS-2126: Slow query in DatabaseCleanup

kylemhall:


lathomas64:
Hooray speedups: https://www.navicat.com/en/company/aboutus/blog/1704-joins-versus-subqueries-which-is-faster

code/cron/src/com/turning_leaf_technologies/cron/DatabaseCleanup.java
```diff
@@ -207,55 +207,55 @@ private void removeUserDataForDeletedUsers(Connection dbConn, Logger logger, Cro
...
 
-			numUpdates = dbConn.prepareStatement("DELETE FROM user_link_blocks where primaryAccountId NOT IN (select id from user)").executeUpdate();
+			numUpdates = dbConn.prepareStatement("DELETE ulb FROM user_link_blocks ulb LEFT JOIN user u ON ulb.primaryAccountId = u.id WHERE u.id IS NULL").executeUpdate();
 			if (numUpdates > 0){
 				processLog.incUpdated();
 				processLog.addNote("Deleted " + numUpdates + " user link blocks where the primary account does not exist");
 			}
 
-			numUpdates = dbConn.prepareStatement("DELETE FROM user_link_blocks where blockedLinkAccountId NOT IN (select id from user)").executeUpdate();
+			numUpdates = dbConn.prepareStatement("DELETE ulb FROM user_link_blocks ulb LEFT JOIN user u ON ulb.blockedLinkAccountId = u.id WHERE u.id IS NULL").executeUpdate();
```
LiYanjun19: `blockedLinkAccountId` can be null, so we need to add a guard to only delete rows where `blockedLinkAccountId IS NOT NULL`

code/cron/src/com/turning_leaf_technologies/cron/DatabaseCleanup.java
LiYanjun19: can use left join 
`DELETE ul FROM user_link ul LEFT JOIN user u ON ul.primaryAccountId = u.id WHERE u.id IS NULL`

LiYanjun19: can use left join
`DELETE ul FROM user_link ul LEFT JOIN user u ON ul.linkedAccountId = u.id WHERE u.id IS NULL`

LiYanjun19 (changes requested):

LiYanjun19:
We should also add a release note.

code/cron/src/com/turning_leaf_technologies/cron/DatabaseCleanup.java
kylemhall: Thanks for the feedback @LiYanjun19!

mdnoble73:
Needs to be targeted to 26.05. Thanks

mdnoble73 (changes requested):
Needs to be targeted to 26.05. Thanks

mdnoble73:
Closing in favor of https://github.com/Aspen-Discovery/aspen-discovery/pull/4339

PR #3942, opened 2026-03-15
DIS-2123 Manage Materials Requests "All" sticky

stephenrwicks:
https://aspen-discovery.atlassian.net/browse/DIS-2123

Setting this dropdown in Manage Materials Requests to "All" attempts to write "all" to user_page_defaults.pageSize, which is an INT column, so it gets set to 0.

ManageRequests.php line ~385 checks !empty($pageDefaults->pageSize) so this gets skipped and defaults to 30.

Here I add an additional check for 0 to correct it to 'all' which is what the template is expecting.

Possibly there is a more holistic approach to fixing this which involves removing "all" as the value but I'm not sure what the standard Aspen uses is, so I'm going with minimal code changes here.


mdnoble73:
@stephenrwicks this doesn't seem to be working for me when I test it.

code/web/services/MaterialsRequest/ManageRequests.php
mdnoble73, line 382: @stephenrwicks looks like we need to convert the $materialsRequestsPerPage to 0 here before updating or update to allow the pagesize to allow text in the database.

mdnoble73 (changes requested):

PR #3957, opened 2026-03-17
DIS-2091: Accessible Browse Category Missing Link bugfix 

JonahCWilson:


catsoup11789:
Changes to the Search API here also affect LiDA, so please make sure you fully test that separately in ByWater's current production version as well as what's production for community code (since sub browse categories are processed different starting 26.01.00, which are not yet included in ByWater's production app which is still at 25.12.00) and include testing notes for that.

PR #3412, opened 2025-11-28
DIS-1597: aspen events registration  - WIP

Chloe070196:

PR #3958, opened 2026-03-18
DIS-1597: feat: aspen events registration

Chloe070196:
Currently undergoing internal testing / final code clean up. Will update once ready for review, raising now for visibility. Replaces [https://github.com/Aspen-Discovery/aspen-discovery/pull/3412](https://github.com/Aspen-Discovery/aspen-discovery/pull/3412)


mdnoble73 (changes requested):
Added these to Jira as well. Here is what I have so far: 

1. Within linkedAccounts.tpl, this sentence no longer makes sense: “The following accounts can view checkout, native Aspen events, and hold information from this account.  If someone is viewing your account that you do not want to have access, please contact library staff.” Also, Patrons will not really know what native Aspen events are.
2. Testing notes here don’t relate to actually doing registrations and all the positive and negative cases of that. Could those be updated? I’m basing my testing of the requirements (newer version from https://aspen-discovery.atlassian.net/browse/DIS-609)
3. In User.php setViewerIds could be better termed loadViewerIds. The set implies they are being changed, but they really aren’t. To avoid duplication of code, getViewerIds could also call getViewers and then just return the Id (or vice versa getViewers could call getViewerIds and then load the User objects from the ID). 
4. For Event Registration, I can see the modal body needing to be overridden per event type because different events will have different registration requirements. This should also be a translatable text block so different languages can be supported. 
5. Permissions Event Settings should reviewed to ensure they are at the appropriate level.

Chloe070196:
Hi @mdnoble73 - many thanks for the feedback so far, much appreciated!

I'd unfortunately been on leave for a bit, however I am back now and will be pushing changes to address it later this week.

Apologies about the test notes, I think the copy past from my work document onto the Jira ticket has truncated them. I'll amend that too.

One thing that may not be included in the present submission however: the override per Event Type for the registration modal body is something I would consider adding in the future, however it is not MVP level I don't think, sounds like a further enhancement to me. Would you agree?

Chloe070196:
I did notice there were conflicts (the usual .jar file and release notes), so I rebased the work on top of 26.04.00 and resolved those

mdnoble73:
@Chloe070196 can you resolve conflicts again?

Chloe070196:
@mdnoble73 all done! rebased on top of `26.05.00`, merge conflicts resolved

mdnoble73:
@Chloe070196 looks like it needs conflicts resolved again. We can work on getting these all merged Monday.

Chloe070196:
@mdnoble73 thanks for letting me know, all done - rebased, and merge conflicts resolved.

mdnoble73 (approved):

PR #3967, opened 2026-03-19
DIS-2250: Add OAuth2 Server

catsoup11789:
**Implements the following grants:**
- Authorization Code
- Password Grant
- Client Credentials Grant

**Includes support for:**
- Refreshing tokens
- Scope-based access control (minimally integrated for this initial implementation)
- Read or write access control per scope
- Rate limiting based on API
- PKCE compliance
- OpenID Connect (still very baby state, needs customization for Aspen)
- Manage OAuth2 clients in Aspen Admin

Based on The PHP League's OAuth2 Server library.

**Some things that could be added or enhanced based off what I've started:**
- Add OAuth2AuditLog to for logging authorization requests (approved/denied), token issuance/refresh, token revocation, failed authentication attempts. Would need data retention policy.
- Allowing admins to rotate/regenerate client_secrets for a client
- Data retention on rate limiting log (notes in the rate limiting section)
- Add a Token Revocation endpoint (/Authentication/OAuth2/Revoke) to allow users and clients to revoke tokens (RFC 7009)
- Add an Introspection endpoint (/Authentication/OAuth2/Introspect) to allow clients to check if a token is valid/active (RFC 7662)
- Add a Pushed Authorization Request endpoint (/Authentication/OAuth2/PushedAuthRequest) to allow clients to send a pre-authorization as post before requesting authorization from the user for added security and privacy of parameters (RFC 9126)
- Ensure all errors return standardized responses (RFC 6749) -- might be some outliers
- OpenID Connect needs fully specified
- Add bypass for rate limiting for trusted clients (like internal services or Aspen LiDA)
- Remove custom code from /lib/ or /dependencies/ and add into wrappers to ensure upgrading doesn't break anything  

# Setup
Generate RSA keys
```bash
cd /path/to/aspen-discovery/code/web/services/Authentication/OAuth2Server

openssl genrsa -out private.key 2048
openssl rsa -in private.key -pubout -out public.key
php -r "echo base64_encode(random_bytes(32));" > encryption.key.tmp && cat encryption.key.tmp

chmod 600 private.key
chmod 600 encryption.key
chmod 600 public.key

# Clean up if needed
rm encryption.key.tmp
```

# Creating Clients
Using a user who has the Administer OAuth2 Clients permission, access the admin interface:
1. System Configuration > OAuth2 Clients
2. Add New Client
3. Complete required fields:
   - **Client Name**: Should be descriptive, human-readable name for this client so you know what its going to be used for (you can't change this later)
   - **Client Type**: Choose based on your use case (see the grant types below, you can't change this later)
   - **OpenID Connect Support**: Check if this client is for OpenID
   - **Scopes**: Select minimum required scopes, not available for OpenID Connect clients
   - **Claims**: If using the client for OpenID Connect, select the claims allowed to request
   - **Redirect URI**: For web apps only, the URL where Aspen will send the user after they authorize the client to use their credentials. This is required for Authorization Code grants and must match the `redirect_url` value that the client sends in their authorization requests.
4. Save

At save, Aspen will auto-generate Client ID and Client Secret, which are needed for actually authorizing.

Note: The scoping still needs to be deeply integrated into the APIs, I have as an example started adding them into List API.

### Authorization Code (Web Application)

*Best for:*
Traditional websites and web applications with a server-side backend that need to access patron data and/or make requests on behalf of a patron (update a list, view holds, etc). With this grant the client never handles or sees patron login info but can make updates to the patron's data (i.e. create a new list or add a new item to a list) if it has permissions to write.

This is also the method used for OpenID Connect. 

*How it works:*
- Patron logs into your website (not Aspen)
- Your backend server (again, not Aspen) makes a call to Aspen for authentication
- Patron gets redirected to Aspen, if they are not logged in they will be prompted to login to Aspen (it also has a safety check to make sure an existing Aspen session is tied to the user the external website is trying to authorize)
- After patron logs in, they are prompted to authorize the app to access the data requested (tied to API permissions when not OpenID, if OpenID it is tied to claims)
- If the patron authorizes, they are redirected back to the `redirect_uri` with an access code
- If the patron cancels the authorization, they are still redirected back to the `redirect_uri` but with an `error=access_denied` parameter and no access code
- Your backend server exchanges the code for tokens using the client secret
- Your website now has an access token to call Aspen APIs on behalf of the patron

Token expires after 1 hour, but refresh tokens last 1 month

```bash
https://aspen.yourlibrary.com/Authentication/OAuth2/Authorize?
  response_type=code&
  client_id=aspen_xxxxx&
  redirect_uri=https://myclient.com/callback&
  scope=api:list:read&
  state=random123

curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -d "grant_type=authorization_code&code=CODE&client_id=aspen_xxxxx&client_secret=SECRET&redirect_uri=https://myclient.com/callback"
```

```bash
curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -H "Authorization: Basic YXNmZ...." \
  -d "grant_type=authorization_code&code=CODE&client_id=aspen_xxxxx&client_secret=SECRET&redirect_uri=https://myclient.com/callback"
```

**Note:** `redirect_uri` must match the value set on the client in Aspen Admin, and it is required for this grant.

#### Proof Key for Code Exchange (PKCE)
Supports the PKCE protocol for more security against potential authorization code interception attacks where a code is exchanged for tokens.

It is optional, but recommended.

When the client makes an authorization request (in addition to the other parameters above):
``` 
https://aspen.yourlibrary.com/Authentication/OAuth2/Authorize?client_id=aspen_xxxx&code_challenge=...&code_challenge_method=...

``` 

Supports `S256` (recommended) and `plain` for `code_challenge_method`

### Password Grant (Mobile Apps)

*Best for:*
Trusted apps like Aspen LiDA or other internal Aspen uses and never third-party apps as the client will have direct access to patron credentials which can raise security concerns.

We might want to add more security around creating clients with this grant, but I'm not sure. The specific use-case for a Password Grant is that there is no `redirect_uri` for the client.

*How it works:*
- Patron enters their barcode/pin in your app
- App sends those creds along with their client id and client secret
- Aspen returns an access token and refresh token
- App stores tokens securely and can now use the access token to call Aspen APIs on behalf of the patron

Token expires after 1 hour, but refresh tokens last 1 month

```bash
curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -d "grant_type=password&username=user&password=pass&client_id=aspen_xxxxx&client_secret=SECRET&scope=work:read"
```

```bash
curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -H "Authorization: Basic YXNmZ...." \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d "grant_type=passwords&username=12334566&passowrd=8sdfsdf&scope=work:read"
```

### Client Credentials (Service-to-Service)

*Best for:*
Backend services, system-to-system, scheduled jobs, or third-party systems. Probably the most common and for general use to "read" data but not "write" it, i.e. you have a WordPress plugin that you want to connect to the List API to display public lists on your website. If you need to "write" to Aspen for updating something on behalf of a patron, I would suggest the Authorization Code grant as it would allow the client to securely make those requests without ever handling the patron login credentials.

*How it works:*
- Your backend service has a client id and client secret that an admin created in Aspen
- Service makes a token request using those creds
- Aspen returns an access token
- Service can use this token to call APIs which persists for 4 hours (these requests are NOT made on behalf of a patron)

Token expires after 4 hours

```bash
curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -d "grant_type=client_credentials&client_id=aspen_xxxxx&client_secret=SECRET&scope=work:read+list:read"
```

```bash
curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -H "Authorization: Basic YXNmZ...." \
  -d "grant_type=client_credentials&scope=work:read+list:read"
```

## Refreshing the token
```bash
curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -d "grant_type=refresh_token&refresh_token=REFRESH_TOKEN&client_id=aspen_xxxxx&client_secret=SECRET"
```

## Using the token in API requests
```bash
curl -X GET "https://aspen.yourlibrary.com/API/ListAPI?method=getLibraryLists" \
  -H "Authorization: Bearer eyJ0eXAi..."
```

## Rate limiting
Uses a sliding window rate limiting approach by tracking requests per endpoint and either per client ID or IP address.

```https://aspen.yourlibrary.com/Authentication/OAuth2/Token``` allows 10 requests in a 1 minute window to prevent brute force token exchange attempts

```https://aspen.yourlibrary.com/Authentication/OAuth2/Authorize``` allows 20 requests in a 5 minute window to prevent brute force authorization attempts

```https://aspen.yourlibrary.com/API/*``` allows 1000 requests in a 60 minute window, this is default to all APIs using Abstract API, but each API can be configured separately in `OAuth2RateLimiter.php` or we can remove the fallback rate limit entirely if we only want to restrict on the authentication endpoints.

When the limit is exceeded the client will get this response (HTTP 429):
``` 
{
   "error": "rate_limit_exceeded",
   "error_description": "Rate limited exceeded. Try again in 45 seconds.",
   "retry_after": 45
}
```

### Monitoring rate limits
In Aspen Admin, there's a "OAuth2 Rate Limits" page to view real-time rate limit data to identify abusive clients, track window cycles, monitor activity patterns, troubleshoot rate limiting, and manual intervention for resetting the client.

Looks something like this:
```
Client ID: my_library_app
IP Address: 203.0.113.0
Endpoint: token
Request Count: 9 (out of 10 allowed)
Window Start: 2026-03-20 14:32:00
Last Request: 2026-03-20 14:32:47
```
### Privacy
**IP Addresses for the rate limiter are masked for anonymization** but not hashed because then it makes it impossible to actually track and identify.

### Data Retention
We probably need to decide on a data retention policy for the rate limit logs. How frequently do we want to purge the table (no sooner than the longest rate limit window)? Do we want to keep an archive table for auditing? Do we want any kind of reporting (i.e. how many requests, how many clients, how many times the rate limit was hit, etc)?

# Integrating into APIs
- Use OAuth2Middleware with Abstract API to authenticate requests and manage scoping using PHP docblock annotations to help maintainability instead of having multiple sources of truth to update when adding new endpoints, changing permissions, etc.
- Ideally, we'd drop support for `@token` after LiDA switches to `@oauth` and only use IP-based access for Aspen-to-Aspen requests (i.e. to an external Greenhouse server connected to many Aspen instances), and determine what a public (no auth, no IP restrictions) endpoint looks like and where it makes sense to provide them.

## Using PHP docblock annotations
- `@oauth (true/false)` indicates whether or not a method accepts OAuth2 Bearer token authentication.
- `@token (true/false)` indicates whether a method accepts the traditional username/password in POST server in conjunction with matching tokens from Greenhouse. Currently only for LiDA requests.
- `@public (true/false)` indicates whether a method allows public access without any authentication (open, free to the world, use sparingly). You can also enable/disable the the IP Address whitelisting (using the Allow API Access setting in IP Addresses) here.
- `@scopes (comma-separated list)` declares what OAuth2 scopes are required to access this method

Example:
```php

/** 
* Add item to list
* @oauth true
* @scopes list:read, list:write
* /

```

If a method does not have docblock annotations, or are incomplete, the following values are inferred by default:

```php

/** 
* Add item to list
* @oauth true
* @token true
* @public false
* @scopes [inferred from method name]
* /

```

For scopes by default it will look at the method name, for example with the List API:

Scope prefix: `'list'`

```
get*,list*,search*,find*,load*,retrieve*,show*,view* = list:read
create*,add*,edit*,delete*,update*,remove*,clear*,set* = list:write
anything else: list:read (fallback)
```

## Adding authentication into launch() function of APIs
Use `$this->handleAPIRequestAuto($_GET['method'], 'api_name')` to automatically discover auth requirements set in docblock annotations. The first parameter will obviously tell the API which method to use and will grab the docblock annotations for that to validate the request. The second parameter is used for rate limiting.

You can also use `$this->handleAPIRequest()` which will have a similar setup to how it works today with arrays for valid methods and will not look at docblock annotations. I included this approach as an example in User API.
```php
function launch() {
$method = $_GET['method'] ?? '';
$oauthMethods = ['getWorks', 'getUserInfo'];
$tokenMethods = ['getUserInfo', 'getList'];
$publicMethods = ['getSearchableLists'];

$this->handleAPIRequest($method, $oauthMethods, $tokenMethods, $publicMethods, 'api_name');
}
```

## Handling delegated API requests for user actions
In the already well used `$this->getUserForAPICall()` throughout the APIs, there is now a check for Password Grant OAuth2 authentication which will return an Aspen user object that was validated when the token was created. No additional work is needed here.

## Available Scopes
This is just what I initially setup based on the types of endpoints already established, can remove or extend as needed:

```
'work:read' => 'Work API Read Access',
'user:read' => 'User API Read Access',
'user:write' => 'User API Write Access',
'list:read' => 'List API Read Access',
'list:write' => 'List API Write Access',
'item:read' => 'Item API Read Access',
'event:read' => 'Event API Read Access',
'event:write' => 'Event API Write Access',
'search:read' => 'Search API Read Access',
'system:read' => 'System API Read Access',
'fine:read' => 'Fine API Read Access',
'community:read' => 'Community API Read Access',
'community:write' => 'Community API Write Access',
```

### Scope validation
This validates scopes in a couple ways:
1. Make sure the scope(s) included are valid as per OAuth2Client::getScopeOptions()
2. Make sure the scope(s) included are allowed for the client
3. Determine what scope is required to make the request and if the client has that scope

If any of these are ever "no" Aspen will return response for Missing required scope

# OpenID Connect (OIDC)
If a client is allowed OpenID connections, it will be able to use Aspen Discovery as an IdP (Identity Provider). This process is the same as the general Authorization Code grant, except after the user authorizes the client, they will only be able to access the OpenID claims instead of the APIs.

**OpenID Connect discovery document** is a JSON metadata file to allow clients to discover Aspen's IdP configuration for endpoints, scopes, and public signing keys.

```
GET https://aspen.yourlibrary.com/.well-known/openid-configuration
```

**OpenID Connect JSON web key set (JWKS)** to fetch the public keys for Aspen's IdP that they can use to verify authenticity and integrity of tokens.

```
GET https://aspen.yourlibrary.com/.well-known/jwks.json
```

**OpenID Connect UserInfo endpoint** is the only one I've initially included here. Depending on the allowed claims for the client, they can use this URL to request data about the user.

```
curl -X GET "https://aspen.yourlibrary.com/Authentication/OAuth2/UserInfo" \
  -H "Authorization: Bearer eyJ0eXAi..."
```
Using the `access_token` as an `Authorization: Bearer` header

# Testing
I included a Postman collection that allows you to test the different authentication flows and accessing a couple APIs, as well as OpenID features.


Chloe070196:
Hi there, have you had a chance to look at https://github.com/Aspen-Discovery/aspen-discovery/pull/3747 ?

3747 doesn't rely on any external libraries, which is something the community are keen on as we like to avoid dependencies.

(The reason not to use libraries is that Aspen does not have any package managers, making hard to ensure that these are kept up to date and secure.)

Looking forward to having more conversations around this!

catsoup11789:
> Hi there, have you had a chance to look at #3747 ?
> 
> 3747 doesn't rely on any external libraries, which is something the community are keen on as we like to avoid dependencies.
> 
> (The reason not to use libraries is that Aspen does not have any package managers, making hard to ensure that these are kept up to date and secure.)
> 
> Looking forward to having more conversations around this!

OAuth is something that doesn't need to be reinvented. We have used external libraries in other cases such as SAML authentication (OneLogin) and RapiDocs (API docs). PHPLeague's library gives us assurance about security, standards compliance (PKCE, PSR, OAuth2 RFC 6749, 6740) and easier integration with other services (like OpenID).

ETA: The PHPLeague library has only had 2 upgrades in the last year, and one of them was essentially a bump for supporting PHP 8.5. We do a lot more to support upgrading the Expo library for Aspen LiDA.

Jacobomara901:
I've just put a message on the Jira for this, probably could have put it here instead in retrospect! TLDR on that comment is I'd love to collaborate here and I'd love to use a library for OAuth (and generally for other things too). I think we should also discuss package management if we want to continue adding libraries to the product as it inevitably increases the maintenance workload. 
We use yarn for LiDA so there is precedent for this in our product stack, Composer then kinda seems like the obvious choice for PHP!

PR #4006, opened 2026-03-24
DIS-2123 Manage Materials Requests "All" sticky

stephenrwicks:
https://aspen-discovery.atlassian.net/browse/DIS-2123

Setting this dropdown in Manage Materials Requests to "All" attempts to write "all" to user_page_defaults.pageSize, which is an INT column, so it gets set to 0.

Updates that column to varchar(10) not null default 'all' - Allows this behavior to work


code/web/sys/DBMaintenance/version_updates/26.04.00.php
```diff
@@ -60,9 +60,16 @@ function getUpdates26_04_00(): array {
 		//tomas
 
 		// stephen
+			'change_user_page_defaults.pageSize_to_varchar' => [
+			'title' => 'Change user_page_defaults.pageSize column to varchar',
+			'description' => 'Modifies the pageSize column to allow the value "all"',
+			'continueOnError' => false,
+			'sql' => [
+				"ALTER TABLE user_page_defaults MODIFY COLUMN pageSize VARCHAR(10) NOT NULL DEFAULT 'all'",
```
mdnoble73: @stephenrwicks we won't want to set the default to all here. Let's leave it set to blank since we handle that when displaying records. That way it will also work on pages where there isn't an "all" option.

mdnoble73 (changes requested):

code/web/sys/DBMaintenance/version_updates/26.04.00.php
stephenrwicks: Changed this to ALTER TABLE user_page_defaults MODIFY COLUMN pageSize VARCHAR(10) NULL

mdnoble73 (approved):

PR #4009, opened 2026-03-25
DIS-2110: Improve loading DVD Covers

catsoup11789:


code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -1217,16 +1228,108 @@ function omdb(OMDBSetting $omdbSettings, $title = null, $shortTitle = null, $yea
...
-		if (!is_array($year)) {
-			$year = [$year];
-		}
-		foreach ($year as $curYear) {
-			if (strpos($curYear, ',')) {
-				$years = explode(',', $curYear);
-				$year = array_merge($year, $years);
+		$years = [];
+		$directors = [];
+		if ($this->loadGroupedWork() && $this->groupedWork) {
```
JonahCWilson, line 1233: Suggestion:

Move everything inside this if-block into one or more sub functions. processRelatedRecords or something.

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -1217,16 +1228,108 @@ function omdb(OMDBSetting $omdbSettings, $title = null, $shortTitle = null, $yea
...
+									$yearFromMarc = substr($data, 7, 4);
+									if (!empty($yearFromMarc) && is_numeric($yearFromMarc) && $yearFromMarc !== '9999' && $yearFromMarc !== '0000') {
+										$years[] = $yearFromMarc;
+									}
+								}
+							}
+						}
+
+						$field508 = $marcRecord->getFields('508');
+						if ($field508) {
```
JonahCWilson, line 1274: Each of these field checks could be its own function.  processField508.

Another thing, you can reduce nesting by using a guard clause. If $field508 is null, we do nothing, so check for that first

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -1217,16 +1228,108 @@ function omdb(OMDBSetting $omdbSettings, $title = null, $shortTitle = null, $yea
...
+							foreach ($field508 as $field) {
+								if ($field->getSubfield('a')) {
```
JonahCWilson, line 1276:
```
if (!$field->getSubField('a')){
    continue;
}
```

// Move everything back an indentation block.

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -697,6 +697,17 @@ private function getCoverFromMarc(File_MARC_Record|false|null $marcRecord = null
 						}
 					}
 				}
+
+				if ($marcField->getIndicator(1) == '7' && $marcField->getSubfield('3')) {
```
JonahCWilson, line 701: My apologies, I didn't realize there was collapsed code so my understanding of the logic was wrong.  Even so, you can pull this out into  separate function and reduce nesting by using guard clauses.

```
if(NOT BOOL){
    return early;
}
// Do next checks
```

code/web/sys/Covers/BookCoverProcessor.php
mdnoble73: Since the functionality is not reused in other places, that would just lead to a clutter of functions. If we thought we would reuse the functionality it should go in the MarcRecordDriver potentially.

code/web/sys/Covers/BookCoverProcessor.php
mdnoble73: This is a stylistic preference. Both are fine and in my mind equally readable.

PR #4017, opened 2026-03-25
DIS-2206: Solr improvements/query optimization

JonahCWilson:
Putting this up as a draft in order to get some feedback.

Ignore the logging, I will remove most of that. Did some minor refactoring of _select to use helper functions. Most of the code is identical.

The main thing I wanted to discuss was in search().  I found that certain queries on partner sites were very slow, and I did an analysis of the query being built and found that there were a number of inefficiencies.  The query being built is a massive nested set of OR clauses which check against every possible field along with boosting.

The proposed solution was to use edismax + qf + fl for queries along with bf for things that are seldom searched, like barcode.

Another aspect is that the boosting is built in a nested structure which is calculated on a per-document basis. Additionally, there are some instances of the same value being boosted twice with different values (e.g. author^200 & author^50, although this may just be my local dev environment).  The "best" suggested solution for this was to precompute the values and index them.  I haven't had time to test that route, but it also suggested doing a bit of preprocessing to flatten the boosting to make it more efficient once it gets to solr.

Please let me know if you have any questions.  The code right now is substantially faster, but the boosting logic is off and I'm still working on getting that part right.  Just wanted to see if anyone had any questions, concerns, or opinions.


mdnoble73:
Repeating from Slack: 
Do you have examples of searches pre/post changes with differences in speed and results (both counts and relevancy)? Is the goal of the changes improving query speed or are there secondary goals you are aiming for? A Jira would be useful to explain some of the intent.

PR #4019, opened 2026-03-26
DIS-2127: Fix koha 952i suppression

lucasmontoya13:
Aspen's Koha reindexer was checking MARC subfield 952$i (inventory/stock number) as a suppression flag, so if its value was 1, the item was marked suppressed and the record disappeared from the catalog. This is incorrect: 952$i maps to items.stocknumber in Koha and has no relation to item visibility. The fix removes that check entirely; suppression continues to be driven by the correct subfields (952$0 for withdrawn, 952$1 for lost status).

Test plan

Before patching:
  - Edit a Koha item and set 952$i to 1
  - Force reindex the record
  - The record disappears from the catalog; Staff View shows "Item X subfield i set to 1" and "Record had no items"

After patching:
  - Edit a Koha item and set 952$i to 1
  - Force reindex the record
  - The record continues to display normally in the catalog; the inventory number has no effect on suppression


mdnoble73 (approved):

mdnoble73:
@lucasmontoya13 can you resolve conflicts on this?

lucasmontoya13:
Done!

mdnoble73 (approved):

PR #4022, opened 2026-03-26
DIS-2012: When loading language from 041 trim anything that isn't a letter

K-Alette:
Add regex to trim any character that isn't a letter when getting language information from 041 
Update release notes


code/reindexer/src/org/aspen_discovery/grouping/RecordGroupingProcessor.java
```diff
@@ -908,7 +908,23 @@ public String getLanguageBasedOnMarcRecord(org.marc4j.marc.Record marcRecord) {
 			}
 		}
 		if (activeLanguage == null || activeLanguage.equals("|||") || activeLanguage.equals("   ") || activeLanguage.contains(" ")){
```
JonahCWilson, line 910: Consider consolidating this to use a regex for matching against the various string values.

Also, you can negate this and return activeLanguage. This allows you to take everything inside and bring it out reducing indentation.

JonahCWilson: At this point, you can just return 'unk', as that appears to be the default if nothing is ever found.

code/reindexer/src/org/aspen_discovery/grouping/RecordGroupingProcessor.java
```diff
@@ -908,7 +908,23 @@ public String getLanguageBasedOnMarcRecord(org.marc4j.marc.Record marcRecord) {
...
-			activeLanguage = "unk";
+			String secondaryLanguageField = "041a";
+			activeLanguage = null;
+			languages = MarcUtil.getFieldList(marcRecord, secondaryLanguageField);
+			for (String language : languages){
+				language = language.replaceAll("^[^a-zA-Z]+|[^a-zA-Z]+$|\\p{Punct}", "");
+				if (activeLanguage == null){
```
JonahCWilson, line 916:
```
if (activeLanguage == null){
    activeLanguage = language;
    continue; // Immediately goes to next value in loop.
}

if(!activeLanguage.equals(language){
    activeLanguage = "mul";
    break;
}
```

You can also just return "mul" and it'll have the same effect.

code/reindexer/src/org/aspen_discovery/grouping/BaseMarcRecordGrouper.java
```diff
@@ -251,12 +251,25 @@ private void setLanguageBasedOnMarcRecord(Record marcRecord, GroupedWork workFor
 		}
 		//Check to see if the language has a space in it.
 		if (activeLanguage == null){
+			String secondaryLanguageField = "041a";
```
JonahCWilson, line 254: Same as down below.  Since this logic is roughly the same, you could probably abstract these out into a small helper function and run `activeLanguage = getActiveLanguage();`

mdnoble73 (approved):

PR #4024, opened 2026-03-26
DIS-2187 Link to all holds after successful hold

stephenrwicks:
https://aspen-discovery.atlassian.net/browse/DIS-2187

Adds a button to the modal after a successful hold that links to the holds page, except when using Sierra.

Doesn't display the button when the user checks "Log me out after requesting the item" since this will most likely break in all cases. Sets buttons to null

It looks like this functionality was basically there already but not used. A an anchor tag is created in User.php around ~2494 etc. The drivers also pass up a button which I am taking advantage of instead. Deleted the hold buttons in User.php

Removed a few lines of code that inject some buttons from record.js and used the showMessageWithButtons function instead. It looks like that code did nothing (possibly copied over from other functions)


mdnoble73:
@stephenrwicks the code you removed was injecting buttons on the main page to tell the patron the title was on hold for them. That should still be included. Using the showMessageWithButtons is correct. 

It also looks like "Log me out after requesting the item" is no longer actually logging the user out although that doesn't look to have been broken as part of these changes. I'll put in a separate Jira for that.

PR #4054, opened 2026-04-03
DIS-2188: Allow web resources to be excluded from Explore More.

attheshow:
DIS-2188: Allow web resources to be excluded from Explore More.


code/web/sys/ExploreMore.php
```diff
@@ -213,6 +213,35 @@ protected function loadWebIndexerOptions($activeSection, $exploreMoreOptions, $s
...
+				// If field exists, re-run with filter.
+				if ($hasShowInExploreMore) {
+					$searchObjectSolr = SearchObjectFactory::initSearchObject('Websites');
+					$searchObjectSolr->init();
+					$searchObjectSolr->disableSpelling();
+					$searchObjectSolr->setSearchTerms([
+						'lookfor' => $searchTerm,
+						'index' => 'WebsiteKeyword',
+					]);
+					$searchObjectSolr->addFilter('show_in_explore_more:true');
```
mdnoble73: @attheshow this is really close. Rather than using addFilter, use addHiddenFilter. That will prevent the show_in_explore_more facet from processing as a sidebar facet which has errors because the facet isn't defined.

mdnoble73 (changes requested):

code/web/sys/ExploreMore.php
attheshow: @mdnoble73 Gotcha. OK, I've updated this PR to use addHiddenFilter rather than addFilter for show_in_explore_more.

mdnoble73 (approved):

mdnoble73 (approved):

PR #4082, opened 2026-04-10
DIS-1191: feat: Koha hyperholds in Aspen

Chloe070196:


code/web/Drivers/Koha.php
```diff
@@ -9335,4 +9335,159 @@ private function getAccountLineDescription($accountline): string {
...
+	}	
+
+	public function supportsHyperholdsGrouping(): bool {
+		return $this->isDisplayAddHoldGroupsEnabledInKoha();
+	}
+
+	private function isDisplayAddHoldGroupsEnabledInKoha(): bool {
+		global $logger;
+
+		try {
```
JonahCWilson, line 9466: This try-block is perfectly fine as-is, so no need to change it, but the intermediate checks may be unnecessary.

For example, if the dbConnection is invalid, the call to mysqli_query will fail, throw an error, then return false in the catch block.

Because this is a SELECT query, mysqli_query will return either an object or false.  In the false case, an exception will be thrown + caught.

The if(!row ...) check is fine and should stay.

code/web/RecordDrivers/GroupedWorkDriver.php
```diff
@@ -2245,6 +2246,34 @@ public function getSearchResult($view = 'list'): string {
...
 
 		$timer->logTime("Assigned all information to show search results");
+
+		$user = UserAccount::getActiveUserObj();
+		$accountProfile = $library->getAccountProfile();
+		$allowHoldsToBeGrouped = false;
+
+		if ($accountProfile) {
+			$ils = $accountProfile ? $accountProfile->ils : '';
+			if ($ils == 'koha') {
```
JonahCWilson: Consider moving this into a helper function.

Right now it's just `if ($ils == 'koha')` but in some amount of time, it could become `else if($ils == '<ils_name>')` and nearly identical logic is copy-pasted multiple times.  If every ILS has different logic, then we can do something like:

```
switch($ils){
    case 'koha':
        doKohaSpecificLogic();
    case 'evergreen':
        doEvergreenSpecificLogic();
    ...
}
```

And if the logic is the same with just some different variables we can do:

```
processHyperHolds($necessaryVar1, $necessaryVar2, $etc);
```

Chloe070196: I think I'd even look at implementing `CatalogFactory::getCatalogConnectionInstance`, then use `supportsHyperholdsGrouping` (defined on AbstractIlsDriver to return false by default)

code/web/Drivers/Koha.php
Chloe070196: Agreed, the dbConnection guard clause should go

code/web/services/MyAccount/AJAX.php
```diff
@@ -3902,6 +3903,50 @@ public function getHolds(): array {
 				$allowSelectingHoldsToExport = $library->allowSelectingHoldsToExport;
 			}
 
+			$accountProfile = $library->getAccountProfile();
+			$allowHoldsToBeGrouped = false;
+			if ($accountProfile) {
```
mdnoble73: This should be moved to User.php which can then check the catalog connection for itself as well as related library settings.

code/web/services/MyAccount/AJAX.php
```diff
@@ -11501,4 +11618,284 @@ function listsTransferProcess(): array {
 
 		return $results;
 	}
+
+	public function groupPatronHolds() {
+		global $interface;
+		global $logger;
+
+		if (!UserAccount::isLoggedIn()) {
```
mdnoble73: This can use the new guard functions

code/web/services/MyAccount/AJAX.php
```diff
@@ -11501,4 +11618,284 @@ function listsTransferProcess(): array {
...
+				'text' => "Ungroup Holds",
+				'isPublicFacing' => true,
+			]) . "</button>",
+		];
+	}
+
+	public function deleteHoldGroup() {
+		require_once ROOT_DIR . '/sys/User/Hold.php';
+		$user = UserAccount::getLoggedInUser();
+
```
mdnoble73: This should also use the new guard functions

code/web/services/Record/AJAX.php
```diff
@@ -353,6 +353,59 @@ function getPlaceHoldForm(): array {
...
 
+			$allowEditionSelection = (isset($_REQUEST['allowEditionSelection']) && $_REQUEST['allowEditionSelection'] == '1');
+			$interface->assign('allowEditionSelection', $allowEditionSelection);
+
+			$currentFormat = null;
+			$accountProfile = $library->getAccountProfile();
+			$allowHoldsToBeGrouped = false;
+			if ($accountProfile) {
+				$ils = $accountProfile ? $accountProfile->ils : '';
+				if ($ils == 'koha') {
```
mdnoble73: This can be moved to User.php where the logic can be reused

code/web/services/Record/AJAX.php
```diff
@@ -624,16 +677,54 @@ function getPlaceHoldForm(): array {
 					'success' => true,
 				];
 				if ($holdType != 'none') {
+					$accountProfile = $library->getAccountProfile();
```
mdnoble73: This can reuse functionality if moved to User.php

code/web/services/Record/AJAX.php
```diff
@@ -2490,4 +2581,315 @@ function getLargeCover() : array {
...
 		];
 	}
+
+	public function requestHyperholdConfirmation() {
+		global $interface;
+
+		$groupedWorkId = $_REQUEST['groupedWorkId'] ?? null;
+		$variationId = $_REQUEST['variationId'] ?? null;
+
+		$user = UserAccount::getLoggedInUser();
```
mdnoble73, line 2571: This should use the new guard functionality

mdnoble73 (changes requested):
Did a partial review of code, see inline comments. Also is this compatible with the new edition selection where a user can select which edition they want a hold on from the popup?

Chloe070196:
@mdnoble73 thanks for the review! Today's a UK holiday so I won't get to looking at addressing the above until later this week I'm afraid. Agreed on the requested refactors, I'll sort those. As for functionality compatibility, that needs assessing. I'm guessing this had better target 26.06.00 then?

mdnoble73:
@Chloe070196 I think in terms of targeting 26.05 vs. 26.06 it's up to you depending on how long changes will take. If they will be after Wednesday we should likely push to 26.06

Chloe070196:
@mdnoble73 in that case, I'll look into it tomorrow and get back to you then. Thanks!

Chloe070196:
Converting this to draft briefly as I work on refactoring and test for any functionality regression the refactor work may introduce.

code/web/RecordDrivers/GroupedWorkDriver.php
Chloe070196: Done!

code/web/services/MyAccount/AJAX.php
Chloe070196: Absolutely agree - refactored to DRY and simplify, and moved the logic into a User util

PR #4085, opened 2026-04-10
DIS-2247: amazon ses settings email bug

JonahCWilson:
Bugfix for https://aspen-discovery.atlassian.net/browse/DIS-2247

This change fixes two bugs:
1. Type mismatch between two functions causing a runtime error.
2. Email not being sent on successful form submission when invalid email provided in custom form builder.


JonahCWilson:
Just noticed it has some extra stuff from 26.04, let me do some maintenance to make it just my changeset.

mdnoble73 (approved):

PR #4100, opened 2026-04-13
DIS-2130: (follow-up) Add comments and use variables for series link

kidclamp:
I used variables for the series links, to remove need for a second set of conditionals

Additional series as well, but I add comments to the function, if I followed, additional series either come from:
series index if enabled
MARC record if using novelist or not

Novelist can only return a single series

### DIFF
--- a/code/web/interface/themes/responsive/GroupedWork/series-shared.tpl
+++ b/code/web/interface/themes/responsive/GroupedWork/series-shared.tpl
@@ -1,21 +1,20 @@
 {if !empty($summSeries)}
 	{assign var=totalSeriesShown value=0}
+	{assign var=additionalSeriesLink value="/Search/Results?lookfor={$summSeries.seriesTitle|escape:url}&searchIndex=Series"}
 	{if !empty($summSeries.fromNovelist)}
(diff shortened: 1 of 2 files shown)


code/web/interface/themes/responsive/GroupedWork/series-shared.tpl
```diff
@@ -1,21 +1,20 @@
 {if !empty($summSeries)}
 	{assign var=totalSeriesShown value=0}
+	{assign var=additionalSeriesLink value="/Search/Results?lookfor={$summSeries.seriesTitle|escape:url}&searchIndex=Series"}
```
mdnoble73, line 3: This is setting the additional seriesLink based on the main link so they would all be the same which isn't correct.

mdnoble73 (changes requested):

code/web/interface/themes/responsive/GroupedWork/series-shared.tpl
kidclamp: Yeah, too early in the AM, I see what I did there

PR #4101, opened 2026-04-13
DIS-2249: Fix permissions for editing 'Layout Settings'

lucasmontoya13:
This patch fix an issue where users with either "Administer All Themes" or "Administer Library Themes" were able to edit 'Layout Settings'

Test plan (KTD & ADB setup)

Before

Log in to Aspen with a user who has either 'Administer All Themes' or 'Administer Library Themes'.
Go to Aspen Administration → Themes and Layout.
The user can edit 'Layout Settings' (this is incorrect).

After
4. Repeat steps 1–2.
5. 'Layout Settings' is no longer visible if the user does not have the appropriate permissions.

### DIFF
--- a/code/web/sys/LibraryLocation/Library.php
+++ b/code/web/sys/LibraryLocation/Library.php
@@ -1293,7 +1293,7 @@ class Library extends DataObject {
 						'values' => $layoutSettings,
 						'label' => 'Layout Settings',
 						'description' => 'Layout Settings to apply to this interface',
-						'permissions' => ['Library Theme Configuration'],
+						'permissions' => ['Administer Library Layout Settings','Administer All Layout Settings'],
 					],
 					'homeLink' => [
 						'property' => 'homeLink',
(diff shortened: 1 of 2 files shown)


code/web/sys/LibraryLocation/Library.php
```diff
@@ -1278,7 +1278,7 @@ static function getObjectStructure(string $context = ''): array {
 						'values' => $layoutSettings,
 						'label' => 'Layout Settings',
 						'description' => 'Layout Settings to apply to this interface',
-						'permissions' => ['Library Theme Configuration'],
+						'permissions' => ['Administer Library Layout Settings'],
```
mdnoble73: @lucasmontoya13 this should also allow Administer Library Layout Settings

mdnoble73 (changes requested):

code/web/sys/LibraryLocation/Library.php
lucasmontoya13: Do you mean 'Administer All Layout Settings'? 
The permission i added is 'Administer Library Layout Settings'

Or do you mean that, in addition to 'Library Theme Configuration', it also need to have 'Library Theme Configuration' as well ?

mdnoble73 (changes requested):
This does not do what the Jira states. Looks like this adds permissions rather than taking anything away.

code/web/sys/LibraryLocation/Library.php
lucasmontoya13: @mdnoble73 based on your last comment, what is wrong with this solution? By replacing "Library Theme Configuration" with "Administer Library Layout Settings" would allow the appropriate administrator to configure the layout settings.

mdnoble73: The changes do not match the description. The description says there is a problem with Layout Settings, but the change made was in Library.php

mdnoble73: @lucasmontoya13 this should allow Administer Library Layout Settings and Administer All Layout Settings

mdnoble73 (changes requested):

lucasmontoya13:
@mdnoble73 Done! Thank you

mdnoble73 (approved):

PR #4111, opened 2026-04-14
DIS-1909: feat: event waiting lists and email notifications

Chloe070196:
DEPENDS ON: https://github.com/Aspen-Discovery/aspen-discovery/pull/3958

The second submission in our series of Aspen Events enhancements: after registration, we add waiting list capabilities. 

In draft state at the minute as we carry out internal code review, and wait for PR 3958 to have been upstreamed (branch DIS-1909_waiting_lists_and_notifications-SUBMISSION includes both waitlist code changes and the registration changes it builds on top of). 

Will rebase / amend as needed, depending on PR 3958 's progress. Release notes to be added.

The test suite in particular is just a draft (did have a fair few successful runs of it locally)


install/updateCron_26.06.00.php
mdnoble73, line 1: This should be renamed updateCron_26.06.00.php

install/updateCron_26.06.00.php
```diff
@@ -0,0 +1,56 @@
...
+		// Go through each line of the cron settings file
+		while (($line = fgets($fhnd)) !== false) {
+			if (strpos($line, 'updateEventRegistrationInvites') > 0) {
+				$insertUpdateEventRegistrationInvites = false;
+			}
+			// Check for the specific marker for event invites cron job
+			if (strpos($line, 'Update Event Registration Invites Job') > 0) {
+				if ($insertUpdateEventRegistrationInvites) {
+					// Add the cron job for sending campaign emails before the end of the file
+					$lines[] = "######################################\n";
```
mdnoble73, line 24: This does not match the timing from the crontab_setting templates, which is correct?

install/updateCron_26.06.00.php
```diff
@@ -0,0 +1,56 @@
...
+					$lines[] = "# Update Event Registration Invites Job\n";
```
mdnoble73, line 25: This does not match the timing from the crontab_setting templates, which is correct?

code/web/sys/Utils/DateUtils.php
mdnoble73, line 1: I think this works for now, but we'll likely want the date to be customizable based on locale. In the U.S. We would do l F jS Y

mdnoble73: Actually, this is probably doing right now. You could: 
$formatter = new IntlDateFormatter('en_US', IntlDateFormatter::FULL, IntlDateFormatter::NONE);
return $formatter->format($date); 

And plugin the proper locale (first param) based on the current language. You might need to check if the - in our locale needs to be converted to an _

code/web/sys/Events/UserAspenEventInstanceRegistration.php
```diff
@@ -17,6 +20,189 @@ public function getUniquenessFields(): array {
...
+	 */
+	public static function getUsersGroupedByStatusForInstances(array $eventInstanceIds): array {
+		if (empty($eventInstanceIds)) {
+			return [];
+		}
+
+		$instanceIdList = implode(',', array_map('intval', $eventInstanceIds));
+
+		$registration = new UserAspenEventInstanceRegistration();
+		$registration->whereAdd("eventInstanceId IN ($instanceIdList)");
```
mdnoble73: There is a whereAddIn function you can use for this.

code/web/sys/Events/EventInstance.php
```diff
@@ -224,4 +374,219 @@ public function hasAvailableSeats(int $requestedSeats = 1): bool {
...
+			'library' => $homeLibrary,
+		]);
+	}
+
+	public function sendEventEmail(int $userId, string $templateName, array $parameters): bool {
+		require_once ROOT_DIR . '/sys/Email/Mailer.php';
+		require_once ROOT_DIR . '/sys/Email/EmailTemplate.php';
+		global $logger;
+
+		$emailTemplate = EmailTemplate::getActiveTemplate($templateName);
```
mdnoble73, line 550: If you pass the event to getActiveTemplate now, it will handle getting the proper language for the user if called from cron.

code/web/sys/Email/EmailTemplate.php
```diff
@@ -361,6 +363,29 @@ private function applyParameters($text, $parameters) {
 			$text = str_replace('%searchHistory.updatedSearchesWithSampleTitles%', $parameters['searchHistory']['updatedSearchesWithSampleTitles'] ?? '', $text);
 			$text = str_replace('%searchHistory.updatedSearchesHtml%', $parameters['searchHistory']['updatedSearchesHtml'] ?? '', $text);
 			$text = str_replace('%searchHistory.updatedSearches%', $parameters['searchHistory']['updatedSearches'] ?? '', $text);
+		} elseif ($this->templateType == 'registerForEventFromWaitingList') {
+			$text = str_replace('%event.title%', $parameters['eventTitle'] ?? '', $text);
+			$text = str_replace('%event.date%', $parameters['eventDate'] ?? '', $text);
+			$text = str_replace('%event.time%', $parameters['eventTime'] ?? '', $text);
+			$text = str_replace('%canRegisterUntil%', $parameters['canRegisterUntil'] ?? '', $text);
+		} elseif ($this->templateType == 'eventCancellation') {
+			$statusMessages = [
```
mdnoble73: The status message and change type need to be translated to match the language for the template.

code/web/sys/Email/EmailTemplate.php
```diff
@@ -40,6 +40,8 @@ static function getObjectStructure(string $context = ''): array {
 				'campaignComplete' => 'Campaign Complete',
 				'milestoneComplete' => 'Milestone Complete',
 				'staffCampaignComplete' => 'Campaign Complete Staff Alert',
+				'registerForEventFromWaitingList' => 'Register for Event From Waiting List',
```
mdnoble73, line 43: These could use default text to make setup easier.

code/web/interface/themes/responsive/events-notifications.tpl
```diff
@@ -0,0 +1,18 @@
+<div class="row" id="event-to-register-notification-container">
+	<div class="card shadow-sm" style="border: 1px solid #337ab7; padding: 20px; margin: 30px;">
+	<div class="card-body">
+		<div class="text-left">
+			<h2 class="card-title text-primary">
+				{translate text="Events to Register in!"}
```
mdnoble73: This needs the isPublicFacing parameter set

code/web/interface/themes/responsive/MyAccount/myPreferences.tpl
```diff
@@ -197,6 +197,26 @@
 							</div>
 						{/if}
 
+						{if array_key_exists('Events', $enabledModules)}
```
mdnoble73: This should check if Aspen Native events are active. Other event integrations would not have notifications (at least not that we control)

mdnoble73 (changes requested):
Took a first pass through

code/web/sys/Email/EmailTemplate.php
Chloe070196: Removing the hardcoded messages entirely. Instead, administrators can set separate event cancellation email templates for each registration status.  When an event is cancelled, the system will check the registration
status of each user and select the correct template accordingly.

install/updateCron_26.06.00.php
Chloe070196: Glad you spotted this! With the introduction of waitingListInviteExpiryHours we really do need this to run at a much higher frequency to accommodate for that setting. 

The solution I have for now is \*/5 * * * *   (so at the latest an invitation is sent 5' too late, and the true window for the promoted patron to register is n\*60 +5' (likely an acceptable discrepancy) instead of n\*60)

code/web/sys/Email/EmailTemplate.php
Chloe070196: What I've done here is add default email templates via db update - based it off your DIS-2142 default email template for saved searches - hope that was the right move!

install/updateCron_26.06.00.php
mdnoble73: Does this actually need to run as root? Should be able to run as the aspen user

sites/template.linux/conf/crontab_settings.txt
```diff
@@ -162,4 +162,10 @@
 #######################################
 0 */12 * * * root php /usr/local/aspen-discovery/code/web/cron/purgeUserAppRequestLogs.php {sitename}
 
+##################################################
+# Update Expired Registration Invites for Events #
+##################################################
+*/5 * * * * root php /usr/local/aspen-discovery/code/web/cron/updateEventRegistrationInvites.php {sitename}
```
mdnoble73: Does this actually need to run as root? Should be able to run as the aspen user

mdnoble73 (approved):

PR #4114, opened 2026-04-15
DIS-2282 Load Syndetics Unbound initiator script async:

Lauren-at-LibraryThing:
Add async attribute to the Syndetics Unbound initiator script tag so slow responses from unbound.syndetics.com no longer block page rendering. Defer the runUnboundWithMetadata call until the initiator has loaded to avoid a race on LibraryThingConnector.

One way to test this fix is using dev tools; in the network tab (chrome), find` initiator.php`, right-click > Throttle requests > Throttle request URL. The catalog page now loads quickly despite the simulated slow load times for Syndetics Unbound. 

image


mdnoble73:
Merged this manually

PR #4117, opened 2026-04-15
DIS-2032 Custom Event Covers

JonahCWilson:
During Aspen Event Creation (This is only for aspen events), add a checkbox so that libraries can choose to have a custom image showup in search results rather than the default date cover.


code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -229,12 +229,7 @@ public function loadCover(array $configArray, Timer $timer, Logger $logger) : bo
 	}
 
 	private function getHooplaCover($id) : bool {
-		if (str_contains($id, ':')) {
-			[
-				,
-				$id,
-			] = explode(":", $id);
-		}
+		$id = $this->normalizeEventId($id);
```
mdnoble73: Hoopla is not an event, the name should be changed if it applies to both

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -1868,13 +1840,18 @@ private function getSeriesMemberCover($id) : bool {
 		}
 	}
 
-	private function getCourseReservesCover(string $id) : bool {
+	private function normalizeEventId(string $id) : string {
```
mdnoble73: The name of this should be changed since it is used generically

mdnoble73 (changes requested):
See inline changes. 

Also Events are intentionally designed to be restrictive for permissions since they are designed for use in prison libraries. The option to show this should be on the Event Type and then should bubble down to the Event if allowed by the Event Type administrator.

PR #4140, opened 2026-04-19
DIS-2189: Add new configuration section for Explore More within Local Catalog Enrichment. Sources can now be added/removed from display and put in order of preference.

attheshow:
DIS-2189: Add new configuration section for Explore More within Local Catalog Enrichment. Sources can now be added/removed from display and put in order of preference.


code/web/sys/ExploreMoreSourceGroup.php
```diff
@@ -0,0 +1,130 @@
...
+				'displayField' => 'sourceName',
+			],
+		];
+		self::$_objectStructure[$context] = $structure;
+		return $structure;
+	}
+
+	public function update(string $context = '') : int|bool {
+		$result = parent::update($context);
+		if (isset($this->exploreMoreSources) && is_array($this->exploreMoreSources)) {
```
mdnoble73, line 119: The exploreMoreSources need to be defined as a property or we see notifications. By convention, we do these as protected $_exploreMoreSources so the DataObject doesn't get confused during saving, but the remaining code understands that it is a property.

code/web/sys/ExploreMoreSourceGroup.php
mdnoble73, line 1: We need to make sure that the insert and update methods save the explore more sources properly. I'm not getting them to update properly currently.

mdnoble73 (changes requested):
I'm not getting the Explore More Sources to update properly when saving.

PR #4169, opened 2026-04-23
DIS-2322: Zero out holdcount + release notes

JonahCWilson:
See Jira for notes + example.  I dug into the logic a bit in terms of where things are being incremented/decremented, but I didn't find the specific area where it was causing this, and this fix is used in other areas within the javascript too.


mdnoble73 (changes requested):
It looks like the root cause is in $this->updateCachesForCancelledHold. That should be updated to take the source rather than hardcoding ils. That will fix the root issue.

mdnoble73 (approved):

PR #4185, opened 2026-04-28
DIS-1723 ampersand covers

lathomas64:
https://aspen-discovery.atlassian.net/browse/DIS-1723

- load a record page with an ampersand in the book cover’s id (example: https://discover.knoxcountylibrary.org/oreilly/on1386275878?searchId=13820819 )
- With the code change the id should be url encoded and the proper image should be loaded
- prior to this change an incorrect cover could load.
- all other covers behave normally.


lathomas64:
@LiYanjun19 @JonahCWilson any reason y'all can see to not urlencode ids for book covers?

JonahCWilson:
Looks fine to me. Out of scope, but maybe it'd be a good idea to add some code to replace whitespace with underscores?  Working with files with spaces in the name is such a pain in any shell.

mdnoble73 (approved):

PR #4186, opened 2026-04-28
DIS-2174: Minimize reading history sent back to Koha.

JonahCWilson:
Right now, when Koha is getting all checkouts, it returns a bunch of extra data which isn't always needed.  Moved that behind an "isNightlyUpdate" flag so that this information is sent less frequently.

@kidclamp Feel free to add anything you see fit.

### DIFF
--- a/code/web/sys/Account/User.php
+++ b/code/web/sys/Account/User.php
@@ -1857,9 +1857,10 @@ class User extends DataObject {
 	 *
 	 * @param bool $includeLinkedUsers
 	 * @param string $source
+	 * @param bool $isNightlyUpdate
 	 * @return Checkout[]
 	 */
-	public function getCheckouts(bool $includeLinkedUsers = true, string $source = 'all'): array {
+	public function getCheckouts(bool $includeLinkedUsers = true, string $source = 'all', bool $isNightlyUpdate): array {
 		require_once ROOT_DIR . '/sys/User/Checkout.php';
 
 		$checkoutsToReturn = [];
@@ -1868,6 +1869,7 @@ class User extends DataObject {
 		global $offlineMode;
 		if ($this->hasIlsConnection() && !$offlineMode) {
 			if ($source == 'all' || $source == 'ils') {
+				$this->getCatalogDriver()->driver->setNightlyUpdateStatus($isNightlyUpdate);
 				$ilsCheckouts = $this->getCatalogDriver()->getCheckouts($this);
 				$checkoutsToReturn = $ilsCheckouts;
 				$timer->logTime("Loaded transactions from catalog. $this->id");
--- a/code/web/CatalogConnection.php
+++ b/code/web/CatalogConnection.php
@@ -1250,7 +1250,7 @@ class CatalogConnection {
 			}
 		}
 
-		$checkouts = $patron->getCheckouts(false);
+		$checkouts = $patron->getCheckouts(false, isNightlyUpdate: $isNightlyUpdate);
 		foreach ($checkouts as $checkout) {
 			$source = $checkout->source;
 			$sourceId = $checkout->sourceId;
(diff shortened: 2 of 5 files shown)


mdnoble73 (changes requested):
This has a bunch of extra changes from other pull requests. Please update to only be your changes.

JonahCWilson:
@mdnoble73 My bad, target branch is now 26.05

code/web/sys/Account/User.php
```diff
@@ -1868,6 +1869,7 @@ public function getCheckouts(bool $includeLinkedUsers = true, string $source = '
 		global $offlineMode;
 		if ($this->hasIlsConnection() && !$offlineMode) {
 			if ($source == 'all' || $source == 'ils') {
+				$this->getCatalogDriver()->driver->setNightlyUpdateStatus($isNightlyUpdate);
```
mdnoble73, line 1872: Why set a property here rather than passing it through to $this->getCatalogDriver()->getCheckouts()? That will be easier for reading and debugging.

mdnoble73 (changes requested):

code/web/sys/Account/User.php
JonahCWilson: Something more along the lines of this?

https://github.com/bywatersolutions/aspen-discovery-dev/commit/135cbd0dd8888d7593ab5d162023758eb944e009

mdnoble73: Yes, but rather than a generic options array use explicit parameter names. That better matches the existing code style.

JonahCWilson: I think there's an argument to be made to keep the generic options array.
1. No matter how much the internal code structure of any one function changes, the signature can remain static
2. If things deviate greatly:
    - Class A has properties (a, b, c, d)
    - Class B has properties (w, x, y, z)
    Then we could create helper objects and cast the array. Even if it's a deeply nested set of properties we can use json_decode or ArrayObject
3. There are multiple instances throughout the Aspen codebase where a function is called in the manner `$this->driver->someFunc($params)`.  If we change these to use explicit parameters, and they grow to differ in the future, the code will inevitably become:
```
if($this->driver == 'A'){
    $this->driver->someFunc(paramA, paramB, paramC);
} elseif ($this->driver == 'B'){
    $this->driver->someFunc(paramA, paramB);
} elseif($this->driver == 'C'){
    $this->driver->someFunc(paramA, paramB, paramC, paramD, paramE);
}
```
I feel it makes more sense to keep it as an options array to allow for maximum abstracting.

As this is a collaborative, open-source community where no one person's opinion takes priority, why don't we solicit feedback from other developers and I'll update based on consensus?

mdnoble73: We definitely could have optional an optional params array, but then they can't be typed (unless we move to a class which is a lot of increased complexity if they didn't have the same parameters). In general these types of functions will all need to think about the same parameters even if they don't initially or if some are ignored for particular ILSs. For example, the same short circuiting that you added for Koha definitely could be done for all ILSs to minimize the amount of time needed for all ILSs to load checkouts to process reading history.

mdnoble73:
This needs Jira updated and conflicts must be resolved in addition to other inline changes.

code/web/sys/Account/User.php
kidclamp: I think for simple options we could have a convention that the 'options' array is only booleans, or strings if we feel we need more than 2 options generally. 

At least in my quick glance, the other indexers retrieved renewability in a single call, it didn't require extra options, it felt like only Koha needed this for now. 

Using the options array allows for any one ILS that requires a specific parameter to have it added without touching other files and reducing possible errors or conflicts. i.e. touch a few extra files now, and avoid touching them in the future. If we explain the options in the documentation for each ILS it makes it easy to understand

PR #4201, opened 2026-05-04
DIS-2355 ISBN Advanced Searches

JonahCWilson:
Fixed a bug with Advanced Search returning a Solr Error whenever ISBN search is done with any a dash (or any other special character) provided in the search field.

Reproduction Instructions:
https://aspen-discovery.atlassian.net/browse/DIS-2355

@mdnoble73 Please review the two flows mentioned in the Jira, I'm not sure which, if either of these, is expected behavior.  Scenario 2 is definitely wrong, not sure about #1.


mdnoble73 (changes requested):
In the first case, we do general checking of ISBNs looking for 10 and 13 digit ISBNs we have not done any checking of hypens within the string. That would be a new enhancement. 

For 2 we’ve added general validation of reserved solr characters for the standard searching, but that has not been extended to advanced searching. That would be a different issue than the first and would need to handle all the special cases.

PR #4202, opened 2026-05-04
DIS-2228: Expand Holiday table 

LiYanjun19:
Currently the Holidays table allows libraries to indicate they are closed on a specific date, but libraries sometimes have early closings or other special hours, rather than being totally closed.

The Holidays table could incorporate settings similar to the Hours table, so that rather than any date in the table automatically displaying as closed, a library would have the option to check a box for "Closed" or enter specific hours for that date.

The Holidays table would also have a locations column, so that each entry applies only to the selected locations. And add "Show in Holidays & Special Hours table" to Locations to control which locations participate in the Holiday and Special Hours table.

As part of this change, the Holidays table could also be renamed as "Holidays and Special Hours".

To test:
1. Confirm there are existing Holiday rows in the Library Systems.
2. Apply PR and run DB updates.
3. Verify existing holiday data migrates correctly to new Holidays and Special Hours table, with "closed" checked and all eligible locations selected by default.
4. Click Add New and verify the new row defaults to all valid locations.
5. Add new rows with different dates and different locations, save, and see the changes are saved.
6. Add new rows with the same dates but different locations, save, and see the changes are saved.
7. Add new rows with the same dates and same locations, save, and see the duplicate entries are not saved.
8. Disable Locations > Show in Holidays & Special Hours table for a location, then go back to Library and see all the holiday rows that are removed for that location.
9. Delete a location and see all the holiday rows that are removed for that location.
10. Log in as a patron and verify that holiday closures and special hours are reflected in library hours messages.
11. Create an invalid holiday row with "closed" unchecked and blank Opening/Closing hours, then verify patron-facing hours fall back to normal location hours.
12. In Koha library, set repeated holidays and special holidays, and verify that the holidays import into Aspen correctly 
13. In Sierra library, use Batch Update Holidays and verify the holidays are parsed and applied correctly.


mdnoble73 (changes requested):
This looks like the new settings are related to a specific location, why are the settings all at the Library level? They seem to belong more at the location level.

mdnoble73 (changes requested):
This also does not seem to handle showing the new information to the user.

mdnoble73 (approved):

PR #4215, opened 2026-05-05
DIS-2294 preserve search type for series and author searches

mdnoble73:


code/web/sys/SearchObject/BaseSearcher.php
JonahCWilson, line 548: Suggestion:
```
// Move SearchTerms into array.  More extendable in the future if you need to add more
$searchTerms = ['basic', 'ebsco_eds', 'summon', 'gale', 'series', 'author'];

// Both if and else use the same format, we can abstract out the logic for the first part of the url string
$urlVal = in_array($this->searchType, $searchTerms) ? "searchIndex=" : "type=";
$params[] = $urlVal . urlencode($this->searchTerms[0]['index']);
```

PR #4232, opened 2026-05-07
DIS-2174: Minimize reading history sent back to Koha. 

JonahCWilson:


code/web/sys/Account/User.php
```diff
@@ -1920,7 +1920,7 @@ public function getCheckouts(bool $includeLinkedUsers = true, string $source = '
 		global $offlineMode;
 		if ($this->hasIlsConnection() && !$offlineMode) {
 			if ($source == 'all' || $source == 'ils') {
-				$ilsCheckouts = $this->getCatalogDriver()->getCheckouts($this);
+				$ilsCheckouts = $this->getCatalogDriver()->getCheckouts($this, false, [ "isNightlyUpdate" => $isNightlyUpdate ]);
```
mdnoble73, line 1923: This is passing 3 parameters rather than 2 as the signature expects.

code/web/Drivers/SirsiDynixROA.php
```diff
@@ -1048,13 +1048,17 @@ protected function staffLoginViaWebService($username, $password) {
...
+	 * This is responsible for retrieving all checkouts (i.e. checked out items)
+	 * by a specific patron.
+	 *
+	 * @param User $patron       The user to load transactions for
+	 * @param array $options     Additional options
+	 * @return Checkout[]        Array of the patron's transactions on success
+	 * @access public
 	 */
-	public function getCheckouts(User $patron, int $page = 1, int $recordsPerPage = -1, string $sortOption = 'dueDate'): array {
+	public function getCheckouts(User $patron, array $options = []): array {
```
mdnoble73, line 1061: This is not providing the default values, nor does it look like the options array is being used now

mdnoble73 (changes requested):
See inline issues changes.

code/web/Drivers/SirsiDynixROA.php
JonahCWilson: In this function, `$page`, `$recordsPerPage`, and `$sortOption` are not being used at all, so they are superfluous and can be removed. `$options` is being added to match the new signature, but doesn't necessarily need to be used in this function.

code/web/sys/Account/User.php
JonahCWilson: Updated the CatalogConnection `getCheckouts` function to use options array which is then passed down to the relevant driver.

JonahCWilson:
@mdnoble73 Made requested changes, please take a look.  I can move this to 26.06 too, if need be.

code/web/release_notes/26.05.00.MD
```diff
@@ -117,6 +117,7 @@
 ### Koha Updates
 - Fix Koha password recovery error mapping to display resetPin error messages. ([DIS-2339](https://aspen-discovery.atlassian.net/browse/DIS-2339)) (*YL*)
 - Added the ability to see and create hyperholds in Aspen for those using Koha as their ILS and with this system preference enabled in Koha. ([DIS-1191](https://aspen-discovery.atlassian.net/browse/DIS-1191)) (*AB*)
+- Add functionality for Koha to only send back relevant data on nightly update (([DIS-2174]) https://aspen-discovery.atlassian.net/browse/DIS-2174) (*NC*)
```
mdnoble73: @JonahCWilson this doesn't follow the correct format for the link. Should be 
([DIS-2174])https://aspen-discovery.atlassian.net/browse/DIS-2174))

Also should this be attributed to you or Nick?

mdnoble73 (changes requested):

code/web/release_notes/26.05.00.MD
JonahCWilson: @mdnoble73 Fixed. It's better to attribute it to Nick. The majority of the work was done by him, I just added some final touch-ups.

mdnoble73 (approved):

PR #4242, opened 2026-05-08
DIS-2374

kidclamp:
This consolidates the code, but ultimately preserves existing behavior - I think we need some comments in the code, or possibly a setting for this to make it clear

### DIFF
--- a/code/web/Drivers/Sierra.php
+++ b/code/web/Drivers/Sierra.php
@@ -499,12 +499,13 @@ class Sierra extends AbstractIlsDriver {
 	}
 
 	/**
-	 * Retrieves valid pickup locations for this patron for this record.
+	 * Returns an array of valid pickup locations for this patron for this record.
 	 * @param string $recordId
 	 * @param User $patron
+	 * @param array $locations
 	 * @return array An array containing valid pickup locations
 	 */
-	public function getValidPickupLocationsForRecordFromILS($recordId, $patron): array {
+	public function getValidPickupLocationsForRecordFromILS($recordId, $patron, $locationSingleton): array {
 		if ($recordId == null) {
 			return [
 				'success' => false,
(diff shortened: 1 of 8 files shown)


code/web/Drivers/Sierra.php
```diff
@@ -520,29 +519,24 @@ public function getValidPickupLocationsForRecordFromILS($recordId, $patron): arr
...
+				$locationCodesFromIls[] = $pickupLocationsResponse->holdshelf->selected->code;
 			}
 			foreach ($pickupLocationsResponse->holdshelf->locations as $location) {
-				$locationCodes[] = trim($location->code);
+				$locationCodesFromIls[] = trim($location->code);
+			}
+			$pickupLocations = [];
+			if (!empty($locationCodesFromIls)) {
+				$pickup_filter = function($locations) use ($locationCodesFromIls){
+					return array_any($locationCodesFromIls, fn($validCode) => str_starts_with($locations->code, $validCode));
```
JonahCWilson: Should this be `str_starts_with` or `strcasecmp`?

kidclamp: @mdnoble73 said this is intentional, so restored this and asked for more info on the Jira

code/web/sys/LibraryLocation/Location.php
```diff
@@ -1676,32 +1680,30 @@ function getPickupBranches(User|bool|null $patronProfile, bool $isLinkedUser = f
 			if (!empty($patronProfile)) {
 				$tmpLocation->addPickupUser($patronProfile->id);
 			}
-			if (($tmpLocation->validHoldPickupBranch == 1) || ($tmpLocation->validHoldPickupBranch == 0 && !empty($patronProfile) && $patronProfile->homeLocationId == $tmpLocation->locationId) || ($tmpLocation->validHoldPickupBranch == 3 && !empty($patronProfile) && $patronProfile->getHomeLibrary()->libraryId == $tmpLocation->libraryId)) {
```
mdnoble73, line 1645: @kidclamp why does this remove checking the validity of if the user wants the pickup location to be active?

code/web/services/Record/AJAX.php
```diff
@@ -2047,24 +2047,7 @@ private function setupHoldForm(string $recordSource, ?bool &$rememberHoldPickupL
 
 		$catalogDriver = $user->getCatalogDriver();
 		if (!empty($catalogDriver) && $catalogDriver->restrictValidPickupLocationsForRecordByILS()) {
-			$getPickupLocationsFromILS = $catalogDriver->getValidPickupLocationsForRecordFromILS($marcRecord->getUniqueID(), $user);
```
mdnoble73, line 1897: @kidclamp is the root issue here that we aren't filtering the list of location codes from Aspen properly based on the return from the ILS?

mdnoble73: Also just verifying that this logic was moved to Sierra.php from AJAX.php and UserAPI.php

code/web/sys/LibraryLocation/Location.php
kidclamp: Removed this bit

code/web/services/Record/AJAX.php
kidclamp: yes on both counts

code/web/Drivers/Sierra.php
kidclamp: new patches, added a system variable

PR #4244, opened 2026-05-13
DIS-2378 more like this arrows

JonahCWilson:
Testing Instructions

Prerequisites:

26.05 Instance
Record with multiple “More Like This” items that would normally generate arrows
Record with fewer “More Like This” items that would NOT generate arrows.

To Reproduce:
1. Load the record with multiple “More Like This” items and notice that the arrows do not appear.
2. Zoom in or out and arrows appear due to a refresh event
3. Now you can navigate.

Apply Patch:
1. Load Record with multiple “More Like this” items and notice the arrows do appear.
2. Click either way, and you’re able to navigate
3. Load a record with fewer “More Like This” and notice that the arrows do not appear.


JonahCWilson:
@mdnoble73 I know it's super early to review this, but I wanted to ask early. Assuming it's good, we were going to backport it into our 26.05 release for a couple of partners.

JonahCWilson:
@mdnoble73 Requesting review.

mdnoble73:
I'll try to take a look today

code/web/interface/themes/responsive/js/aspen/grouped-work.js
mdnoble73, line 1: This is attempting to iterate through similarTitleData.titles even if the similarTitleData is null which will cause errors in this case. 

The important part of the fix is making sure the jcarousel is created after the html is set.

mdnoble73 (changes requested):

code/web/interface/themes/responsive/js/aspen/grouped-work.js
JonahCWilson: @mdnoble73 

Good call, I didn't see that the first time through.

The output object from `AJAX.php` is as follows:
```
$enrichmentResult = [
	'similarTitles' => [
		'titles' => [],
	],
];
```
I was operating under the assumption that I would probably have that `data.similarTitles` structure, but I agree with you that it's better to be defensive.

Since the original code also makes assumptions about the data (assumes that `similarTitleData.titles` exists, which would cause an error if `.length` is called on null), I went ahead and refactored a bit to add defensive validation so we know for a fact that the data we're working with has correct structure.

While I was in there, I also cleaned up the old try/catch flow since it was masking errors and relying on carousel state that could become invalid after the DOM update.

mdnoble73 (approved):

PR #4245, opened 2026-05-13
DIS-2337: adding support for vendor login link

lathomas64:
https://aspen-discovery.atlassian.net/browse/DIS-2337
- [x] Created a Jira Issue for the change
- [x] Include a description in the Jira issue
- [x] Add Testing notes to Jira Issue
- [x] Include the testing notes in the pull request
- [x] pull request title should follow the template DIS-### Description
	- This ensures that JIRA recognizes the PR
- [x] pull request should include release notes
- [ ] Jira also needs to be updated with developer information and the ticket should be in Code Review status.

---
- run database updates
- enable the single sign-on module (/Admin/Modules)
- add an IP entry for your current IP with the allow vendor single sign-on checkbox checked (/Admin/IPAddresses)
- create or identify single sign on settings you wish to use for the vendor sign on link (/Admin/SSOSettings)
- make sure the service field on the settings is SAML 2 or OAUTH 2.0
- add a Vendor section to your config file with a line SSOSettingID equal to the id of the setting you wish to use
- log out and on the home screen if you are on the correct ip you should see the Vendor Login link in the bottom left.
- If you have a staff login link you should see both with the Vendor Login link below the Staff one.
- Set up a staff login with different SSO Settings and ensure both work simultaneously
- turn off allow vendor single sign-on checkbox for your current IP
- confirm that the vendor login link no longer apperas
- attempt to navigate directly to the url the link pointed to (/init_oauth.php?vendor OR /Authentication/SAML2?init&vendor depending on if you were using oauth or saml)
- You should get redirected to /Search/Home and an error line in messages.log should have a note acknowledging the login attempt and what IP it came from. 
- if no settings exist for the path you chose the default behavior for on settings existed should happen instead


code/web/sys/Authentication/SSOSetting.php
```diff
@@ -947,6 +947,92 @@ static function getObjectStructure(string $context = ''): array {
...
+			$SSOSetting = new SSOSetting();
+			$SSOSetting->id = $library->ssoSettingId;
+			$SSOSetting->service = $service;
+			if($SSOSetting->find(true))
+			{
+				return $SSOSetting;
+			}
+			return false;
+		}
+	}
```
JonahCWilson, line 1034: These are all nits:
1. Please add type information for parameters `$vendor` and `service`
2. `if($vendor)` returns, so `else{}` isn't strictly necessary and can be pulled out.
3. Final return value can be shortened to `return $SSOSetting->find(true) ? $SSOSetting : false;`

code/web/sys/IP/IPAddress.php
```diff
@@ -602,6 +611,19 @@ public static function allowSSOAccessForClientIP() : bool {
...
+		//check $ipInfo to see if we are allowed to use vendor SSO
+		$clientIP = IPAddress::getClientIP();
+		$ipInfo = IPAddress::getIPAddressForIP($clientIP);
+		//if we don't have an IP don't allow access
+		if(empty($ipInfo))
+		{
+			return false;
+		}
+		//if we do know the IP determine if that IP is allowed
+		return $ipInfo->vendorSSOLogin;
```
JonahCWilson, line 624: Nit:
`return empty($ipInfo) ? false : $ipInfo->vendorSSOLogin`

Alternatively, I don't know about the truthiness of ipInfo object but you could possibly do:
`return $ipInfo ? $ipInfo->vendorSSOLogin : false;`

gmcharlt (changes requested):
Since it looks like part of activating this would be adding a `[Vendor]` section to config.ini with a key to identify the SSO service to use, I suggest adding a commented-out example to the template `config.ini`.

JonahCWilson (commented):
I forgot to hit "submit review" so these have been sitting here for months.

PR #4246, opened 2026-05-13
DIS-2381: Duration DIsplay

K-Alette:
Update duration display so that is shows in both vertical and horizontal display of formats 
Update release notes


code/web/services/GroupedWork/AJAX.php
```diff
@@ -2982,11 +2982,47 @@ public function getHorizDisplayFormatEdition () : array {
 
 			if ($foundVariation) {
 				$relatedRecords = $variation->getRelatedRecords();
+				if (count($relatedRecords) > 1) {
+					// Extract and filter out null/empty durations
+					$durations = array_values(array_filter(
+						array_map(fn($item) => $item->duration ?? null, $relatedRecords),
+						fn($d) => $d !== null && $d !== ''
+					));
```
JonahCWilson, line 2990: 1. Love to see the functional approach, very nice!
2. What values could $item->duration have?  If it could be null, `$item->duration ?? null` seems redundant, and it appears your filter function is filtering out both null and empty strings anyway.  Could it just be `$item->duration`?

JonahCWilson: Also, just a suggestion, but maybe move the contents of the `if (count($relatedRecords) > 1)` block into a helper function, something like `$this->assignDurations`. You could move the 3022-3025 block in there too.

mdnoble73 (approved):

PR #4253, opened 2026-05-15
DIS-2214: Add OAuth2 Server & OpenID Connect

catsoup11789:
**Implements the following grants:**
- Authorization Code with PCKE
- Resource Owner Password Credentials (ROPC)
- Client Credentials

**Includes support for:**
- Refreshing tokens
- Scope-based access control (minimally integrated for this initial implementation)
- Read or write access control per scope
- Rate limiting based on API
- PKCE compliance
- OpenID Connect (still very baby state, needs customization for Aspen)
- Manage OAuth2 clients & OpenID Connect clients in Aspen Admin

Based on The PHP League's OAuth2 Server library.

**To do:**
- Add OAuth event logging
- Add page in My Account for users to revoke any authenticated apps
- For the patron authorization page allow patron to reduce granted permissions from what was requested (provided they are still required to share the minimum data needed... likely first name, last name, and email?)

# Setup
Generate RSA keys
```bash
# Generate RSA private key (2048-bit)
openssl genpkey -algorithm RSA -out /path/to/aspen-discovery/code/web/sys/Authentication/OAuth2/private.key -pkeyopt rsa_keygen_bits:2048

# Extract public key from private key
openssl pkey -in /path/to/aspen-discovery/code/web/sys/Authentication/OAuth2/private.key -pubout -out /path/to/aspen-discovery/code/web/sys/Authentication/OAuth2/public.key

# Generate random encryption key (base64-encoded 32 bytes)
openssl rand -base64 32 > /path/to/aspen-discovery/code/web/sys/Authentication/OAuth2/encryption.key

chmod 600 /path/to/aspen-discovery/code/web/sys/Authentication/OAuth2/private.key
chmod 600 /path/to/aspen-discovery/code/web/sys/Authentication/OAuth2/public.key
chmod 600 /path/to/aspen-discovery/code/web/sys/Authentication/OAuth2/encryption.key

```

# Creating Clients
## OAuth2
Using a user who has the Administer OAuth2 Clients permission, access the admin interface:
1. System Configuration > OAuth2 Clients
2. Add New Client
3. Complete required fields:
   - **Client Name**: Should be descriptive, human-readable name for this client so you know what its going to be used for (you can't change this later)
   - **Client Type**: Choose based on your use case (see the grant types below, you can't change this later)
   - **Scopes**: Select minimum required scopes
   - **Redirect URI**: For web apps only, the URL where Aspen will send the user after they authorize the client to use their credentials. This is required for Authorization Code grants and must match the `redirect_url` value that the client sends in their authorization requests.
4. Save

At save, Aspen will auto-generate Client ID and Client Secret, which are needed for actually authorizing.

Note: The scoping still needs to be deeply integrated into the APIs, I have as an example started adding them into List API.

## OpenID Connect
Using a user who has the Administer OpenID Connect permission, access the admin interface:
1. System Configuration > OpenID Connect (OIDC) Clients
2. Add New Client
3. Complete required fields:
   - **Client Name**: Should be descriptive, human-readable name for this client so you know what its going to be used for (you can't change this later)
   - **Claims**: Select minimum required claims available to this client
   - **Redirect URI**: The URL where Aspen will send the user after they authorize the client to use their credentials and must match the `redirect_url` value that the client sends in their authorization requests.
4. Save

At save, Aspen will auto-generate Client ID and Client Secret, which are needed for actually authorizing.

### Authorization Code (Web Application)

*Best for:*
Traditional websites and web applications with a server-side backend that need to access patron data and/or make requests on behalf of a patron (update a list, view holds, etc). With this grant the client never handles or sees patron login info but can make updates to the patron's data (i.e. create a new list or add a new item to a list) if it has permissions to write.

This is also the method used for OpenID Connect. 

*How it works:*
- Patron logs into your website (not Aspen)
- Your backend server (again, not Aspen) makes a call to Aspen for authentication
- Patron gets redirected to Aspen, if they are not logged in they will be prompted to login to Aspen (it also has a safety check to make sure an existing Aspen session is tied to the user the external website is trying to authorize)
- After patron logs in, they are prompted to authorize the app to access the data requested (tied to API permissions when not OpenID, if OpenID it is tied to claims)
- If the patron authorizes, they are redirected back to the `redirect_uri` with an access code
- If the patron cancels the authorization, they are still redirected back to the `redirect_uri` but with an `error=access_denied` parameter and no access code
- Your backend server exchanges the code for tokens using the client secret
- Your website now has an access token to call Aspen APIs on behalf of the patron

Token expires after 1 hour, but refresh tokens last 1 month

```bash
https://aspen.yourlibrary.com/Authentication/OAuth2/Authorize?
  response_type=code&
  client_id=aspen_xxxxx&
  redirect_uri=https://myclient.com/callback&
  scope=list:read&
  state=random123&
  code_challenge=....&
  code_challenge_method=S256

curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -d "grant_type=authorization_code&code=CODE&client_id=aspen_xxxxx&client_secret=SECRET&redirect_uri=https://myclient.com/callback"
```

Supports `S256` (recommended) and `plain` for `code_challenge_method`

```bash
curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -H "Authorization: Basic YXNmZ...." \
  -d "grant_type=authorization_code&code=CODE&client_id=aspen_xxxxx&client_secret=SECRET&redirect_uri=https://myclient.com/callback"
```

**Note:** `redirect_uri` must match the value set on the client in Aspen Admin, and it is required for this grant.

### Client Credentials (Service-to-Service)

*Best for:*
Backend services, system-to-system, scheduled jobs, or third-party systems. Probably the most common and for general use to "read" data but not "write" it, i.e. you have a WordPress plugin that you want to connect to the List API to display public lists on your website. If you need to "write" to Aspen for updating something on behalf of a patron, I would suggest the Authorization Code grant as it would allow the client to securely make those requests without ever handling the patron login credentials.

*How it works:*
- Your backend service has a client id and client secret that an admin created in Aspen
- Service makes a token request using those creds
- Aspen returns an access token
- Service can use this token to call APIs which persists for 4 hours (these requests are NOT made on behalf of a patron)

Token expires after 4 hours

```bash
curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -H "Authorization: Basic YXNmZ...." \
  -d "grant_type=client_credentials&scope=work:read+list:read"
```

### Resource Owner Password Grant (Mobile Apps)

*Best for:*
Trusted apps like Aspen LiDA or other internal Aspen uses and never third-party apps as the client will have direct access to patron credentials which can raise security concerns.

We might want to add more security around creating clients with this grant, but I'm not sure. The specific use-case for a Password Grant is that there is no `redirect_uri` for the client.

*How it works:*
- Patron enters their barcode/pin in your app
- App sends those creds along with their client id and client secret
- Aspen returns an access token and refresh token
- App stores tokens securely and can now use the access token to call Aspen APIs on behalf of the patron

Token expires after 1 hour, but refresh tokens last 1 month

```bash
curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -d "grant_type=password&username=user&password=pass&client_id=aspen_xxxxx&client_secret=SECRET&scope=work:read"
```

```bash
curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -H "Authorization: Basic YXNmZ...." \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d "grant_type=passwords&username=12334566&passowrd=8sdfsdf&scope=work:read"
```

## Refreshing the token
```bash
curl -X POST https://aspen.yourlibrary.com/Authentication/OAuth2/Token \
  -d "grant_type=refresh_token&refresh_token=REFRESH_TOKEN&client_id=aspen_xxxxx&client_secret=SECRET"
```

## Using the token in API requests
```bash
curl -X GET "https://aspen.yourlibrary.com/API/ListAPI?method=getLibraryLists" \
  -H "Authorization: Bearer eyJ0eXAi..."
```

## Rate limiting
Uses a sliding window rate limiting approach by tracking requests per endpoint and either per client ID or IP address.

```https://aspen.yourlibrary.com/Authentication/OAuth2/Token``` allows 10 requests in a 1 minute window to prevent brute force token exchange attempts

```https://aspen.yourlibrary.com/Authentication/OAuth2/Authorize``` allows 20 requests in a 5 minute window to prevent brute force authorization attempts

```https://aspen.yourlibrary.com/API/*``` allows 1000 requests in a 60 minute window, this is default to all APIs using Abstract API, but each API can be configured separately in `OAuth2RateLimiter.php` or we can remove the fallback rate limit entirely if we only want to restrict on the authentication endpoints.

When the limit is exceeded the client will get this response (HTTP 429):
``` 
{
   "error": "rate_limit_exceeded",
   "error_description": "Rate limited exceeded. Try again in 45 seconds.",
   "retry_after": 45
}
```

### Monitoring rate limits
In Aspen Admin, there's a "OAuth Rate Limits" page to view real-time rate limit data to identify abusive clients, track window cycles, monitor activity patterns, troubleshoot rate limiting, and manual intervention for resetting the client.

Looks something like this:
```
Client ID: my_library_app
IP Address: 203.0.113.0
Endpoint: token
Request Count: 9 (out of 10 allowed)
Window Start: 2026-03-20 14:32:00
Last Request: 2026-03-20 14:32:47
```
### Privacy
**IP Addresses for the rate limiter are masked for anonymization** but not hashed because then it makes it impossible to actually track and identify.

# Integrating into APIs
- Use OAuth2Middleware with Abstract API to authenticate requests and manage scoping using PHP docblock annotations to help maintainability instead of having multiple sources of truth to update when adding new endpoints, changing permissions, etc.
- Ideally, we'd drop support for `@token` after LiDA switches to `@oauth` and only use IP-based access for Aspen-to-Aspen requests (i.e. to an external Greenhouse server connected to many Aspen instances), and determine what a public (no auth, no IP restrictions) endpoint looks like and where it makes sense to provide them.

## Using PHP docblock annotations
- `@oauth (true/false)` indicates whether or not a method accepts OAuth2 Bearer token authentication.
- `@token (true/false)` indicates whether a method accepts the traditional username/password in POST server in conjunction with matching tokens from Greenhouse. Currently only for LiDA requests.
- `@public (true/false)` indicates whether a method allows public access without any authentication (open, free to the world, use sparingly). You can also enable/disable the the IP Address whitelisting (using the Allow API Access setting in IP Addresses) here.
- `@scopes (comma-separated list)` declares what OAuth2 scopes are required to access this method

Example:
```php

/** 
* Add item to list
* @oauth true
* @scopes list:read, list:write
* /

```

If a method does not have docblock annotations, or are incomplete, the following values are inferred by default:

```php

/** 
* Add item to list
* @oauth true
* @token true
* @public false
* @scopes [inferred from method name]
* /

```

For scopes by default it will look at the method name, for example with the List API:

Scope prefix: `'list'`

```
get*,list*,search*,find*,load*,retrieve*,show*,view* = list:read
create*,add*,edit*,delete*,update*,remove*,clear*,set* = list:write
anything else: list:read (fallback)
```

## Adding authentication into launch() function of APIs
Use `$this->handleAPIRequestAuto($_GET['method'], 'api_name')` to automatically discover auth requirements set in docblock annotations. The first parameter will obviously tell the API which method to use and will grab the docblock annotations for that to validate the request. The second parameter is used for rate limiting.

You can also use `$this->handleAPIRequest()` which will have a similar setup to how it works today with arrays for valid methods and will not look at docblock annotations. I included this approach as an example in User API.
```php
function launch() {
$method = $_GET['method'] ?? '';
$oauthMethods = ['getWorks', 'getUserInfo'];
$tokenMethods = ['getUserInfo', 'getList'];
$publicMethods = ['getSearchableLists'];

$this->handleAPIRequest($method, $oauthMethods, $tokenMethods, $publicMethods, 'api_name');
}
```

## Handling delegated API requests for user actions
In the already well used `$this->getUserForAPICall()` throughout the APIs, there is now a check for Password Grant OAuth2 authentication which will return an Aspen user object that was validated when the token was created. No additional work is needed here.

## Available Scopes
This is just what I initially setup based on the types of endpoints already established, can remove or extend as needed:

```
'work:read' => 'Work API Read Access',
'user:read' => 'User API Read Access',
'user:write' => 'User API Write Access',
'list:read' => 'List API Read Access',
'list:write' => 'List API Write Access',
'item:read' => 'Item API Read Access',
'event:read' => 'Event API Read Access',
'event:write' => 'Event API Write Access',
'search:read' => 'Search API Read Access',
'system:read' => 'System API Read Access',
'fine:read' => 'Fine API Read Access',
'community:read' => 'Community API Read Access',
'community:write' => 'Community API Write Access',
```

### Scope validation
This validates scopes in a couple ways:
1. Make sure the scope(s) included are valid as per OAuth2Client::getScopeOptions()
2. Make sure the scope(s) included are allowed for the client
3. Determine what scope is required to make the request and if the client has that scope

If any of these are ever "no" Aspen will return response for Missing required scope

# Testing
I included a Postman collection that allows you to test the different authentication flows and accessing a couple APIs, as well as OpenID features.


Jacobomara901:
Hiya, as Mark requested, I've taken a look at the above and made a few notes. No particular order, just noted things as I spotted them. That being said, I did stop QAing as I couldn't get aspects of it working (as per the bottom of my notes!) 
Let me know if any of my comments miss the mark or I've mis-interpreted at all!

Notes:
- In `Logger.php`, you have your log path hardcoded and committed.

- `CGIPassAuth On` needs to be added to the apache configs for docker as it uses php-fpm. Authentication headers will not be passed through otherwise.

- There's some logging that I think should be removed. For instance you have logging on the POST for `services/Authentication/OAuth2_Authorize.php`:

  ```php
  $logger->log("[OAuth2] OAuth2_Authorize - REQUEST METHOD: " . $_SERVER['REQUEST_METHOD'], Logger::LOG_DEBUG);
  $logger->log("[OAuth2] OAuth2_Authorize - GET params: " . json_encode($_GET), Logger::LOG_DEBUG);
  $logger->log("[OAuth2] OAuth2_Authorize - POST params: " . json_encode($_POST), Logger::LOG_DEBUG);
  ```

  Another instance: `sys/Authentication/OAuth2/Repositories/ClientRepository.php` logs out 20 chars of the secret and should probably also be gone.

  There's a couple more, probably just needs a quick grep through the diff.

- `Add OAuth2 Server` db update is rather large and might want to be split out into multiple updates instead of doing all the creations in one.

- In `sys/Authentication/OAuth2/OAuth2Client.php` you declare `protected $supports_openid;` and `protected $allowed_claims;`, but neither column is in the `oauth2_clients` migration in `sys/DBMaintenance/version_updates/26.06.00.php`. `allowed_claims` only exists on `oauth2_openid_clients`. So at runtime those properties stay null on a fetched `OAuth2Client`.

  `services/Authentication/OAuth2_UserInfo.php` does `new OAuth2Client()` and then gates on `(bool)$client->supports_openid` (lines 89, 101). Because the column doesn't exist, that's always false and the endpoint returns 403 — even for clients you'd flag for OIDC.

  It's also looking in the wrong place: OIDC clients live in `oauth2_openid_clients`, not `oauth2_clients`. So the `$client->find(true)` would fail for them regardless. UserInfo is broken both ways — wrong table and missing column.

  I haven't actually booted up the OIDC stuff but this is purely from code review.

- I love that you've introduced foreign keys, I spoke about the possibility of this with Mark on my original PR, I would add one more in the definitions for `oauth2_refresh_tokens` linking to the `oauth2_access_tokens` so that if the access token is removed, so are any session tokens.

- For `exchangePasswordForToken()` you default to (`catalog:read catalog:write`). These don't exist in `OAuth2Client::getScopeOptions()` so the default will fail unless an explicit scope is passed with the request.

- In some tables you're using `created_at`, in others you're using `created_date`. Should these be matching?

- In `AuthenticationAPI::exchangePasswordForToken` you try to validate the client secret with `password_verify()`, but the secret is decrypted on fetch, not bcrypt-hashed. `password_verify` only matches bcrypt hashes, so that comparison can't work — the check should always return `invalid_client`.

  The method then calls `validateAccount` directly and `curl_exec`s the `/Authentication/OAuth2/Token` endpoint, which re-runs user validation through `UserRepository`. So every login does two credential checks and a loopback HTTP call.

  It also seems to be a dupe of `/Authentication/OAuth2/Token` with `grant_type=password` — same inputs, same outputs, just wrapped in the AuthenticationAPI's `{success: true, ...}`. Nothing else in the repo or the Postman collection calls it; everything goes through `/Authentication/OAuth2/Token` instead — is this old and should be removed?

- Should we be migrating the other APIs like we have `ListAPI`? Or is this for after?

- Composer is not handled for fresh installs, only upgrade. Should there be a separate PR for Composer install (since there is a DIS for it)? Will also need to be added to the dockerfile for the images as this will break the docker without it as autoloader will fail on start.

- Doesn't give you the secret until you go back in to edit the creds, might be worth adding an explanation of the behaviour?

- **OAuth2 vs OIDC:** I've had to do some reading here — I think I've wrapped my head around it: is OIDC a superset of OAuth2, and so an OIDC client could just be a normal OAuth2 client with the `openid` scope? I'm wondering if we need the two classes, or if we could simplify this?

  *(follow-up)* I've done some additional googling and since we don't want to reinvent the wheel and we use a library for OAuth, surely that applies to OIDC as well? Can't we just use the same plugin for `oauth2-server` that VuFind uses? [`steverhoades/oauth2-openid-connect-server`](https://github.com/steverhoades/oauth2-openid-connect-server). We already know it's compatible as they're also using the same version of `league/oauth2-server` that we are.

- Permissions need to be swapped around as they currently have the requirements below:
  - `/Admin/OAuth2Clients` → needs `Administer OpenID Connect`
  - `/Admin/OpenIDClients` → needs `Administer OAuth2`

I've also given it a test as we'll be looking to test adding our Events endpoints on top of this soon. From a migration standpoint, looks like it shouldn't be too much work at all to get it moved over, though I had some issues with getting requests on this branch working.

I used:

```bash
curl -v -X POST 'http://localhost:8083/Authentication/OAuth2/Token' \
    -d 'grant_type=client_credentials' \
    -d 'client_id=aspen_54198055a42ea2aa156f29354f128926' \
    -d 'client_secret=270caa9115e84bdf0326b18c0e7570acb0ad8be81f519429bbbd68549e58bfa1'
```
and couldn't get a token with client creds as there seem to be missing methods being referenced. Have you noticed the same in your testing?

I'm going to stop QAing here as there's no point in QAing if it's possibly not currently working as it definitely will end up with at least some sort of change if that's the case — happy to look again properly when you're ready!

code/web/bootstrap.php
```diff
@@ -1,6 +1,11 @@
 <?php
 define('ROOT_DIR', __DIR__);
 
+/**
```
mdnoble73, line 4: We should wrap this to handle cases where composer is not setup

global $composerActive;
if (file_exists(ROOT_DIR . '/vendor/autoload.php')) {
	$composerActive = true;
	require_once ROOT_DIR . '/vendor/autoload.php';
}else{
	$composerActive = false;
}

code/web/interface/themes/responsive/OAuth2/oauth2_error.tpl
mdnoble73, line 1: We don't necessarily need to wrap this in Aspen? Could just be (not sure though if anyone is likely to see this)
{strip}
<html lang="en">
	<head>
		<title>{$title}</title>
	</head>
	<body>
		<div class="container" style="max-width: 450px;">
			<div class="row">
				<div class="col-xs-12">
					<div class="text-center">
						<h1>{translate text="Authorization Error" isPublicFacing=true}</h1>
					</div>
					<div class="alert alert-danger">
						<strong>{translate text="Error:" isPublicFacing=true}</strong> {$error}<br>
						{$errorDescription}
					</div>

					<p>
						<a href="/" class="btn btn-default">
							{translate text="Cancel" isPublicFacing=true}
						</a>
					</p>
				</div>
			</div>
		</div>
	</body>
</html>
{/strip}

code/web/services/Admin/OAuth2Clients.php
```diff
@@ -0,0 +1,99 @@
...
+	function getBreadcrumbs(): array {
+		$breadcrumbs = [];
+		$breadcrumbs[] = new Breadcrumb('/Admin/Home', 'Administration Home');
+		$breadcrumbs[] = new Breadcrumb('/Admin/Home#system_admin', 'System Administration');
+		$breadcrumbs[] = new Breadcrumb('/Admin/OAuth2Clients', 'OAuth2 Clients');
+		return $breadcrumbs;
+	}
+
+	function getActiveAdminSection(): string {
+		return 'admin';
```
mdnoble73: This should be system_admin

code/web/services/Admin/OAuth2Clients.php
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	function getViewPermissions(): array {
+		return ['Administer OpenID Connect'];
```
mdnoble73: Should be Administer OAuth

code/web/services/Admin/OpenIDClients.php
```diff
@@ -0,0 +1,83 @@
...
+	function getBreadcrumbs(): array {
+		$breadcrumbs = [];
+		$breadcrumbs[] = new Breadcrumb('/Admin/Home', 'Administration Home');
+		$breadcrumbs[] = new Breadcrumb('/Admin/Home#system_admin', 'System Administration');
+		$breadcrumbs[] = new Breadcrumb('/Admin/OpenIDClients', 'OpenID Connect (OIDC) Clients');
+		return $breadcrumbs;
+	}
+
+	function getActiveAdminSection(): string {
+		return 'admin';
```
mdnoble73: Should be system_admin

code/web/services/Admin/OpenIDClients.php
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	function getViewPermissions(): array {
+		return ['Administer OAuth2'];
```
mdnoble73, line 65: Should be Administer OpenID Connect

code/web/services/API/AbstractAPI.php
```diff
@@ -1,6 +1,11 @@
 <?php
 
+require_once ROOT_DIR . '/sys/Authentication/OAuth2/OAuth2Middleware.php';
```
mdnoble73, line 3: Wrap with 
global $composerActive;
if ($composerActive) {
	require_once ROOT_DIR . '/sys/Authentication/OAuth2/OAuth2Middleware.php';
}

code/web/services/API/AbstractAPI.php
```diff
@@ -103,17 +108,36 @@ function logPatronRequest($userId): void {
...
-		if ($this->_userForAPICall === null) {
-			[$username, $password] = $this->loadUsernameAndPassword();
-			$user = UserAccount::validateAccount($username, $password);
-			if ($user !== false && $user->source == 'admin') {
-				//Admin users are not allowed with API calls
-				$this->_userForAPICall = false;
-				return $this->_userForAPICall;
+	function getUserForApiCall() {
+		// Check if this is an OAuth2 authenticated request first
+		$oauthUser = OAuth2Middleware::getAuthenticatedUser();
```
mdnoble73, line 134: This should check if $composerActive

code/web/services/Authentication/OAuth2_Authorize.php
```diff
@@ -0,0 +1,272 @@
...
+			UserAccount::login($user);
+			return true;
+		}
+		return false;
+	}
+
+	/**
+	 * Handle OAuth server exceptions
+	 */
+	private function handleOAuthError(OAuthServerException $exception): void {
```
mdnoble73, line 248: If we don't want to wrap this in Aspen it could be: 
private function handleOAuthError(OAuthServerException $exception): void {
		global $interface;
		$interface->assign('title', 'Authorization Error');
		$interface->assign('error', $exception->getErrorType());
		$interface->assign('errorDescription', $exception->getMessage());
		http_response_code($exception->getHttpStatusCode());

		echo $interface->fetch(ROOT_DIR . '/interface/themes/responsive/OAuth2/oauth2_error.tpl');
	}

code/web/services/Authentication/OAuth2_Authorize.php
```diff
@@ -0,0 +1,272 @@
...
+		$interface->assign('errorDescription', $exception->getMessage());
+		http_response_code($exception->getHttpStatusCode());
+
+		$this->display('../OAuth2/oauth2_error.tpl', 'Authorization Error', false, true);
+	}
+
+	/**
+	 * Handle general exceptions
+	 */
+	private function handleGeneralError(Exception $exception): void {
```
mdnoble73, line 260: Update similar to the above if we don't want to wrap it in Aspen

code/web/sys/Authentication/OAuth2/Repositories/ScopeRepository.php
```diff
@@ -0,0 +1,101 @@
...
+require_once ROOT_DIR . '/sys/Authentication/OAuth2/Entities/OAuth2ScopeEntity.php';
+
+class ScopeRepository implements ScopeRepositoryInterface {
+
+	public function getScopeEntityByIdentifier($identifier): ?ScopeEntityInterface {
+		global $logger;
+		$logger->log("[OAuth2] ScopeRepository::getScopeEntityByIdentifier() - Looking up scope: " . $identifier, Logger::LOG_DEBUG);
+		
+		$validScopes = OAuth2Client::getScopeOptions();
+		$validClaims = OAuth2Client::getClaimsOptions();
```
mdnoble73: getClaimsOptions are not currently defined on OAuth2Client, but probably should be

install/upgrade_26.06.00.sh
mdnoble73, line 1: Should have executable bit set. We may also need to do a runuser on the composer commands so they don't run as root.

install/upgrade_debian_26.06.00.sh
mdnoble73, line 1: Should have executable bit set. We may also need to do a runuser on the composer commands so they don't run as root.

mdnoble73 (changes requested):

code/web/interface/themes/responsive/OAuth2/oauth2_error.tpl
catsoup11789: This is the template used when a user is trying to authorize a client and they provide invalid credentials (or some other error happens in the process).

mdnoble73 (approved):

PR #4257, opened 2026-05-18
DIS-2100: Add check for allowAppRequestLogging in API requests for LiDA

catsoup11789:
- Adds logPatronRequestExternal function to Abstract API that is used when returning $output from LiDA requests for each API
- Adds `debug` true if user has allowAppRequestLogging enabled (set in MyAccount/AppRequests/) which forces logging of the API request in Aspen LiDA's new API Error Log (as part of DIS-2100), which by default only logs requests with errors or malformed data

### DIFF
--- a/code/web/services/API/AbstractAPI.php
+++ b/code/web/services/API/AbstractAPI.php
@@ -101,6 +101,24 @@ abstract class AbstractAPI extends Action{
 		}
 	}
 
+	/**
+	 * Add debug output if user has enabled App Request Logging in Discovery
+	 * This forces responses to be logged into Aspen LiDA instead of just errors
+	 * @param mixed $result
+	 * @return mixed
+	 */
+	protected function logPatronRequestExternal(mixed $result): mixed {
+		$user = $this->getUserForApiCall();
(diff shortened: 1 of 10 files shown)


code/web/services/API/AbstractAPI.php
```diff
@@ -101,6 +101,24 @@ function logPatronRequest($userId): void {
...
 
+	/**
+	 * Add debug output if user has enabled App Request Logging in Discovery
+	 * This forces responses to be logged into Aspen LiDA instead of just errors
+	 * @param mixed $result
+	 * @return mixed
+	 */
+	protected function logPatronRequestExternal(mixed $result): mixed {
+		$user = $this->getUserForApiCall();
+		if ($user !== false && $user->allowAppRequestLogging) {
```
JonahCWilson, line 112: Nit:
`getUserForApiCall` returns `false` or a truthy object value, so `$user && $user->allowAppRequestLogging` is sufficient.

JonahCWilson:
Minor style nit, but everything looks good!

mdnoble73 (approved):

PR #4258, opened 2026-05-18
DIS-2411: Improve mechanism for detecting and setting a language

lucasmontoya13:
Aspen Discovery had no way to configure a default language for unauthenticated users. The fallback was hardcoded to 'en' in index.php.

Three improvements were implemented:

1. Admin-configurable default language:
Added an isDefault column to the languages table. The Languages admin page now shows a "Default Language" checkbox. Language::update() enforces single-default by clearing other rows when a new default is saved. Language::getDefaultLanguageCode() reads this flag and validates the result against active languages before returning it.

2. Browser language auto-detection
Added Language::getLanguageFromBrowser(), which parses the Accept-Language HTTP header on the user's first visit. It tries an exact tag match first, then falls back to primary subtag matching (e.g., es-AR matches es).

3. Persistent language cookie
Language selection is now stored in an aspenInterfaceLanguage cookie with a 3-year expiry. Both switch paths write to it: the navbar myLang param (in index.php) and the display settings modal AJAX call (JSON.php:updateDisplaySettings()).

Test plan

1. Admin default is respected on first visit
   - Set Spanish as default in Languages admin.
   - Open a new incognito window with `Accept-Language` set to a language not installed (e.g. French).
   - Navigate to the catalog as an unauthenticated user.
   - Expected: interface displays in Spanish.

2. Browser language auto-detection
   - Set English as default. Keep both English and Spanish installed.
   - Open a new incognito window with browser language set to Spanish (`es` or `es-AR`).
   - Navigate to the catalog.
   - Expected: interface displays in Spanish (browser preference matched an installed language).

3. Navbar language switch persists across page navigation
   - From the catalog, switch language via the navbar link (appends `?myLang=es`).
   - Navigate to another page without using `myLang`.
   - Expected: language is still Spanish (cookie was set).

4. Display settings modal language switch
   - Open the display settings modal and change the language, then click Update Settings.
   - Navigate to another page.
   - Expected: language selection is retained (AJAX handler wrote the cookie).

5. Cookie survives browser restart
   - Select a non-default language via either switch method.
   - Close and reopen the browser.
   - Navigate to the catalog.
   - Expected: previously selected language is still active.

6. Admin default change takes effect for new visitors
   - Change the default language in admin from Spanish to English.
   - Open a new incognito window with no matching `Accept-Language`.
   - Expected: interface displays in English.

7. Authenticated users are unaffected
   - Log in as a user with a saved language preference (e.g. French).
   - Expected: interface displays in French regardless of cookie or DB default.

### DIFF
--- a/code/web/index.php
+++ b/code/web/index.php
@@ -56,9 +56,17 @@ $interface->assign('cssJsCacheCounter', 47);
 global $language;
 global $serverName;
 //Get the active language
+require_once ROOT_DIR . '/sys/Translation/Language.php';
 $userLanguage = UserAccount::getUserInterfaceLanguage();
 if ($userLanguage == '') {
-	$language = strip_tags((isset($_SESSION['language'])) ? $_SESSION['language'] : 'en');
+	if (!empty($_COOKIE['aspenInterfaceLanguage'])) {
+		$language = strip_tags($_COOKIE['aspenInterfaceLanguage']);
+	} elseif (!empty($_SESSION['language'])) {
+		$language = strip_tags($_SESSION['language']);
+	} else {
+		$browserLanguage = Language::getLanguageFromBrowser();
+		$language = $browserLanguage ?: Language::getDefaultLanguageCode();
+	}
 } else {
 	$language = $userLanguage;
 }
@@ -71,7 +79,8 @@ if (isset($_REQUEST['myLang'])) {
 	}
 	if ($language != $newLanguage) {
 		$language = $newLanguage;
-		$_SESSION['language'] = $language;
+		setcookie('aspenInterfaceLanguage', $language, time() + (3 * 365 * 24 * 3600), '/');
+		$_COOKIE['aspenInterfaceLanguage'] = $language;
 		//Clear the preference cookie
 		if (isset($_COOKIE['searchPreferenceLanguage'])) {
 			//Clear the cookie when we change languages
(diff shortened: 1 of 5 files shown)


JonahCWilson:
This seems like a really nice feature, good work!

lucasmontoya13:
Thanks Jonah! I appreciate it!

code/web/index.php
```diff
@@ -56,9 +56,17 @@
...
 if ($userLanguage == '') {
-	$language = strip_tags((isset($_SESSION['language'])) ? $_SESSION['language'] : 'en');
+	if (isset($_COOKIE['aspenInterfaceLanguage'])) {
+		$language = strip_tags($_COOKIE['aspenInterfaceLanguage']);
+	} elseif (isset($_SESSION['language'])) {
+		$language = strip_tags($_SESSION['language']);
+	} else {
+		$browserLanguage = Language::getLanguageFromBrowser();
+		$language = $browserLanguage ?: Language::getDefaultLanguageCode();
+	}
```
JonahCWilson: Is there ever a case where the language value for `$_SESSION` or `$_COOKIE` would be the empty string? What happens in that scenario, and do we have any fallbacks for it?

lucasmontoya13: As you said, I wasn’t covering that case. There is a realistic scenario where the language value could be an empty string, which could happen by modifying the myLang query parameter, so thank you for pointing that out, Jonah! I’ve already fixed it by changing the method used from `isset` to `!empty` .

mdnoble73 (approved):

PR #4286, opened 2026-05-22
DIS-1723 ampersand covers

lathomas64:
https://aspen-discovery.atlassian.net/browse/DIS-1723

- load a record page with an ampersand in the book cover’s id (example: https://discover.knoxcountylibrary.org/oreilly/on1386275878?searchId=13820819 )
- With the code change the id should be url encoded and the proper image should be loaded
- prior to this change an incorrect cover could load.
- all other covers behave normally.
- covers of the style ils:1234 should be unaffected.


lathomas64:
@mdnoble73 changed that url_encode to a str_replace to avoid the ils:1234 problem

mdnoble73 (approved):

PR #4287, opened 2026-05-22
DIS-2228: Fix nested multiSelect handling for checkboxList

LiYanjun19:
DIS-2228: Fix nested multiSelect handling for checkboxList

### DIFF
--- a/code/web/sys/DataObjectUtil.php
+++ b/code/web/sys/DataObjectUtil.php
@@ -661,7 +661,8 @@ class DataObjectUtil {
 									'storedPassword'
 								])) {
 									$oldValue = $subObject->$subPropertyName;
-									$newValue = $_REQUEST[$requestKey][$id] ?? ($subProperty['type'] == 'multiSelect' ? [] : null);
+									$isCheckboxList = $subProperty['type'] == 'multiSelect' && ($subProperty['listStyle'] ?? '') == 'checkboxList';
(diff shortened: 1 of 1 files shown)


code/web/sys/DataObjectUtil.php
```diff
@@ -661,7 +661,8 @@ static function processProperty(DataObject $object, $property, $fieldLocks) : vo
 									'storedPassword'
 								])) {
 									$oldValue = $subObject->$subPropertyName;
-									$newValue = $_REQUEST[$requestKey][$id] ?? ($subProperty['type'] == 'multiSelect' ? [] : null);
+									$isCheckboxList = $subProperty['type'] == 'multiSelect' && ($subProperty['listStyle'] ?? '') == 'checkboxList';
```
JonahCWilson, line 664: The second half of this and conditional can be simplified:

`($subProperty['listStyle'] ?? '') == 'checkboxList';`

If `$subProperty['listStyle']` is `null`, or `'someOtherStringValue'`, then it'll still be false, so the `?? ''` is unnecessary.

We can just have `$subProperty['type'] == 'multiSelect' && $subProperty['listStyle'] == 'checkboxList'`;

LiYanjun19: Yes, but I wanted to handle the case where a multiSelect doesn't have listStyle

mdnoble73 (approved):

PR #4290, opened 2026-05-25
DIS-2368 Fix Go to Holds button not showing

stephenrwicks:
https://aspen-discovery.atlassian.net/browse/DIS-2368

"Go to Holds" in modal wasn't showing in cases where showMessage was hit instead of showMessageWithButtons

I simplify this showPlaceHold function and call showMessageWithButtons once. It looks like there are currently several branches in the function that do the same thing.


mdnoble73 (approved):
Nevermind, looks good

PR #4305, opened 2026-05-27
DIS-1145: MARC Privacy Indicators

JonahCWilson:
**Type of Feature:**
- Enhancement

**Testing Instructions:**
- Load the newly compiled reindexer.jar onto a test server.
- In ILS (I tested on Koha) find a record and update it with the following MARC fields:
  - 541 1 _ a|This data will show
  - 541 _ _ a|This data will also show
  - 541 0 _ a|This data will not show
- Repeat step 2 for fields 542, 561, and 583
- Run the koha_export.jar file for whichever singleWork you updated.
- When you view the record in Aspen, you should see the positive examples appearing, and the fields marked with 0 do not appear. 

**Jira Link**:
https://aspen-discovery.atlassian.net/browse/DIS-1145

I am admittedly not super familiar with this flow, so if there is anything missing or something I've forgotten, feel free to make suggestions for improvement!


mdnoble73 (changes requested):
This needs conflicts resolved. 

You can also not commit the jar files and I'll rebuild them all after a series of merges and before each beta/GA release.

mdnoble73 (approved):

PR #4312, opened 2026-05-28
DIS-2464 Normalize Item dueDate

LiYanjun19:
Where is it? modal shows incorrect due dates when item due dates uses YYYYMMDD format.

To test:
1. Set the item due date using YYYYMMDD format, and set Grouped Work Display Settings > Where Is It Display Style > Summary. 
2. See Where is it? shows incorrect due dates for the checked out items 
3. Apply PR
4. Reload the Where is it? modal and see the correct due dates 
5. Run tests on other ILSes with YYYY-MM-DD and MM-DD-YYYY formats, see due dates display correctly as expected.


mdnoble73 (changes requested):
@LiYanjun19 the due date format is defined within the Indexing Profile. That information should likely be used to determine the actual format for parsing the duedate.

mdnoble73 (approved):

PR #4313, opened 2026-05-28
DIS-2465: Sierra ILL item renewals 

LiYanjun19:
For Sierra libraries, ILL items are currently treated as renewable by default. Add a library setting that allows libraries to control whether ILL items should be renewable.

To Test:
1. Apply PR to a Sierra Library and run db update
2. Log in as a patron who has an ILL item checked out 
3. Confirm that the ILL item is renewable
4. Deselect Library Systems > Interlibrary loans > Allow ILL Renewals
5. Confirm that the ILL item is not renewable

### DIFF
--- a/code/web/Drivers/Sierra.php
+++ b/code/web/Drivers/Sierra.php
@@ -846,9 +846,7 @@ class Sierra extends AbstractIlsDriver {
 					}
 					$curCheckout->formats = ['Unknown'];
 					$homeLibrary = $patron->getHomeLibrary();
-					if ($homeLibrary !== null && !$homeLibrary->allowToRenewILL) {
-						$curCheckout->canRenew = false;
-					}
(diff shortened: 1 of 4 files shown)


code/web/Drivers/Sierra.php
```diff
@@ -845,6 +845,10 @@ public function getCheckouts(User $patron, array $options = []): array {
 						$curCheckout->author = 'Unknown';
 					}
 					$curCheckout->formats = ['Unknown'];
+					$homeLibrary = $patron->getHomeLibrary();
+					if ($homeLibrary !== null && !$homeLibrary->allowToRenewILL) {
```
JonahCWilson: Nit:
As `curCheckout->canRenew` is defined as `true` above this, we can shorten this assignment to:
`$curCheckout->canRenew = !($homeLibrary && !homeLibrary->allowToRenewILL);`

mdnoble73 (approved):

PR #4314, opened 2026-05-28
DIS-2192 Curl Wrapper rate limit

lathomas64:
- [x] Created a Jira Issue for the change
- [x] Include a description in the Jira issue
- [x] Add Testing notes to Jira Issue
- [x] Include the testing notes in the pull request
- [x] pull request title should follow the template DIS-### Description
	- This ensures that JIRA recognizes the PR
- [x] pull request should include release notes
- [x] Jira also needs to be updated with developer information and the ticket should be in Code Review status.

---
- phpunit tests at tests/phpunit/tests/sys/throttlerTests.php
- normal page loading should be unaffected.
- I tested locally using a plugin that loaded a page which called curlGetPage to [example.com](http://example.com/) in a loop 5 times and looking at the timing. I copied the code for the main page of it in the code. 
- copy the commented CurlWrapper section added to the sites/default/conf/config.ini to your local sites config.ini and uncomment the CurlWrapper section and browse the site normally.
- run some cron jobs that do multiple requests to CurlWrappers like updateCommunityEngagementMilestones.php does when it pulls holds for users with milestones.


lathomas64:
@JonahCWilson @LiYanjun19 I think this one is good either of you see any issues here before in mark it ready?

sites/default/conf/config.ini
mdnoble73, line 1: The rateLimit does not seem to be used anywhere

mdnoble73 (changes requested):
@lathomas64 from what I can tell the throttler is not actually being used anywhere. Is that the case and this is just opening up future functionality?

sites/default/conf/config.ini
lathomas64: ah yes I changed that from rateLimit to requestInterval elsewhere.

lathomas64:
> @lathomas64 from what I can tell the throttler is not actually being used anywhere. Is that the case and this is just opening up future functionality?

the Throttler is being used in the CurlWrapper it just should preserve default behaviour unless you set a requestInterval in your config. I accidentally had the old name for the variable in the comment in the default config.ini but I updated it to reflect the changes in the code.

mdnoble73 (approved):

PR #4333, opened 2026-06-01
DIS-2286: Size-specific original URLs 

LiYanjun19:
bookcover_info.original_url is currently a single value per grouped work. When System Variables > Use Original Cover URLs is enabled, early redirect returns only the original_url stored in DB regardless of requested size and reload=1. 

We should store original URLs by size (small, medium, large) instead of one shared URL and allow regenerating the URL on reload=1 

To test:
1. Enable System Variables > Use Original Cover URLs.
2. Configured Syndetics or another third-party cover.
3. Find a grouped work that uses a third-party cover and open a cover URL like: `https://AAA.aspendiscovery.org/bookcover.php?id={groupedWorkId}&size=medium&type=grouped_work&category=Books`
5. Confirm the request redirects to the vendor URL, for example: `https://syndetics.com/index.aspx?type=xw12&pagename=MC.GIF&client={client}&isbn={isbn}`
7. Change `size=medium` to `size=large` in the Aspen bookcover.php URL and reload.
8. Confirm it still redirects to the same medium Syndetics URL (MC.GIF) instead of a large one.
9. Add `&reload=1` to the Aspen bookcover.php URL and reload.
10. Before the fix, confirm it still redirects to the same medium Syndetics URL (MC.GIF).
11. Apply the PR.
12. Open the same bookcover.php URL with `size=large` and reload.
13. Confirm it now redirects to the large vendor URL with `LC.GIF`.
14. Add `&reload=1` to the same large Aspen URL and reload again.
15. Confirm it still redirects to the large vendor URL with `LC.GIF`.

### DIFF
--- a/code/web/sys/Covers/BookCoverInfo.php
+++ b/code/web/sys/Covers/BookCoverInfo.php
@@ -18,6 +18,9 @@ class BookCoverInfo extends DataObject {
 	protected $uploadedImage;
 	protected $disallowThirdPartyCover;
 	protected $original_url;
+	protected $original_url_small;
+	protected $original_url_medium;
+	protected $original_url_large;
 	protected $last_url_validation;
 
 	public function getNumericColumnNames(): array {
@@ -92,8 +95,12 @@ class BookCoverInfo extends DataObject {
 	 * Get the original URL of the cover image
 	 * @return string|null
 	 */
-	public function getOriginalUrl(): ?string
+	public function getOriginalUrl(string $size): ?string
 	{
+		$sizeProperty = $this->getOriginalUrlPropertyForSize($size);
+		if ($sizeProperty !== null && !empty($this->$sizeProperty)) {
+			return $this->$sizeProperty;
+		}
 		return $this->original_url;
 	}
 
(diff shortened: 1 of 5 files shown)


code/web/sys/Covers/BookCoverInfo.php
```diff
@@ -182,4 +202,13 @@ public function getFirstLoaded() : ?int {
...
-}
\ No newline at end of file
+
+	private function getOriginalUrlPropertyForSize(string $size): ?string {
+		return match ($size) {
+			'small' => 'original_url_small',
+			'medium' => 'original_url_medium',
+			'large' => 'original_url_large',
+			default => null,
+		};
```
JonahCWilson, line 212: Nice use of the match statement!

code/web/sys/Covers/BookCoverProcessor.php
```diff
@@ -46,10 +46,9 @@ public function loadCover(array $configArray, Timer $timer, Logger $logger) : bo
 			if ($this->getCachedCover()) {
 				return true;
 			}
-		}
-
-		if ($this->checkForEarlyRedirect()) {
-			return true;
+			if ($this->checkForEarlyRedirect()) {
+				return true;
+			}
```
JonahCWilson: Nit: Can probably combine the above if and have:

`if ($this->getCachedCover() || $this->checkForEarlyRedirect())` but it's fine either way.

mdnoble73 (approved):

PR #4353, opened 2026-06-02
DIS-2484: Refactor to use DRY principle

JonahCWilson:
Minor refactor of code that does not follow the widely accepted  [DRY Principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

Extending the current code will involve copying the same 20-lines of code (already repeated 3 times) and repeating them for every hold type that may exist in the future.

By refactoring this code now, we're making it extensible with a single line change by simply adding the new holds type to the list of statuses.


mdnoble73:
https://aspen-discovery.atlassian.net/browse/DIS-2389 is still in development with changes needed and the filters are being adjusted for each section so they will be different. We should wait for refactoring until after coding is complete, and refactoring should be done in collaboration with the original developer rather overriding their work especially while development is being done.

JonahCWilson:
@catsoup11789 Feel free to use this PR as a model for future development in this section.  Even if the requirements for each one become unique, it's still considered best practice to have a helper function for each rather than aggregating all the code into one place.

JonahCWilson:
> https://aspen-discovery.atlassian.net/browse/DIS-2389 is still in development with changes needed and the filters are being adjusted for each section so they will be different. We should wait for refactoring until after coding is complete, and refactoring should be done in collaboration with the original developer rather overriding their work especially while development is being done.

@mdnoble73 Could you please point me to the appropriate document, page number, and paragraph within Aspen Community documents where this rule is stated so that I can have a better understanding of the Aspen workflow?

mdnoble73:
I don't think we have it written down as a rule per se. However, we have discussed situations like this in the Developer's group (possibly before you started?). Currently, if someone has an active Jira that they are working on other developers shouldn't duplicate their work or make conflicting changes without talking to that developer first. 

So at this point it's courtesy rather than a hard and fast rule, but it is a courtesy the community has tried to follow.

kylemhall:
> I don't think we have it written down as a rule per se. However, we have discussed situations like this in the Developer's group (possibly before you started?). Currently, if someone has an active Jira that they are working on other developers shouldn't duplicate their work or make conflicting changes without talking to that developer first.

Can you point us to a reference in the meeting notes?
 
> So at this point it's courtesy rather than a hard and fast rule, but it is a courtesy the community has tried to follow.

Can you point us to some examples where this has occurred?

Thanks!

mdnoble73:
Yep for sure.
Developer's meeting notes are here: https://docs.google.com/document/d/1AZUPjUEsSA6B9sGilqcjt36b9L9Y5RmLeMm_dayXySo/edit?tab=t.0#heading=h.4i3jiu4nf0xq

We talked about not making pure refactoring pull requests March 10th 
We talked about overlapping dev on jira issues April 8, 2025

kylemhall:
> Yep for sure. Developer's meeting notes are here: https://docs.google.com/document/d/1AZUPjUEsSA6B9sGilqcjt36b9L9Y5RmLeMm_dayXySo/edit?tab=t.0#heading=h.4i3jiu4nf0xq
> 
> We talked about not making pure refactoring pull requests March 10th We talked about overlapping dev on jira issues April 8, 2025

Thanks for the links! I went through the notes for both dates.

On April 8, 2025, there's nothing about overlapping development. The only related item that day was @sekjal's  discussion on etiquette around reassignment of issues (DIS-315), which is about handing off an inactive issue's assignee, not two people developing in parallel. It doesn't apply here.

On the March 10 note, "Please don't refactor code unless changes or enhancements are being made" was discussed in the meeting, and it means don't open PRs that only refactor. That's exactly why this
  shouldn't have become a standalone PR in the first place. @JonahCWilson raised this exact point in a review on #4298 on May 27. The review was resolved without discussion or change, and #4298 was merged 4 days
  later on May 31. Had that review been addressed in #4298, there would be no separate "refactoring" PR at all. It would simply have been part of that change.

mdnoble73:
The technical committee will discuss and work on policies at their next meeting.

PR #4358, opened 2026-06-03
DIS-2489: Update Koha Login Op

JonahCWilson:
The op value changed on the Koha side, so updating on our side to match.

### DIFF
--- a/code/web/Drivers/Koha.php
+++ b/code/web/Drivers/Koha.php
@@ -3743,32 +3743,23 @@ class Koha extends AbstractIlsDriver {
 		];
 
 		if (empty($user->ils_password)) {
-			if ($user->isLoggedInViaSSO) {
-				// This is a limitation of using Koha pages to perform logins rather than API requests.
-				// In the future, with an API request, user verification could be performed without a password when a user has logged in to Aspen via SSO.
-				$result['message'] = "Update Failed: Your password does not exist in Aspen because you performed a SSO without ever having performed a local login to Aspen."; // (i.e., local login = ILS-based login)
-			}
-			else {
-				$result['message'] = "Update Failed: The patron's password does not exist in Aspen because the patron has not logged in to Aspen for the first time yet.";
-			}
+			// This is a limitation of using Koha pages to perform logins rather than API requests.
+			// In the future, with an API request, user verification could be performed without a password when a user has logged in to Aspen via SSO.
+			$failureReason = $user->isLoggedInViaSSO ? 
+				"Your password does not exist in Aspen because you performed a SSO without ever having performed a local login to Aspen." :
+				"The patron's password does not exist in Aspen because the patron has not logged in to Aspen for the first time yet.";
+			$result['message'] = "Update Failed: $failureReason";
 		}
 
 		$kohaVersion = $this->getKohaVersion();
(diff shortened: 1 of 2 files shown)


mdnoble73 (commented):
@JonahCWilson is this version specific for Koha? If so we should check the version of Koha that we are connected to prior to changing the value we are sending.

JonahCWilson:
@mdnoble73 It is, but this change is already contained in a version-checked if-block:

```
kohaVersion = $this->getKohaVersion();
$csrfToken = '';
if ($kohaVersion >= 24.05) {
	//First get the page to get the csrf token
	$getResults = $this->getKohaPage("$catalogUrl/cgi-bin/koha/opac-user.pl");
	if (preg_match('/<input type="hidden" name="csrf_token" value="(.*?)" \/>/', $getResults, $matches)) {
		$csrfToken = $matches[1];
	}

	$postParams = [
		'koha_login_context' => 'opac',
		'login_password' => $user->ils_password,
		'login_userid' => $user->ils_barcode,
		'csrf_token' => $csrfToken,
		'op' => 'cud-login'
		'login_op' => 'cud-login'
	];
}
```

mdnoble73:
Was the value changed in 24.05 or was it changed in a later version of Koha?

code/web/Drivers/Koha.php
```diff
@@ -3743,32 +3741,21 @@ private function loginToKohaOpac(User $user) {
...
-		$csrfToken = '';
 		if ($kohaVersion >= 24.05) {
-			//First get the page to get the csrf token
-			$getResults = $this->getKohaPage("$catalogUrl/cgi-bin/koha/opac-user.pl");
-			if (preg_match('/<input type="hidden" name="csrf_token" value="(.*?)" \/>/', $getResults, $matches)) {
-				$csrfToken = $matches[1];
-			}
+			$postParams['csrf_token'] = $this->getCSRFToken("$catalogUrl/cgi-bin/koha/opac-user.pl");
 
-			$postParams = [
```
mdnoble73, line 3765: By removing this block, it looks like the login_password and login_userid are no longer sent. Previous values were password and userid. Is that change intentional/correct? Does it work in all versions targeted?

JonahCWilson:
@mdnoble73 It was changed in a later version of Koha, 25.11.04.  Bug for it is [here](https://bugs.koha-community.org/bugzilla3/show_bug.cgi?id=39055). 

Updated to it use the correct value on either side of Koha. Also added a helper function for CSRF gathering and found a few other places in the code which can use it.

mdnoble73:
@JonahCWilson the bug you linked does not mention removing the login_password and login_userid parameters. Why were those removed?

JonahCWilson:
@mdnoble73 I'm addressing these right now, I'll update once everything is finished.

code/web/Drivers/Koha.php
JonahCWilson: Corrected

JonahCWilson:
@mdnoble73 Should be good to go now. I fixed the error with the parameter names, and I have tested against pre 25.11.04 and post.

Let me know if you have any further questions or suggestions.

mdnoble73 (approved):

PR #4388, opened 2026-06-08
DIS-2509: load current and past event instance

Chloe070196:


code/web/services/Events/AttendanceManagement.php
mdnoble73: @Chloe070196 should this function be renamed if it is no longer just loading the upcoming events

Chloe070196: @mdnoble73 agreed! Amended accordingly (both the method name and the names of the variables storing its return value). 

Also changed the text describing the purpose of the Event Attendance landing page:

- clarified that administrators will be selecting event  instances (as opposed to parent events)
- removing the word 'upcoming' (inaccurate)
- list 'view registration' and 'manage attendance' as the two main available features* 


*Some administrators will also be able to use this to manage registrations themselves, but that is subject to a specific permission: it is an additional feature for some admins only, rather than the main purpose.

mdnoble73 (approved):

PR #4390, opened 2026-06-08
DIS-2508: db transaction for native events registration

Chloe070196:


mdnoble73 (commented):
@Chloe070196 I'm ok with this in concept if we want to bring it into Aspen

Chloe070196:
Thanks @mdnoble73 ! I do want to get back to that one and submit it properly - reckon it'll be for `26.10.00` if we're happy with that?

PR #4396, opened 2026-06-09
DIS-2288: Better support for more currencies

olivia-openfifth:
This PR does a bunch of preliminary work, but the primary user-visible change is in allowing any currency code to be specified in a system variable (with suggestions for the existing options). To support this wider range, this PR also switches to using PHP's `NumberFormatter` everywhere a currency symbol is needed.

*Note*: Under some circumstances, this may be an visible change. For example, using the currency code `CAD` when the user's locale is `en-US` will result in the currency symbol "CA$". However, if the locale is set to `en-CA` as appropriate, the currency symbol will be "$" as before.

### DIFF
--- a/code/web/interface/themes/responsive/DataObjectUtil/property.tpl
+++ b/code/web/interface/themes/responsive/DataObjectUtil/property.tpl
@@ -225,21 +225,9 @@
 					<a class="btn btn-default btn-sm" href="{$property.editLink|replace:'propertyValue':$propValue}">Edit {$property.label}</a>
 				</div>
 			</div>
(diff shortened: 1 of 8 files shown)


mdnoble73:
@olivia-openfifth could this be targeted to 26.07 rather than 26.06 since it's currently classified as a minor bug?

olivia-openfifth:
> @olivia-openfifth could this be targeted to 26.07 rather than 26.06 since it's currently classified as a minor bug?

No problem, I've rebased and updated the PR's target branch.

mdnoble73 (approved):

PR #4433, opened 2026-06-17
DIS-2553 linked holds

lathomas64:
to test:
* log into LiDA with an account with linked accounts
* go to place a hold and choose a linked account in the hold prompt
* the hold should be placed on the linked account instead of the logged in account


lathomas64:
@mdnoble73

PR #4450, opened 2026-06-22
DIS-2572: add configurable complexity score header for Cloudflare rate limiting

twstagg:
## Summary

Adds an opt-in `x-complexity-score` response header that enables per-endpoint rate limiting in Cloudflare WAF Rate Limiting Rules. Heavy endpoints (search results, AJAX) get higher scores so they consume more of a client's rate limit budget.

## What this does

- Adds `setCloudflareComplexityHeader()` called early in `index.php` request lifecycle
- New `matchEndpointWeight()` function using `switch(true)` for clean, extensible matching
- New `[Security]` config section in `config.ini` (disabled by default)

## Endpoint weights

| Score | Endpoints |
|---|---|
| **25** | `Search/Results`, `Union/Search` |
| **20** | `Search/AJAX`, `API/SearchAPI` |
| **15** | Any `GroupedWork` sub-path (e.g. `/GroupedWork/{id}/Home`) |
| **10** | Any `Record` sub-path, `Author/Home`, `Hoopla`, `OverDrive`, `PalaceProject`, `Axis360`, `CloudLibrary` |
| **1** | Everything else (default) |

## Configuration

Add to `config.ini`:

```ini
[Security]
enableComplexityHeader  = true
complexityHeaderName    = x-complexity-score
```

Both options are configurable per-site. The header name can be changed if your Cloudflare setup expects a different field name.

## How to use in Cloudflare

1. Go to **WAF → Rate Limiting Rules**
2. Create a rule that evaluates the `x-complexity-score` request header
3. Set thresholds based on your expected legitimate traffic

## Extending

To add new endpoint weights, add a `case` block in `matchEndpointWeight()`:

```php
case $module === 'YourModule' && $action === 'HeavyAction':
    return 30;
```

Match on `$module` alone for any sub-path, or add `&& $action === 'X'` to narrow it down.

## Files changed

- `code/web/index.php` — header logic (+24 lines net)
- `sites/default/conf/config.ini` — `[Security]` section (+8 lines)

### DIFF
--- a/code/web/index.php
+++ b/code/web/index.php
@@ -1675,41 +1675,45 @@ function setCloudflareComplexityHeader(): void {
 	}
 
 	$headerName = $configArray['Security']['complexityHeaderName'] ?: 'x-complexity-score';
-	$uri = $_SERVER['REQUEST_URI'] ?? '/';
 
-	// Match the URI to a weight bucket. Add new cases here to extend.
-	$complexity = matchEndpointWeight($uri);
+	// Module and action are already parsed by loadModuleActionId()
+	$module = $_GET['module'] ?? '';
+	$action = $_GET['action'] ?? '';
+
+	$complexity = matchEndpointWeight($module, $action);
 
 	header("$headerName: $complexity");
 }
 
 /**
- * Map a request URI to a complexity weight score.
+ * Map module/action to a complexity weight score.
  *
  * Add new case blocks here to cover additional endpoints.
  *
- * @param string $uri The request URI (e.g. /Search/Results?page=2)
+ * @param string $module The request module (e.g. Search)
+ * @param string $action The request action (e.g. Results)
  * @return int The complexity weight (default 1)
  */
-function matchEndpointWeight(string $uri): int {
-	switch (true) {
+function matchEndpointWeight(string $module, string $action): int {
+	switch ($module . '/' . $action) {
 		// Score: 25 - Heavy search result pages
-		case preg_match('#^/Search/Results(?:\?|/|$)#', $uri):
-		case preg_match('#^/Union/Search(?:\?|/|$)#', $uri):
+		case 'Search/Results':
+		case 'Union/Search':
 			return 25;
 
 		// Score: 20 - AJAX/API search endpoints
-		case preg_match('#^/Search/AJAX(?:\?|/|$)#', $uri):
-		case preg_match('#^/API/SearchAPI(?:\?|/|$)#', $uri):
+		case 'Search/AJAX':
+		case 'API/SearchAPI':
 			return 20;
 
 		// Score: 15 - GroupedWork detail and AJAX endpoints
-		case preg_match('#^/GroupedWork(?:/|$)#', $uri):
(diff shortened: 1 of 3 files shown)


twstagg:
https://aspen-discovery.atlassian.net/browse/DIS-2572

code/web/index.php
```diff
@@ -1644,3 +1647,69 @@ function setRoute(string $module, string $action = 'Home', ?string $id = null):
...
+		case preg_match('#^/Union/Search(?:\?|/|$)#', $uri):
+			return 25;
+
+		// Score: 20 - AJAX/API search endpoints
+		case preg_match('#^/Search/AJAX(?:\?|/|$)#', $uri):
+		case preg_match('#^/API/SearchAPI(?:\?|/|$)#', $uri):
+			return 20;
+
+		// Score: 15 - GroupedWork detail and AJAX endpoints
+		case preg_match('#^/GroupedWork/[0-9]+/(?:Home|AJAX)(?:\?|/|$)#', $uri):
```
LiYanjun19: grouped work id contains numbers, letters and dashes 

`case preg_match('#^/GroupedWork(?:/|$)#', $uri):`

code/web/index.php
```diff
@@ -1644,3 +1647,69 @@ function setRoute(string $module, string $action = 'Home', ?string $id = null):
...
+			return 15;
+
+		// Score: 10 - Record detail and Author home pages
+		case preg_match('#^/Record/[0-9]+(?:\?|/|$)#', $uri):
```
LiYanjun19: record id can also contain numbers and letters

`case preg_match('#^/Record(?:/|$)#', $uri):` would match more cases

LiYanjun19 (changes requested):

code/web/index.php
twstagg: Thank you, this should be fixed!

code/web/index.php
twstagg: Thank you, this should be fixed!

code/web/index.php
```diff
@@ -1644,3 +1647,69 @@ function setRoute(string $module, string $action = 'Home', ?string $id = null):
...
+ *
+ * Add new case blocks here to cover additional endpoints.
+ *
+ * @param string $uri The request URI (e.g. /Search/Results?page=2)
+ * @return int The complexity weight (default 1)
+ */
+function matchEndpointWeight(string $uri): int {
+	switch (true) {
+		// Score: 25 - Heavy search result pages
+		case preg_match('#^/Search/Results(?:\?|/|$)#', $uri):
```
mdnoble73: This can use $module and $action rather than doing a preg_match which will be significantly faster. 

You will also want to account for things like Hoopla Records, OverDrive Records, etc.

mdnoble73 (changes requested):

code/web/index.php
twstagg: I think this should be better now!

mdnoble73 (approved):

PR #4482, opened 2026-06-24
DIS-1938: feat: koha patron self renewals

Chloe070196:
replaces https://github.com/Aspen-Discovery/aspen-discovery/pull/3797


Chloe070196:
Rebased against 26.07.00, updated to include caching (otherwise every page load within MyAccount would have sent an API request to Koha - the logs got very noisy very quickly  on early versions of this development), and also to handle displaying the renew card button (and therefore the expiry notice) more than 30 days before expiry if the Koha configuration requires it, as requested by the sponsoring library who have been testing and using this development for a while now.


Ready for review, I'm particularly keen to get feedback on the stepper and the caching both!

mdnoble73 (approved):

mdnoble73 (approved):

PR #4485, opened 2026-06-24
DIS-600: JS translation system

lucasmontoya13:
Hardcoded strings in JavaScript files were not reachable by Aspen's translation system. This adds a server-side
  phrase registry (JsTranslations.php), a Smarty plugin that resolves all phrases into translated JSON on each page
  load, and a __() global function in JavaScript that performs the lookup with named variable substitution. All
  previous AspenDiscovery.translate() calls were migrated to __() and the dead method was removed.

For making JS phrases translatable : 
  1. Wrap the string: Replace any hardcoded JS string with __('phrase') or __('phrase with {var}', { var: value }) in the JS source file
  2. Run the extractor: Execute `php code/web/sys/Translation/extract_js_translations.php` from the repo root to scan for new __() calls and regenerate JsTranslations.php
  3. Commit both: The JS source change and the updated JsTranslations.php go in the same commit
  4. Rebuild aspen.js: Run `php code/web/interface/themes/responsive/js/merge_javascript.php` and commit the result in a separate standalone commit

  Test plan

  Before patching:

  1. Open any Aspen page in a non-English language configured in the admin panel
  2. Open the browser console and inspect Globals.jsTranslations
  3. See that the property is undefined or an empty object -- JS strings such as "Loading" or "Request Failed" appear
  in English regardless of language

  After patching:

  4. Repeat steps 1-2
  5. See that Globals.jsTranslations is a populated object with phrase keys mapped to their translated strings (e.g.
  "Loading" -> "Cargando" in Spanish)
  6. Trigger a JS action that uses a translated string (e.g. open a loading modal) -- confirm the string appears in the
   active language
  7. Switch the library to a language with no translations configured -- confirm that strings fall back to the English
  phrase and a [i18n] Missing JS translation warning appears in the console
  8. As a translation admin, enable translation mode, browse any page, and confirm that the JS phrases appear as
  translatable terms in the translation interface


mdnoble73:
@lucasmontoya13 this will need conflicts resolved. Can you also explain a bit more what all is needed from developers to use this system? Will we need to run a process to update translations regularly?

lucasmontoya13:
Done! I've just added the needed four steps for making JS phrases translatable

mdnoble73 (approved):

PR #4487, opened 2026-06-25
DIS-2463:  Fix 'More Options' error when selecting advanced search facet 

lucasmontoya13:
When a facet is configured to appear only in the Advanced Search (showInAdvancedSearch = 1, showInResults = 0), clicking "More options..." in the dropdown fails with "That facet could not be found."

  The root cause is that getFacetConfig() uses the global $action variable to decide which facets to return. During AJAX calls $action is 'AJAX', so advanced-search-only facets are excluded from both the Solr query and the config lookup.

  The fix sets $action = 'Advanced' before the search object is initialized in both getAdvancedSearchFacetPopup() and searchAdvancedFacetTerms() in Search/AJAX.php, ensuring the correct facet set is used throughout.
  
 Before patching:
  1 - Run UPDATE grouped_work_facet SET showInResults = 0 WHERE facetName = 'language';
  2 - Open Advanced Search
  3 - Open the Language dropdown and select "More options..."
  4 - See an error modal: "That facet could not be found"

  After patching:
  5 - Repeat steps 3 - 4
  6 - The popup opens with the full list of language values
  7 - Type a keyword (e.g. "spa") in the search field -- filtered results appear
  8 - Select a value -- the Language dropdown reflects it
  9 - Run UPDATE grouped_work_facet SET showInResults = 1 WHERE facetName = 'language';


JonahCWilson:
@lucasmontoya13 Is 26.04.00 the branch you want to target for this?

lucasmontoya13:
Hi Jonah! Since the error was reported for that version, I focused on patching it there without realizing that I should have targeted 26.07 first and then backported it to the older versions. I'm going to change the target branch. Thanks!

lucasmontoya13:
I'm not able to change the target from here. I'm going to close the PR and create a new one! Thanks again @JonahCWilson

PR #4496, opened 2026-06-28
DIS-2477: Remove legacy code for versions earlier than 24.11

LiYanjun19:
Koha support only goes back to 24.11 now - we should remove code referencing earlier versions

To Test:
1. Set up Koha 24.11+ environment 
2. Test listed functions:
    1. Patron login 
    2. Patron checkouts, holds, fines, and patron information loading 
    3. Patron info update, including preferred name, middle name, pronoun, and any configured extended attributes
    4. Place hold, including standard holds and volume-level holds
    5. Self-registration 
    6. Check out


LiYanjun19:
@lathomas64 @JonahCWilson The changes are mostly just deleting some code and adjusting the indentation, but I'd like to get more eyes on this PR since it touches many functions.

code/web/Drivers/Koha.php
```diff
@@ -5633,128 +5330,79 @@ function getNumMaterialsRequests(User $user) {
 
 	function getMaterialsRequests(User $user) {
 		//Just use the database to get the requests
-		if (false && $this->getKohaVersion() > 21.05) {
-			/** @noinspection PhpUnreachableStatementInspection */
-			$result = [
+		/** @noinspection PhpUnreachableStatementInspection */
```
lathomas64: The original had if(false on this branch. Do we want to change what code is being executed here?

code/web/Drivers/Koha.php
```diff
@@ -6937,11 +6564,7 @@ public function completeFinePayment(User $patron, UserPayment $payment) {
 			'message' => 'Unknown error completing fine payment',
 		];
 
-		$kohaVersion = $this->getKohaVersion();
 		$creditType = 'payment';
```
lathomas64: we were changing payment to all caps with newer versions of KOHA here, do we need to preserve that behavior or does Koha no longer care about the capitalization here?

code/web/Drivers/Koha.php
```diff
@@ -8713,198 +8263,188 @@ public function bypassReadingHistoryUpdate($patron, $isNightlyUpdate) : bool {
 	}
 
 	public function hasAPICheckout() : bool {
-		if($this->getKohaVersion() >= 23.11) {
-			return true;
-		}else {
-			return false;
-		}
+		return true;
```
lathomas64, line 8673: if this is always going to return true now do we still need this function?

lathomas64 (commented):
have a few questions on a few spots but looks good outside of that.

code/web/Drivers/Koha.php
lathomas64: this one is obviously fine just a potential spot for greater simplification. Not too much point in a function that always returns true.

LiYanjun19: Yes, we'll have this function always return true. Koha, Sierra, and Symphon return true for this function, while the other drivers return false.

code/web/Drivers/Koha.php
LiYanjun19: I restored the function and only removed the `&& $this->getKohaVersion() > 21.05`

code/web/Drivers/Koha.php
```diff
@@ -1099,13 +1041,9 @@ public function patronLogin($username, $password, $validatedViaSSO) {
...
-					if (isset($error) && strpos($apiURL, '/cgi-bin/koha/ilsdi.pl') !== false) {
-						global $logger;
-						$logger->log("ILS-DI is disabled", Logger::LOG_ERROR);
-					} else if (isset($error) && strpos($apiURL, "/api/v1/auth/password/validation") !== false) {
-						global $logger;
-						$logger->log("OAuth2 is disabled", Logger::LOG_ERROR);
-					}
+					if (isset($error) && strpos($apiURL, "/api/v1/auth/password/validation") !== false) {
+					global $logger;
+					$logger->log("OAuth2 is disabled", Logger::LOG_ERROR);
```
lathomas64: This if statement at 1044 is missing a closing bracket } after 1046

lathomas64: could not log in until I fixed that.

lathomas64 (changes requested):
looks good besides the one missing } I changed that locally and am able to do most of the things. Do we have an Aspen page we can check out from or is checking that just through the API?

mdnoble73 (approved):

PR #4509, opened 2026-06-29
DIS-1737: Link author names to author search results 

LiYanjun19:
Link author names to author search results instead of Author Home page. 

To Test:
1. Open search results page, a record, grouped work, or My Account checkouts holds titles etc that display an author link
2. Click the author name and confirm it opens the author home page 
3. Apply the PR
4. Refresh the same page and click the author name again
5. Confirm it now opens search results for that author


mdnoble73 (changes requested):
The 2 urls have different functionality in them including the option to find similar authors. If this is desired it should be an option rather than being the only option.

PR #4524, opened 2026-07-01
DIS-2624: Fix fatal TypeError when exporting all holds to CSV

kylemhall:
Clicking "Export All Holds to CSV" (and the per-section "Export All"
links) triggers MyAccount/AJAX.php?method=exportHolds. When no individual
holds are selected, exportHolds() calls filterHolds() with $selectedUser
(a string) as the second argument, but filterHolds() now requires that
parameter as array $filters. PHP throws an uncaught TypeError, producing
an HTTP 500 and a broken download for every patron.

Test Plan:
1) Log in as a patron with one or more holds
2) Go to My Account > Holds
3) Click "Export All Holds to CSV" without selecting any holds
4) Note the CSV downloads without error
5) On a linked account, select a specific user and export again
6) Note the CSV is scoped to that user's holds
7) Select individual holds and use "Export only selected"
8) Note that still works as before


mdnoble73:
@kylemhall Can you submit this against 26.06.01?

Also the selected user control no longer exist since that is part of the base filters now. This should apply the filters selected by the user. If you want me to look at a fix that uses the applied filters let me know.

kylemhall:
> @kylemhall Can you submit this against 26.06.01?
> 
> Also the selected user control no longer exist since that is part of the base filters now. This should apply the filters selected by the user. If you want me to look at a fix that uses the applied filters let me know.

So the "Export All Holds to CSV" will then only export what they see in the table? Is that correct and what you are asking for?

mdnoble73:
@kylemhall we can leave as is for now and then update the language to be Export All / Export Selected later depending on if filters are applied or not. Users will expect them to be filtered, but with the button titles as they are now it makes sense to not filter.

PR #4540, opened 2026-07-02
DIS-2627: Add Date fields to campaign email template

JonahCWilson:
IN DRAFT


JonahCWilson:
@AlexanderBlanchardAC I added some minor changes to work you have previously done.  Please take a look at your earliest convenience and let me know what you think.

The main use-case is that I have a partner who wants to be able to add dates from the campaign, I figured these would be nice-to-haves.

PR #4552, opened 2026-07-08
DIS-2622: Allow configuration of Search types and Search sorts option

K-Alette:
Add new Admin Section for Search Settings
Add new permissions for administering search settings Search Types can be disabled, enabled for basic & advanced, or enabled for advanced searches only Search Options can be enabled or disabled
If only one search type is being used, it will not show on the search bar Update release notes


code/web/sys/SearchObject/WebsitesSearcher.php
```diff
@@ -203,4 +203,9 @@ public function getEngineName() {
 	public function getDefaultIndex() {
 		return 'WebsiteKeyword';
 	}
+
+	public function getAdvancedSearchIndexes()
+	{
+		// TODO: Implement getAdvancedSearchIndexes() method.
```
mdnoble73, line 209: @K-Alette do we want to return an empty array for these so we know we get a good value back?

code/web/sys/SearchObject/AbstractGroupedWorkSearcher.php
```diff
@@ -1184,11 +1184,221 @@ public function __destruct() {
 	}
 
 	public function getSearchIndexes() : array {
+		global $library;
+		global $location;
```
mdnoble73, line 1188: This should use: 
global $locationSingleton;
$location = $locationSingleton->getActiveLocation();

We don't currently have a global $location

code/web/sys/SearchObject/AbstractGroupedWorkSearcher.php
```diff
@@ -1184,11 +1184,221 @@ public function __destruct() {
...
+					if ($searchTypes->enabled != 0) {
+						$searchType = $searchTypes->type;
+						$searchIndexes[$searchType] = translate([
+							'text' => $searchTypes->label,
+							'isPublicFacing' => true,
+							'inAttribute' => true,
+						]);
+					}
+				}
+				$searchIndexes2 = [
```
mdnoble73: We should make these configurable as well, but default them to only show for Advanced search.

code/web/sys/SearchObject/CloudSourceSearcher.php
```diff
@@ -623,4 +623,8 @@ public function getRecords($ids) {
 	}
 
 
+	public function getAdvancedSearchIndexes()
+	{
+		// TODO: Implement getAdvancedSearchIndexes() method.
```
mdnoble73, line 627: Would be good to have this return an empty array for consistency?

mdnoble73 (changes requested):
@Kodi Lein this is looking good so far. It does look like we need to include the search types that are currently only available in advanced search in the configuration. I.e. Publisher, Year of publication, Table of Contents, and Record Number. These would default to Enabled in Advanced Search Only, but could be completely disabled or promoted to the main interface.

mdnoble73 (approved):
Looks great

PR #4559, opened 2026-07-11
DIS-2640: UPCs should be stripped of trailing punctuation

K-Alette:
- Add process to strip any trailing punctuation for UPCs in MarcRecordProcessor.java 
- Update release notes


mdnoble73 (changes requested):
@K-Alette it looks like there are some cases where there can be data in the 024 with non-numeric data (https://www.loc.gov/marc/bibliographic/bd024.html). Let's update this to:
1) strip off the first word (similar to what we do in addIsbn)
2) remove ([\W]+)$ (whitespace from the end)
3) Maybe move to addUpc so it happens regardless of source?

mdnoble73 (approved):

PR #4561, opened 2026-07-13
DIS-2158: Allow sorting and filtering events by instanceCount

olivia-openfifth:
This PR adds infrastructure to DataObjects to allow them to specify SQL expressions for calculated columns. As these expressions are then used consistently for `SELECT` statements, they can also be used in sorts and filters. This new functionality is then used to support sorting and filtering events by their upcoming instance count.

### DIFF
--- a/code/web/release_notes/26.08.00.MD
+++ b/code/web/release_notes/26.08.00.MD
@@ -92,6 +92,11 @@
 // pedro
 
 // olivia
+### Aspen Events Updates
+- Allow sorting events by their upcoming count ([DIS-2158](https://aspen-discovery.atlassian.net/browse/DIS-2158)) (*OR*)
+
+### Other Updates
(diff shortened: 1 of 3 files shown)


mdnoble73 (changes requested):
@olivia-openfifth this should use the same mechanism as is used when filtering series by number of series members if possible. That has uses having clauses. 

If that won't work let me know.

olivia-openfifth:
Thanks for the review and for pointing me at the prior art. I've made a similar change for this issue [in a separate branch](https://github.com/openfifth/aspen-discovery/commit/492562eb177a1af01794ccd69f62c6828fb2c993) for reference.

I'm not a fan of repeating the same subquery in multiple locations, and prefer the approach in this PR of centralising a model's calculated columns in its `DataObject`. Is that approach something you'd be open to - perhaps with some adjustments? I can port the existing series count calculated column to the same approach in a follow up ticket.

mdnoble73 (approved):

mdnoble73 (approved):

PR #4564, opened 2026-07-13
DIS-2650: New Author Search Type

K-Alette:
Rename Author search to Authors and Contributors
Add new Author search that only searches Primary Author and not contributors

Update Release notes


code/web/sys/SearchObject/GroupedWorkSearcher2.php
```diff
@@ -423,7 +423,7 @@ public function processSearch($returnIndexErrors = false, $recommendations = fal
 		if (preg_match('/^"[^\"]+?\"$/', $this->query)) {
 			if ($handler == 'Keyword') {
 				$handler = 'KeywordProper';
-			} elseif ($handler == 'Author') {
+			} elseif ($handler == 'Author' || $handler == 'PrimaryAuthor') {
```
mdnoble73: @K-Alette I think we should create a PrimaryAuthorProper as well that will only check author_exact. That way if someone does a quoted search they won't get contributors with this search.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #4565, opened 2026-07-14
DIS-2677 update grantTokenAccess and setLibraryCache to parse the result wrapper

lathomas64:
* enable api requests for your ip
* ensure you have site listings set up as community in the Greenhouse Site Listings page
* navigate to /API/GreenhouseAPI?method=getLibraries&reload=true 
* you should get a full lists of community sites in your listings
* without this fix you should see errors in your log about $searchData not having success

### DIFF
--- a/code/web/Action.php
+++ b/code/web/Action.php
@@ -192,7 +192,7 @@ abstract class Action
 			$result = $curlWrapper->curlPostPage($systemVariables->greenhouseUrl . '/API/GreenhouseAPI?method=authenticateTokens', $postData);
 			if (!empty($result)) {
 				$data = json_decode($result, true);
-				$isValid = $data['success'];
+				$isValid = $data['result']['success'];
 
 				if($isValid) {
 					return true;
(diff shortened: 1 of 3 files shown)


lathomas64:
@LiYanjun19 @JonahCWilson @sekjal Any issues any of y'all see here?

JonahCWilson:
What's written looks fine so I'll mark "Approved", but for the two changes in which we're accounting for the `result`, I had a question.

I assume the error was that we previously had some object:
```
{
k: v,
k1:v2,
...
}
```

which then became:
```
{
result: {
     k:v,
     k1:v1,
     ...
    }
}
```

Is it likely that these sorts of changes will be implemented in the future?  Would it be worthwhile to add some sort of plumbing to check for the "result" field and assign appropriately?

So like `$data = get_some_response()` and then we check `if array_key_exists($data, 'result') then $data = $data['result']`


Just thinking out loud (and in text)

lathomas64:
good call, that change will make this backwards compatible

mdnoble73 (approved):

PR #4567, opened 2026-07-14
DIS-2398: Allow Merging of Series

K-Alette:
Undo $this->__set('dateUpdated', time()); change in update function for series.php as it was causing groupedWithSeriesId and isIndexed to not be changed properly when update was called in processGroupSeriesForm(). Instead, update dateUpdated within this function.

Also add seriesToGroupWithId as a visible field for testing when looking at the object edit pages for series


code/web/sys/DBMaintenance/version_updates/26.07.00.php
```diff
@@ -67,7 +67,22 @@ function getUpdates26_07_00(): array {
...
+		], // series_to_group_with
+		'remove_duplicate_series_members' => [
+			'title' => 'Remove Duplicate Series Members',
+			'description' => 'Remove duplicate series members.',
+			'continueOnError' => false,
+			'sql' => [
+				'removeDuplicateSeriesMembers',
+			]
+		], //remove_duplicate_series_members
+		'unique_series_members' => [
```
mdnoble73, line 79: @K-Alette Is this going to prevent the v1 series from working? I wonder if we need to set this depending on the version being used. 

We also have cases where a grouped work that is a collection can be multiple volumes. So we should add the volume number. 

I.e. some editions will collection books 3,4, and 5 so libraries will put them all in individually.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #4570, opened 2026-07-15
DIS-2641: Create embeddable events calendar.

attheshow:
DIS-2641: Create embeddable events calendar.


mdnoble73 (changes requested):
@attheshow this is looking close. The actual embed works for me, but when I click on any button in the preview or put it into another page, we lose the embed information so the full Aspen header shows. 

I think we'll need to update the links to preserve the embed parameter so it gets passed along when the view changes.

attheshow:
@mdnoble73 I see. Thanks for spotting that! I've made another update to the PR that should carry the embed parameter forward better throughout the links.

PR #4575, opened 2026-07-15
DIS-2521: Narrator/Reader, Contributor, etc Metadata Ingestion and Use for Hoopla Titles

K-Alette:
- Fix issue where contributors were not displaying on Hoopla record views
- Add facet for Narrator in grouped_works_v2 schema and in php 
- Update Narrator indexing for marc records, Hoopla, and Libby processors 
- Update release notes


mdnoble73 (changes requested):
Let's expand the Narrator facet to be Contributors. That will be useful in the future for illustrators, directors, actors, etc without having to have individual facets for each. 

We can possibly use the author2-role field for that facet?

mdnoble73 (approved):

PR #4578, opened 2026-07-16
DIS-2689: Preserve capitalization for grouped work series names

LiYanjun19:
Series from MARC records can display with lowercase in grouped work records even though the source records contain 490 0_ $a with the correct capitalization, and there is no NoveList/econtent source involved.

To Test:
1. Find a MARC record with Series names in all lowercase even 490_$a has correct capitalization
2. Apply the PR
3. Force reindex the record
4. Verify that the Series names are displayed with expected capitalization.


mdnoble73 (approved):

mdnoble73:
@LiYanjun19 can you resolve conflicts on this when you get a chance?

PR #4593, opened 2026-07-16
DIS-2680: aspen.sql regeneration ci

Jacobomara901:
This is stacked on top of the regeneration script rewrite (DIS-2679). 

Bit of a pain to test this one as it requires pushes to branches for testing. You can test with a dummy 99.99.99 or equivalent branch if needed.

It should only trigger on the aspen-discovery/aspen-discovery repo, and should only trigger if changes to the schema files are detected. It then will boot an adb, load a blank aspen.sql database, run pending db updates then run the regeneration script. 
It then checks for a diff (meaning there are aspen.sql updates) and then commits the regenerated file to the repo under the github ci bot's name.

We could consider also doing this for the JS, CSS and jar compilation too


Jacobomara901:
Also worth noting: 
It looks like there's massive changes to Aspen.sql for the rewrite of the generation script. This is only because inserts that were previously compounded to one line are now spread over multiple. 
The end result of the DB is identical from my testing

PR #4617, opened 2026-07-21
DIS-2647: Opening a Web Resource takes an extra click if not logged in

K-Alette:
- Update getWebResource() in web-builder.js to better handle opening resources if the user is not initially logged in If the web resource does not open in a new tab, the user will be redirected to the web resource page after a successful login 

- Add two new functions:
- getWebResourceAfterLogin()
- promptContinueToResource() - for when the resource is to be opened in a new tab, the user will be prompted to continue after a successful login and the resource will open in a new tab. Due to constraints with browser popup blocking, an extra click is unfortunately needed from the user

- Update release notes


mdnoble73:
@K-Alette can you resolve conflicts on this?

mdnoble73 (changes requested):
@K-Alette noted some issues found during testing in the Jira

mdnoble73 (approved):

PR #4620, opened 2026-07-21
26.08.00 DIS-2459 simplified search box

librarianbryan:
This needs a lot of review. I vibe coded with Claude and could not make these changes on my own. @tnlibrarian

### DIFF
--- a/code/web/release_notes/26.08.00.MD
+++ b/code/web/release_notes/26.08.00.MD
@@ -8,6 +8,11 @@
 // stephen
 
 // bryan
+## Simplified Search Box
+- Makes Keyword the singular, default search, ([DIS-2459](https://aspen-discovery.atlassian.net/browse/DIS-2459)) (*BNJ*) 
+- Puts the search button to the right of the search box
+- Moves ‘Advanced Search’ to the Search Dropdown. 
+- Primary Configuration > [a library] > Searching > Search Box > Use Simplified Search Box
 
 // chloe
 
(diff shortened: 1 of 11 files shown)


code/web/sys/DBMaintenance/version_updates/26.04.00.php
mdnoble73, line 1: This should be moved to 26.08.00.php

install/aspen.sql
mdnoble73, line 1: This is automatically generated and does not need to be added.

code/web/release_notes/26.08.00.MD
```diff
@@ -8,6 +8,11 @@
 // stephen
 
 // bryan
+## Simplified Search Box
```
mdnoble73, line 11: This should be updated to be more of a description of what is done in a single sentence. The new settings needs to be formatted properly for release notes. If you need an example let me know.

mdnoble73 (commented):
A few structural things prior to reviewing the full functionality.

mdnoble73:
@librarianbryan can you look at the functionality in https://aspen-discovery.atlassian.net/browse/DIS-2622? That overlaps with this a bit, but not 100%. It would allow you to completely remove the search type dropdown, but we probably need to look at what to do with the advanced search and the order of input fields does not change in that.

librarianbryan:
About DIS-2622, for us it is matter of design philosophy. We have two concerns which should have perhaps been separate commits:

1) UI--we have always is disliked the (functionless) magnifying glass on the left and the (largely functionless) search button on the right. We want to clean that up.

2) UX--what's the difference between "Authors" and "Authors + Contributors"? If your search needs an explainer you are doing it wrong. We want to eliminate the noise.

PR #4623, opened 2026-07-22
DIS-1689: fix: pre upgrade usagedata handling

Chloe070196:
In response to the community feedback received on DIS-1689, aims to fix issues with how 'old' (pre-DIS-1689) usage day is handled by custom and daily mode.

Leaving against 26.07.00 in case it proves a good 26.07.01 candidate.


mdnoble73:
@Chloe070196 I don't think this really needs a patch to go into 26.07.01. Could you raise a new Jira and submit against 26.08?

PR #4625, opened 2026-07-23
DIS-2375: fix: display informative PIN reset error messages

Chloe070196:
See [DIS-2375](https://aspen-discovery.atlassian.net/browse/DIS-2375)

[DIS-2375]: https://aspen-discovery.atlassian.net/browse/DIS-2375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/code/web/CatalogConnection.php
+++ b/code/web/CatalogConnection.php
@@ -1636,6 +1636,10 @@ class CatalogConnection {
 					}
 				} else {
 					$result = $this->driver->processEmailResetPinForm();
+					if (empty($result['success']) && empty($result['foundPatron'])) {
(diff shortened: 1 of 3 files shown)


Chloe070196:
Currently seems dependent on account profile weighting - if ILS gets evaluated before aspen admin,  the username-prefill guard may not be sufficient to clarify whether the ILS should have authority on whether the account's found / not found status is the purview of the ILS or of Aspen (I'm not certain, but this may be Koha version dependent). This could block admin PIN reset if the ILS pass runs first and hits a not found response. 

I'll re-work this.

mdnoble73 (approved):

PR #4646, opened 2026-07-29
DIS-2742: Array Offset Bug in "Titles on Hold"

JonahCWilson:
Tracked down a bug to Hold object format value being set to an empty list.  Updated it to use 'Unknown' and things seem to be populating again.


mdnoble73:
@JonahCWilson Jira needs to be updated properly before this can be reviewed. Thanks

JonahCWilson:
@mdnoble73 Sorry about that, Jira is up to date with testing instructions.

mdnoble73 (changes requested):
This needs the Jira updated appropriately. This also doesn't seem to fix the problem listed in the Jira which implies that the field is being passed in as an array?

mdnoble73 (approved):

PR #4661, opened 2026-08-02
DIS-2679: Refactor default database export into a PDO-based class 

Jacobomara901:
Should be merged before DIS-2680
To test:
Run DefaultDatabaseExporterTests
Start up fresh DB from aspen.sql generated from old script
Start up fresh DB from aspen.sql generated from new script
Verify end shapes are identical.


JonahCWilson:
Haven't reviewed yet, but this was on my mind this weekend, so thank you for writing this code.

mdnoble73 (commented):
@Jacobomara901 since this is not inserting known values, it does depend on the seed values to already be in the database. Is that correct?

Jacobomara901:
That's correct. The idea is that this will only be run on a fresh db containing the previous aspen.sql and any new pending db updates. Then there's only one source of truth and less chance for drift to occur in the regeneration if the schema changes. 
Ideally having this pair with DIS-2680 so that the CI handles this automatically on push.
Then we would jusr make the PR guidance to be don't modify the aspen.sql when you create new tables (which isn't done 90% of the time anyway)

mdnoble73:
This has the potential to include additional data that is not seed data if used outside a CI environment. To date I've done this by running the database generation while unit testing. If that is done, additional information like testing account profiles are included in the sql script which is incorrect.

Jacobomara901:
I agree, that would be possible if you were to run this on an "unclean" db and then commit the result. 
This is actually already a risk currently though. There are multiple tables that are ungated currently that can already do this in the original version. I believe its currently: bad_words, db_update, modules, permissions, role_permissions and roles.
Happy to add to the pr checks for changes to aspen.sql, or even better than that, we can work on some pre-commit/pre-push hooks? That's something I've been wanting to do for a while.
My question would also be, why would you want to be running this on an "unclean" db anyways? Especially if we're regenerating the schema on every push? Also, I'm contractually obligated to plug adb as the solution to running it locally, as this will always give you a fresh db (when run with the --ils=none option, else you'll get account profiles!)
Shameless, I know!

mdnoble73:
I'll talk to the Technical Committee since this will change some workflows.

mdnoble73:
With this model, what would be the best way to add new seed data to an existing/new table? A database update that inserts the data while updating the DefaultDatabaseExporter to mark the table as a seed table?

Jacobomara901:
More than reasonable, let me know if you want me to draft up some git hooks. We could possibly commit them and have a recommended one time setup to point your githook directory at them? Would stop people being able to commit up the aspen.sql (without a —force commit) and could resolve one of your concerns? Happy to support in any way.

Yeah, I think that having any additional seed rows be in a db update makes the most sense. You could also hand roll in to aspen.sql if you were so inclined but I am personally against that option! Either way, the seed table would need to be marked, yes. The above also allows for where clauses (see users) where appropriate.

As a complete aside: I think at TC it may also be beneficial to bring up the possibility of changing the db updates from being one file to being a directory and instead have the db updates be in individual dis based update files.
Would mean less conflicts for you, even if they do only take a few seconds to resolve. I’m sure you have a lot on your plate with the number of PRs you have to review now!

so 
26.08.00 (dir)
   | - dis_1.php
   | - dis_2.php

Let me know your thoughts here or if it’s already come up previously. 
Thanks again.

mdnoble73:
@Jacobomara901 I think it might be good to have a way to either seed with known data or with what already exists within the table with a preference towards loading with known data (maybe pulling from a CSV). 

If you are able to make changes would be great to get this into 26.09. Then we can sort out the github actions for auto regenerating.

Jacobomara901:
@mdnoble73 Sure thing, I'll add two flags to the script to allow export of seed data and the input of seed data bypassing the filters.
--seed-data and --export-seed-data
I'll make the filters a little more strict as well while I'm there. 

Then if we think it makes sense, I can have the CI also regenerate a seed.sql file for manual/local use? 
And we'll add new seed data with a db update as discussed previously.

The only use case I can think of then for running the regeneration locally would be for running tests on columns that weren't in the aspen.sql. I have another branch that I'll upstream for allowing pending db updates to run for test runs if that's okay? (I know the deadline was yesterday evening but I've just been off for a week - then bank holiday! -  so didn't get to it sooner.)

Let me get that sorted in a sec. Anything you think I've missed or misunderstood here?

PR #4663, opened 2026-08-03
DIS-2760: fix: retain unsaved changes

Chloe070196:


Chloe070196:
Any feedback much appreciated - particularly on [01c6da4](https://github.com/Aspen-Discovery/aspen-discovery/pull/4663/commits/01c6da4cb256bf7ff7b9ba7310c0ec9b91251fa3) as this certainly was the trickiest part here. I do wonder whether in practice we could consider trusting JS's RegExp to be close enough to the PHP regex comparison, but there are discrepancies between the two, so I did want to propose something that accounted for that. 

If, in practice, those discrepancies are negligeable, I'd be happy to drop [01c6da4](https://github.com/Aspen-Discovery/aspen-discovery/pull/4663/commits/01c6da4cb256bf7ff7b9ba7310c0ec9b91251fa3) as it does have the benefit of simplifying the code quite a bit.

mdnoble73 (approved):
looks great

PR #4668, opened 2026-08-03
DIS-2768 dockerfile update

JonahCWilson:
Update dockerfile to reintroduce xdebug + fix bug associated with sites folder being part of .gitignore.


mdnoble73 (commented):
This docker instance is intended for production usage. Installing xdebug on production servers is generally not recommended. Were you intending to install it for ADB?

PR #4669, opened 2026-08-03
DIS-2763: Return empty array for Website Search facet config when no facet group exits

poojary9991:


mdnoble73 (changes requested):
@poojary9991 this needs the Jira updated to reflect changes. 

Otherwise this looks great, happy to merge once Jira is all set.

PR #4677, opened 2026-08-04
DIS-2743: Make panel heading fully clickable 

samyoung-maker:
This will only drop down if you click on the text. The + implies it is something you can click on. Also the entire area has the cursor change to a pointer finger implying that it is clickable. Is it clickable and there is just something wrong with ours?


mdnoble73 (changes requested):
@samyoung-maker this needs release notes and Jira to be updated properly. Also needs to be submitted against 26.08. 

Thanks, looking forward to getting this incorporated!

PR #4680, opened 2026-08-04
DIS-2728: Add option to assign 2FA by role

catsoup11789:


LiYanjun19 (changes requested):
@catsoup11789 Hi, I found an issue here. `getTwoFactorAuthenticationSetting()` reuses `$this->_twoFactorAuthenticationSetting` for both the role lookup and the patron type fallback. Since changing `assignToUsersBy` does not clear stale assignments, a role can still point at a 2FA setting even after the setting is no longer role-based. 

To replicate: 
1. User has role A and patron type B.
2. 2FA setting #1 is assigned to role A. 2FA setting #2 is assigned to patron type B.
3. User logs in and gets 2FA setting #1, which is expected.
4. Change setting #1 from role to patron type, and assign it to patron type C.
5. User logs in again. The role lookup still finds setting #1 through the stale role assignment, but it does not return here because setting #1's `assignToUsersBy` is no longer 'role'`.
6. Then it falls back to patron type and sets the id to setting #2, but the same `TwoFactorAuthSetting` object still has fields populated from setting #1. so the fallback lookup can fail.

We can use separate TwoFactorAuthSetting objects and clear stale assignments when `assignToUserBy` changes.

code/web/sys/Account/User.php
```diff
@@ -5863,12 +5863,27 @@ public function getTwoFactorAuthenticationSetting() : ?TwoFactorAuthSetting {
...
+				if (!empty($permissionRoles)) {
+					foreach ($permissionRoles as $role) {
+						$this->_twoFactorAuthenticationSetting->id = $role->twoFactorAuthSettingId;
+						if ($this->_twoFactorAuthenticationSetting->find(true)) {
+							if ($this->_twoFactorAuthenticationSetting->assignToUsersBy == "role") {
+								return $this->_twoFactorAuthenticationSetting;
+							}
+						}
+					}
+				}
```
JonahCWilson, line 5907: This can be simplified a bit. Because the middle if-block doesn't do anything except check the inner-most if-block, we can just combine these:

```
$permissionRoles = $this->getRoles();
if (!empty($permissionRoles)) {
	foreach ($permissionRoles as $role) {
		$setting = $this->_twoFactorAuthenticationSetting;
		$setting->id = $role->twoFactorAuthSettingId;
		
		if ($setting->find(true) && $setting->assignToUsersBy == "role") {
			return $setting;
		}
	}
}
```

code/web/sys/TwoFactorAuthSetting.php
```diff
@@ -164,6 +192,17 @@ public function __get($name) {
...
+			if (!isset($this->_roles) && $this->id) {
+				$this->_roles = [];
+				$obj = new Role();
+				$obj->twoFactorAuthSettingId = $this->id;
+				$obj->find();
+				while ($obj->fetch()) {
+					$this->_roles[$obj->roleId] = $obj->roleId;
+				}
+			}
+			return $this->_roles;
```
JonahCWilson, line 205: This is a perfect opportunity for some light refactoring.  This function isn't that long, but who knows where it could end up in 6 months or a couple years' time?

If it needs modified again in the future, we're adding a minimum of 12 more lines of nearly identical code.

If we switched it something like below, we only have to add one line to $mappings and we're set.

```
public function __get($name) {
    $mappings = [
        'libraries' => ['class' => 'Library', 'property' => 'libraryId', 'var' => '_libraries'],
        'ptypes'    => ['class' => 'PType', 'property' => 'id', 'var' => '_ptypes'],
        'roles'     => ['class' => 'Role', 'property' => 'roleId', 'var' => '_roles'],
    ];
    
    if (isset($mappings[$name])) {
        return $this->_loadRelatedObjects($name, $mappings[$name]);
    }
    
    return parent::__get($name);
}

private function _loadRelatedObjects($name, $config) {
    $varName = $config['var'];
    
    if (!isset($this->$varName) && $this->id) {
        $this->$varName = [];
        $obj = new $config['class']();
        $obj->twoFactorAuthSettingId = $this->id;
        $obj->find();
        
        $property = $config['property'];
        while ($obj->fetch()) {
            $this->$varName[$obj->$property] = $obj->$property;
        }
    }
    
    return $this->$varName;
}

```

code/web/sys/TwoFactorAuthSetting.php
JonahCWilson, line 221: In the same vein as above, this is another great opportunity for future-proofing. We could do something like:

```
$mappings = [
    "libraries" => "_libraries",
    "ptypes" => "_ptypes",
    "roles" => "_roles"
];

if(array_key_exists($name, $mappings)){
    $varName = $mappings[$name];
    $this->$varName = $value;
} else {
    parent::__set($name, $value);
}
```

code/web/sys/TwoFactorAuthSetting.php
```diff
@@ -247,6 +290,31 @@ public function savePatronTypes() : void {
 		}
 	}
 
+	public function saveRoles(): void {
```
JonahCWilson, line 295: There is an opportunity to reduce some replication here. Something like below would eliminate duplicated code and reduce nesting:

```
public function saveRoles(): void {
	if (!isset($this->_roles) || !is_array($this->_roles)) {
		return;
	}

	$roleList = Role::getRoleList();
	foreach ($roleList as $id => $roleName) {
		$role = new Role();
		$role->roleId = $id;
		$role->find(true);

		$shouldApply = in_array($role->roleId, $this->_roles);
		$isCurrentlyApplied = $role->twoFactorAuthSettingId == $this->id;

		if ($shouldApply !== $isCurrentlyApplied) {
			$role->twoFactorAuthSettingId = $shouldApply ? $this->id : -1;
			$role->update();
		}
	}

	unset($this->_roles);
}
```

JonahCWilson (changes requested):

code/web/sys/TwoFactorAuthSetting.php
catsoup11789: This is a pre-existing code structure used throughout the code base and would be inappropriate to change it in this PR

code/web/sys/TwoFactorAuthSetting.php
catsoup11789: Again, this is preexisting structure in the codebase and would be an inappropriate time to change it.

code/web/sys/TwoFactorAuthSetting.php
catsoup11789: Again, this is pre-existing code structure and would not be a good time to try and change it (look at every other saveX function in any other class)

JonahCWilson: This question pertains to every comment related to refactoring:

If refactoring for the sake of refactoring is a no-go, when is the best time to introduce refactoring changes, if not when introducing new code to the codebase?

code/web/sys/Account/User.php
```diff
@@ -5889,15 +5889,37 @@ public function getTwoFactorAuthenticationSetting() : ?TwoFactorAuthSetting {
...
+						if ($roleSetting->find(true) && $roleSetting->assignToUsersBy == "role") {
+							$this->_twoFactorAuthenticationSetting = $roleSetting;
+							return $this->_twoFactorAuthenticationSetting;
+						}
+					}
 				}
-				if (!$this->_twoFactorAuthenticationSetting->find(true)) {
+
+				// As a backup, we will check if the user is required to use 2FA based on their patron type or account profile.
+				$fallbackSetting = new TwoFactorAuthSetting();
```
LiYanjun19, line 5911: Missing `$patronType = $this->getPTypeObj();`here.

LiYanjun19 (changes requested):

code/web/sys/Account/User.php
catsoup11789: Doh. Removed too many lines after updating a comment when I was done.

LiYanjun19 (approved):

PR #4682, opened 2026-08-05
DIS-2764 Optimize Solr Schema to maximize retrieval speed and minimize size on disk

mdnoble73:
Updated to create a grouped_works_v3 schema and solrconfig. 

Updated fields within solr to set indexed, docValues, omitNorms, and omitTermFrequency appropriately

Will need to set the Solr Indexer Version to 3 within System Variables. After the index finishes, Aspen will automatically move the Search version to 3.

Test by ensuring that all data within grouped work display shows properly. 

record_details, item_details, and scoping_details that were previously deprecated are removed (the data is now stored within the database.

Ensure that all facets work properly

Ensure suggestions work properly

Size on disk for solr should be roughly half or less of the size of grouped_work_v2 (not a reason to fail if it’s not, but should be smaller). 

Performance of searches should be the same or better to previous versions.


code/web/sys/SystemVariables.php
```diff
@@ -200,9 +201,10 @@ static function getObjectStructure(string $context = ''): array {
 						'type' => 'enum',
 						'values' => [
 							2 => 'Version 2 (Edition information)',
+							3 => 'Version 3 (Optimization of solr fields)'
 						],
 						'label' => 'Grouped Work Search Version',
-						'description' => 'The Solr Core Version to search with.  In 22.06 and above this should be version 2 in most cases.',
+						'description' => 'The Solr Core Version to search with.  In 22.08 and above this should be version 3 in most cases.',
```
LiYanjun19: Typo, "26.08"

code/web/sys/SearchObject/GroupedWorkSearcher3.php
```diff
@@ -0,0 +1,1056 @@
...
+					if (!empty($facetInfo)) {
+						$availabilityToggleId = $facetInfo->id;
+					}
+				} elseif ($facetName == 'available_at' || $facetName == "available_at_$solrScope") {
+					$selectedAvailableAtValues[] = $value;
+				} elseif ($facetName == 'format_category') {
+					$selectedFormatCategoryValues[] = $value;
+				} elseif ($facetName == 'format') {
+					$selectedFormatValues[] = $value;
+				}
```
LiYanjun19, line 208: Scope, format, format category, and availability are in `edition_info`, meanwhile, `available_at, format_category, format` etc are `indexed="false"` in v3 schema.xml so these fields are not searchable. Should we use `edition_info` to build the query here?

code/reindexer/src/org/aspen_discovery/reindexer/GroupedWorkSolr3.java
```diff
@@ -0,0 +1,762 @@
...
+			doc.addField("user_rating_link", userRatingLink);
+			doc.addField("user_not_interested_link", userNotInterestedLink);
+			doc.addField("user_reading_history_link", userReadingHistoryLink);
+			doc.addField("list_link", listLink);
+			for (Long listId : listEntryWeights.keySet()) {
+				doc.addField("list_entry_weight_" + listId, listEntryWeights.get(listId));
+			}
+			for (Long listId : listEntryDatesAdded.keySet()) {
+				doc.addField("list_entry_date_added_" + listId, listEntryDatesAdded.get(listId));
+			}
```
LiYanjun19, line 365: These three fields are in v2 schema but not in v3 schema.

LiYanjun19 (changes requested):

code/web/sys/SearchObject/GroupedWorkSearcher3.php
mdnoble73: This will change as part of DIS-2535 but for this update it just matches what is in GroupedWorkSearcher2

code/reindexer/src/org/aspen_discovery/reindexer/GroupedWorkSolr3.java
mdnoble73: Fixed

code/web/sys/SystemVariables.php
mdnoble73: Fixed

LiYanjun19 (approved):

PR #4690, opened 2026-08-06
DIS-2780 Collection Spotlight Caching

JonahCWilson:
The current flow for API calls to `getCollectionSpotlight` is as follows:

1. getCollectionSpotlight >
2. collectionSpotlight.tpl > collectionSpotlightTabs.tpl >
3. Search/AJAX/getSpotlightTitles >
4. From here, things go into individual RecordDrivers which then provide a value leading to /bookcover.php for loading.

In doing so, fetching bookcovers, even if they're cached, comes at the expense of more network traffic through /bookcover.php.

The new flow is as follows:
1. getCollectionSpotlight
   - New Functionality: Using incoming data, prepopulate the previously rendered HTML and assign it to a variable which is passed down to collectionSpotlight.tpl. At this point, all of the HTML has been update to go to a new endpoint in SearchAPI, getCollectionSpotlightCover. This process is cached using memcache, so any repeated lookups will return the cached version, or follow the flow to populate the value, then cache it for future uses.
3. collectionSpotlight.tpl > collectionSpotlightTabs.tpl >
4. getCollectionSpotlightCover

In making these changed, a couple things are accomplished:
1. Cached requests are handled without extra network traffic.  In the current version, accessing a cached bookcover through bookcover.php still requires an extra network call to get it.
2. API Consolidation.  By keeping all requests under the SearchAPI umbrella, developers can more easily write rules pertaining to traffic handling and caching.

Jira:
Link: https://aspen-discovery.atlassian.net/jira/for-you?tab=assigned&selectedIssue=DIS-2780
Status: Code-Review
Target: 26.09.00
Testing-Instructions: Added and present in Jira

Testing:
- Tested against beta site and partner test sites.

Any questions, comments, or concerns, please let me know.  If you find that I'm missing something in Jira, please inform me of what needs updating rather than a generic "Update the Jira".  Thank you.


mmaurakis-bws:
Tested on a high load server that typically takes a while to respond with PNG based covers and the caching has made a significant difference allowing instant response!

JonahCWilson:
@mdnoble73 Please review at your earliest convenience.

@mmaurakis-bws @kylemhall For visibility.

code/web/services/Search/AJAX.php
```diff
@@ -247,83 +247,181 @@ function getListTitles() : array {
...
+		
+		// 3. Prepare Lightweight Payload for Caching
+		$cachePayload = $result;
+		if (!empty($cachePayload['titles'])) {
+			foreach ($cachePayload['titles'] as &$item) {
+				// Extract record ID from HTML string if not explicitly set
+				if (empty($item['id']) && !empty($item['formattedTitle'])) {
+					if (preg_match('#/GroupedWork/([^/]+)/Home#', $item['formattedTitle'], $matches)) {
+						$item['id'] = $matches[1];
+					} elseif (preg_match('#bookcover\.php\?[^"]*id=([^&"#]+)#', $item['formattedTitle'], $matches)) {
```
mdnoble73, line 313: This is not a valid regular expression. There is an unknown ] modifier.

code/web/services/Search/AJAX.php
```diff
@@ -247,83 +247,181 @@ function getListTitles() : array {
...
-	function getSpotlightTitles(): array {
-		global $interface;
-		$listName = strip_tags($_GET['scrollerName'] ?? 'List' . $_GET['id']);
-		$interface->assign('listName', $listName);
+	function getSpotlightTitlesForList(int $listId, string $scrollerName, string $coverSize = 'medium', bool $reload = false): array {
+		global $interface, $memCache;
+
+		$cacheKey = "spotlight_titles_" . $listId;
+
+		$interface->assign('listName', $scrollerName);
```
mdnoble73, line 258: Collection spotlight results can be different based on the servername that a user is using. I.e. librarya vs. librayrb. If we are adding caching these will need to be cached based on the active servername

code/web/services/Search/AJAX.php
```diff
@@ -247,83 +247,181 @@ function getListTitles() : array {
...
+		'</div>';
+	}
+
+	/**
+	 * Routes spotlight cover requests through SearchAPI so an embedded spotlight
+	 * does not require direct access to /bookcover.php.
+	 */
+	private function rewriteSpotlightCoverUrls(string $formattedTitle): string {
+		return preg_replace(
+			'#(?:(?:https?:)?//[^/]+)?/bookcover\\.php\\?#',
```
mdnoble73, line 421: What is the rationale for rewriting bookcover.php? This seems like an unneeded rewrite.

mdnoble73 (changes requested):

PR #4702, opened 2026-08-10
DIS-2753: Run fewer cronjobs as root

olivia-openfifth:
This PR updates several cronjobs to run as the `aspen` user, rather than `root`. I believe that one cronjob was added which required `root`, and then future cronjobs used this as a template.

I've marked this as draft to indicate that this is worthy of more intensive review. I've audited every cronjob I've changed the user of, but I may have missed portions of them that do require `root`. Similarly, I've tested the migration script and tried to make it robust, but there is still a chance it could break existing installations.

### DIFF
--- a/install/updateCron_26.09.00.php
+++ b/install/updateCron_26.09.00.php
@@ -1,0 +1,111 @@
+<?php
+
+$jobsToChange = [
+	'updateCommunityTranslations.php',
+	'generateMaterialRequestHoldCandidates.php',
+	'dismissYearInReviewMessages.php',
+	'cleanupSharedSessions.php',
+	'talpaWorksCron.php',
+	'talpaRecalculationCron.php',
+	'sendCampaignEmails.php',
+	'sendCampaignEndingEmails.php',
+	'loadInitialReadingHistory.php',
+	'fetchILSMessages.php',
+	'sendILSMessages.php',
+	'purgeSoftDeleted.php',
+	'updateCommunityEngagementMilestones.php',
+	'purgeUserAppRequestLogs.php'
+];
+
+# Match a possible comment delimiter and indent as well as the {5} timing
+# sections of a cron line in a group to retain them, followed by "root" which we
+# replace with "aspen".
+$pattern = '/^((?:#\\s*)?(?:\\S+\\s+){5})root(\\s)/';
+$replacement = '\\1aspen\\2';
+
+function validateCron($crontab) {
+	$proc = proc_open(
+		'crontab -n -',
+		[
+			0 => ['pipe', 'r'],
+			1 => ['pipe', 'w'],
+			2 => ['pipe', 'w'],
+		],
+		$pipes
+	);
+
+	if(!$proc) {
+		// Error creating process, assume the crontab is valid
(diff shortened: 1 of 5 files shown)


install/updateCron_26.09.00.php
```diff
@@ -0,0 +1,66 @@
...
+	'fetchILSMessages.php',
+	'sendILSMessages.php',
+	'purgeSoftDeleted.php',
+	'updateCommunityEngagementMilestones.php',
+	'purgeUserAppRequestLogs.php'
+];
+
+# Match the {5} timing sections of a cron line in a group to retain them
+# followed by "root" which we replace with "aspen".
+$pattern = '/^((?:[0-9\\-\\/*]+ +){5})root /';
```
gmcharlt: This misses various valid timing specifiers, e.g.,

* `59 23 * * Sun`
* `17,37 * * * *`
* `1 15 * * Feb 2`

It also misses lines that have been (temporarily?) commented out; we may still want to change those to the correct user.

[This](https://stackoverflow.com/questions/14203122/create-a-regular-expression-for-cron-statement/63729682#63729682) from StackOverflow may be one way to boil this particular ocean, but I wonder if something like `/^((?:\S+ +){5})root /` would suffice.

gmcharlt (commented):
I've looked this over.  All of the cronjobs you identified do appear to be better run as the Aspen user. I do have a comment about the migration script.

install/updateCron_26.09.00.php
gmcharlt: Two more thoughts on making it more robust:

* Consider, as part of changing a line, inserting the previous version of the line as a comment along with a note to the effect that it was changed by the DIS-2753 update.
* Consider running `crontab -n` as a syntax check off the updated version before overwriting the file.

olivia-openfifth: Thanks for the review! Good spot on the possible timing specifiers - with so many possible extensions, I've gone with the more lenient `\S` approach. I've also allowed for changing the user in commented lines.

Both of your suggestions for robustness sounded great to me, so I've implemented both of them. At the moment, it fails safe if `crontab -n` can't be run, but maybe we want to go ahead in that case?

gmcharlt: > Both of your suggestions for robustness sounded great to me, so I've implemented both of them. At the moment, it fails safe if `crontab -n` can't be run, but maybe we want to go ahead in that case?

I think that's probably OK. I would guess that if the `crontab` command isn't present, the server likely is not using `cron` in the first place, so the operator perforce would have to set something else up to schedule the jobs.

gmcharlt: Also, I tested the migration script and it works for me, so from my POV, the only pending matter left is settling on the behavior when `crontab` is not present.

olivia-openfifth:
Thanks to @gmcharlt for the initial review, which has given me enough confidence to mark this as a non-draft.

I've opted to accept the updated crontab if there are issues running `crontab -n`, but I'm happy to change back to failing safe if that's preferred.

mdnoble73 (approved):

PR #4704, opened 2026-08-11
DIS-2785: Bump Docker's Solr image to 9.10.1

lucasmontoya13:
Updates docker/files/solr/Dockerfile to build Solr's Docker image from solr:9.10.1 instead of solr:8.11.2, matching the version already used by the production install/upgrade scripts. Since Solr 9 moved ICU-based text analysis filters (used by every core's schema) out of core into an optional module, the Dockerfile now copies the analysis-extras module into Solr's shared library path so those filters keep loading correctly. Also adds Docker provisioning for the grouped_works_v3 core alongside the existing grouped_works_v2, so it's buildable locally in addition to production. docker/docker-compose.yml is updated accordingly to precreate the new core at container startup.


mdnoble73:
@lucasmontoya13 could you resubmit this against the 26.08 branch?

PR #4706, opened 2026-08-11
DIS-2798: Aspen Discovery CI - Static Analysis tools

JonahCWilson:
Add PHPStan to Composer's Dev requirements.

Once installed, can be run across branches to check for bugs as code is being developed.

install composer 
cd /code/web
set phpstan.neon level 
    0  Basic checks: unknown classes, unknown functions, wrong method calls, basic syntax/type discovery.
    1  Adds checks for possibly undefined variables and more basic code correctness.
    2  Checks unknown methods/properties more aggressively across expressions.
    3  Checks return types and assigned types more deeply.
    4  Checks dead code / unreachable branches and stricter type compatibility.
    5  Checks argument types passed to methods/functions.
    6  Requires and checks more precise type information, especially missing typehints.
    7  Reports partially wrong union types, like calling a method that exists on only some possible types.
    8  Reports nullable type issues, like calling a method on `Foo|null`.
    9  Very strict about `mixed`; allows only passing `mixed` to another `mixed`.
    10 Strictest level; more exhaustive handling of implicit/explicit `mixed` and edge cases.

./vendor/bin/phpstan analyse . --memory-limit=1G

* The memory limit may or may not be necessary depending on your system.


JonahCWilson:
For a future PR, when we install composer dependencies in production, we need to add the `--no-dev` flag.

LiYanjun19:
Forgot to hit approve. 
This PR has been reviewed, tested, and approved.

PR #4710, opened 2026-08-12
DIS-1871: fix crash when attempting payment of certain fines from Evergreen

gmcharlt:
Two changes:

* Make trimStringToLengthAtWordBoundary() return the empty string if given a null value to trim
* Update the Evergreen driver to use the empty string as the fine reason if the billing note happens to be null


gmcharlt:
Noting that the tabs-vs-space failure is an artifact of my filing the pull request with the wrong base branch initially.

PR #4711, opened 2026-08-12
DIS-1871: fix crash recording payments of certain Evergreen fines

gmcharlt:
Two changes

* Make trimStringToLengthAtWordBoundary() return the empty string if given a null value to trim
* Update the Evergreen driver to use the empty string as the fine reason if the billing note happens to be null

### DIFF
--- a/code/web/sys/Utils/StringUtils.php
+++ b/code/web/sys/Utils/StringUtils.php
@@ -12,7 +12,7 @@ class StringUtils {
 				$string .= '...';
 			}
 		}
-		return $string;
+		return $string ?? '';
 	}
 
 	static function getCurrencyFormatter() : NumberFormatter {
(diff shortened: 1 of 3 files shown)


gmcharlt:
Redid the pull request to get a clean pass of the checks.

mdnoble73:
This should be named
DIS-1871 (description)
to match pull request guidelines.

code/web/sys/Utils/StringUtils.php
```diff
@@ -12,7 +12,7 @@ public static function trimStringToLengthAtWordBoundary($string, $maxLength, $ad
 				$string .= '...';
 			}
 		}
-		return $string;
+		return $string ?? '';
```
mdnoble73, line 15: Is this change required? I could result in other errors if $string comes in null, but that would have also caused the return to fail so $string is likely never null coming in. Setting a parameter type of string for $string or checking for null earlier may be more appropriate.

gmcharlt:
> This should be named DIS-1871 (description) to match pull request guidelines.

That's what I get for rebasing the PR in a hurry. Name updated.

mdnoble73 (approved):

PR #4717, opened 2026-08-13
DIS-2830: Overdrive _callUrl error handling

JonahCWilson:
This code may be overkill:

`$validResponse = $response && $response->message != 'An unexpected error has occurred.' && $response->errorCode != null;`

I haven't encountered the 'An unexpected error has occurred' response, but I imagine it's probably part of a response object with a populated 'errorCode' value.  If that's the case, we can eliminate that check and just use the check for errorCode.


mdnoble73 (changes requested):
This changes the actual error handling and no longer accounts for the "An unexpected error has occurred" message. Although you didn't experience it during your testing, that doesn't mean that it doesn't occur. Based on the error it is an error that would happen only if the OverDrive servers were having issues so replicating would be difficult other than with Mocks. 

Adding the unit tests provided in the Jira would be great for preventing future regression.

code/web/Drivers/OverDriveDriver.php
```diff
@@ -367,28 +367,26 @@ private function _connectToPatronAPI(OverDriveSetting $settings, User $user, str
...
+		global $interface;
+		$this->apiCurlWrapper->addCustomHeaders([
+			"Authorization: $tokenData->token_type $tokenData->access_token",
+			"User-Agent: Aspen Discovery " . $interface->getVariable('aspenVersion'),
+		], true);
+
+		$content = $this->apiCurlWrapper->curlGetPage($url);
+		ExternalRequestLogEntry::logRequest('overdrive.callUrl_' . $methodName, 'GET', $url, $this->apiCurlWrapper->getHeaders(), false, $this->apiCurlWrapper->getResponseCode(), $content, []);
+		$response = json_decode($content);
+		$validResponse = $response && !($response->errorCode || $response->message === 'An unexpected error has occurred.');
```
mdnoble73: @JonahCWilson This is still not matching the previous test. The equivalent would be:
$validResponse = $response && !(isset($response->message) || $response->message === 'An unexpected error has occurred.');

The change might be valid if in every case a message is provided an errorCode is provided, but not sure the API documentation specifies that.

mdnoble73 (changes requested):

mdnoble73 (approved):

PR #4723, opened 2026-08-14
DIS-594: Open Archives - improve retrieval of thumbnail images via image regular expression

gmcharlt:


NashvillePublicLibrary (approved):
Looks good. That Omeka converts special characters in thumbnail URIs to html entities seems nuts.

PR #4735, opened 2026-08-17
DIS-2840: PHPStan Baseline Bugfixes

JonahCWilson:
This PR has a dependency on https://github.com/Aspen-Discovery/aspen-discovery/pull/4706

The changes in this PR are pretty minor, mainly adding return type values to functions and missing return statements. Nearly everything touched here is a legacy file.


code/web/Drivers/SIP2Driver.php
```diff
@@ -516,5 +516,7 @@ public function getHolds(User $patron): array {
 				break;
 			}
 		}
+
+		return [];
```
gmcharlt: This should be `return $holds;`; otherwise, this routine ends up making a patron information request and throwing away the results.

The lack of a valid return predates this change, of course, but I assume that SIP2Driver is only lightly used, if at all.

code/web/RecordDrivers/Axis360RecordDriver.php
```diff
@@ -11,7 +11,7 @@ class Axis360RecordDriver extends GroupedWorkSubDriver {
 	private $axis360RawMetadata;
 	private $valid;
 
-	public function __construct($recordId, $groupedWork = null) {
+	public function __construct($recordId=null, $groupedWork = null) {
```
gmcharlt: Nitpick: I think usual style is to have spaces around the equals sign.

More substantively, this change and its peer changes to GroupedWorkSubDriver subclasses bothers me for two reasons:

* For most of them, trying to construct them without a record ID doesn't make sense
* I thought overriden constructors were exempt from the normal signature compatibility rules.

gmcharlt (changes requested):

code/web/RecordDrivers/Axis360RecordDriver.php
JonahCWilson: I can update the spacing around the `=`, no problem.

To your second points, I agree for the most part.  Several of these constructors are abstract and will never be instantiated, so it doesn't really make sense either way to update their constructor.  What PHPStan is complaining about is violation of the [Liskov Substitution Principle](https://phpstan.org/error-identifiers/parameter.notOptional). You are right, in that overriding sub-constructors is okay and exempt.

In terms of the Aspen Codebase as-is, adding these default values isn't going to hurt anything. However, if this is frowned upon or a no-go, our options are as follows:
- Add `// @phpstan-ignore parameter.notOptional` decorators for all of these (Not ideal)
- Update the phpstan.neon file to ignore this error. (Preferred)

Given the variance in terms of coding styles which find their way into Aspen, I'm rather reticent to just globally ignore errors. Having default values in these constructors is a small price to pay to catch potential bugs if/when new code is introduced later on.

gmcharlt: I would lean towards updating `phpstan.neon` for these. Defaulting the `$recordId` parameter to null, while I could live with it, would mean losing the possibility of catching somebody (say) doing a `new OverDriveRecordDriver()` and omitting the ID.

PR #4748, opened 2026-08-20
DIS-2789: AspenUsage counters use read-modify-write, losing page view counts and creating duplicate daily rows

kylemhall:


code/web/sys/SystemLogging/AspenUsage.php
```diff
@@ -124,27 +130,130 @@ public function getInstance() {
...
+
+	public function incBlockedRequests() : bool {
+		return $this->incrementField('blockedRequests');
+	}
+
+	public function incBlockedApiRequests() : bool {
+		return $this->incrementField('blockedApiRequests');
+	}
+
+	private function incrementField(string $fieldName) : bool {
```
JonahCWilson, line 229: Minor suggested refactor for this function:

```
private function incrementField(string $fieldName) : bool {
	if (!property_exists($this, $fieldName)) {
		return false;
	}
	
	$this->$fieldName++;
	$getQuery = fn($id) => "UPDATE aspen_usage SET $fieldName = $fieldName + 1 WHERE id = $id";

	try {
		if (!empty($this->id)) {
			return $this->query($getQuery($this->id));
		}

		//insert catches PDO exceptions internally and returns false, a duplicate key error does not throw
		if ($this->insert() !== false) {
			return true;
		}
		//Another request created the row for today first, load it and increment atomically
		$today = new AspenUsage();
		$today->instance = $this->instance;
		$today->year = $this->year;
		$today->month = $this->month;
		$today->day = $this->day;
		if ($today->find(true)) {
			$this->id = $today->id;
			return $this->query($getQuery($this->id));
		}
		return false;
	} catch (Exception $ex) {
		//Ignore this, the table has not been created yet
		return true;
	}
}
```

mdnoble73 (approved):

mdnoble73 (changes requested):
@kylemhall functionality in the code looks good, but this needs Jira (https://aspen-discovery.atlassian.net/browse/DIS-2789) updated with testing notes, developer information, etc.

mdnoble73 (changes requested):
This still needs Jira updated properly. It will need to be resubmitted to 26.10 since the deadline for 26.09 is passed.

kylemhall:
> This still needs Jira updated properly. It will need to be resubmitted to 26.10 since the deadline for 26.09 is passed.

Please be more specific. I added testing notes and developer information!

mdnoble73:
> > This still needs Jira updated properly. It will need to be resubmitted to 26.10 since the deadline for 26.09 is passed.
> 
> Please be more specific. I added testing notes and developer information!

It is not in the correct status, does not have a fix version. This is too late for 26.09, will need to be resubmitted to 26.10.

kylemhall:
> It is not in the correct status, does not have a fix version. This is too late for 26.09, will need to be resubmitted to 26.10.

Which status should it be in? I do not see a field for fix version.

It seems odd to block a _bug fix_ for a _major bug_ for such minor issues that aren't even related to the code itself. Are you not able to update those fields yourself?

mdnoble73:
@LiYanjun19 could you help @kylemhall get things into an appropriate status?

kylemhall:
@mdnoble73 can you point me to the relevant documentation so I know what fields should be set in the future?

mdnoble73:
It's all in the pull request documentation that we have been reviewing for the past several months. https://docs.google.com/document/d/1bw1utU6V9GGdJNRrvuq3E2uMB_YWY6XFhLBI2t59ZIw/edit?tab=t.0#heading=h.xpsmbkhoeqle

PR #4769, opened 2026-08-24
DIS-2826: Searching for call numbers with a colon

K-Alette:
- Update Solr.php to not strip colons when the regex matches typical call number patterns 
- Update AbstractGroupedWorkSearcher.php to not strip colons when unnecessary 
- Update release notes


mdnoble73 (changes requested):
@K-Alette do we need to escape the colons when we preserve the character since it's a reserved character for Solr.

K-Alette:
> @K-Alette do we need to escape the colons when we preserve the character since it's a reserved character for Solr.

@mdnoble73 Yeah good catch I think we do. Added a line for that even if we match on an expected id pattern we'll escape solr-reserved characters

mdnoble73 (approved):

PR #4783, opened 2026-08-28
DIS-1202: Fix OverDrive Products to Reindex 

LiYanjun19:
There is a field in OverDrive settings called “Products to Reindex”, where you should be able to list the OverDrive IDs and force specific updates for those records on the next index, but it’s not currently working. When entering OverDrive GUIDs into the "Products To Reindex" text field, the corresponding OverDrive records did not appear to reindex (no status change, no change in update/availability info/time in staff view) although the indexing pass did take/clear the entered GUIDs. 

To test:
1. Apply the PR.
2. Go to OverDrive Settings > Products to Reindex and enter the OverDrive record IDs.
3. Wait for the next indexing run.
4. Check the log for: `Added record <productId> to process queue`
5. Go to the record pages and verify that Staff View > Last Update has changed.

### DIFF
--- a/code/overdrive_extract/src/com/turning_leaf_technologies/overdrive/ExtractOverDriveInfo.java
+++ b/code/overdrive_extract/src/com/turning_leaf_technologies/overdrive/ExtractOverDriveInfo.java
@@ -183,14 +183,13 @@ class ExtractOverDriveInfo implements AutoCloseable {
 					logEntry.addNote("There are " + numProductsToUpdate + " products that need to be checked for updates");
 					logEntry.saveResults();
 
+					addProductsToUpdate();
+
 					//Do some counts of the records that will be updated for logging purposes
 					int numRecordsToUpdate = 0;
 					int numNewRecords = 0;
 					int totalRecordsWithChanges = 0;
 					for (OverDriveRecordInfo curRecord : allProductsInOverDrive.values()) {
-						if (settings.getProductsToUpdate().contains(curRecord.getId().toLowerCase())){
-							curRecord.hasChanges = true;
-						}
 						//Extract data from Libby and update the database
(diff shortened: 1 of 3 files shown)


code/overdrive_extract/src/com/turning_leaf_technologies/overdrive/ExtractOverDriveInfo.java
```diff
@@ -1135,6 +1134,24 @@ private synchronized void setLastSeenForProduct(long startTime, OverDriveRecordI
...
+		for (String productId : settings.getProductsToUpdate()) {
+			OverDriveRecordInfo recordInfo = allProductsInOverDrive.get(productId);
+
+			if (recordInfo == null) {
+				recordInfo = new OverDriveRecordInfo();
+				recordInfo.setId(productId);
+				getExistingRecordInformationForProduct(recordInfo);
+				allProductsInOverDrive.put(recordInfo.getId(), recordInfo);
+			}
+
```
JonahCWilson, line 1148: Can we batch these?  If `getProductsToUpdate()` is 1000 records, this is 1000 DB calls.  We could potentially get them all at once, then add them all to the `allProductsInOverdrive` in a for-loop after the fact.

JonahCWilson (changes requested):

mdnoble73 (changes requested):
Rather than continuing to use the productsToReindex field within settings which will be subject to having updates during indexing getting missed, we should track if a record needs to be reindexed in the OverDrive Products table and/or another table similar to the grouped_works_scheduled_index to schedule the reindex and avoid race conditions. That method would need to handle records that are available in multiple settings.

LiYanjun19:
will make changes and submit to 26.10. Closing this PR.

PR #4784, opened 2026-08-29
DIS-2784: Add Hoopla Records to Include

LiYanjun19:
For Hoopla version 2, add a Records To Reindex setting that allows admins to enter Hoopla record IDs, one per line, to refresh metadata, entitlements, and Flex availability before reindexing.

To test:
1. Apply the PR.
2. Go to Hoopla Settings > Record to Reindex and enter the Hoopla record IDs.
3. Wait for the next indexing run.
4. Check the log for: Doing extract of single work <recordID>
5. Go to the record pages and verify that Staff View > Last Update has changed.
6. Repeat step 2 but enter invalid IDs, for letters or special characters 
7. Wait for the next indexing run, check the logs for: Skipped invalid Hoopla record ID <invalid ID>

### DIFF
--- a/code/hoopla_export/src/com/turning_leaf_technologies/hoopla/HooplaExportMain.java
+++ b/code/hoopla_export/src/com/turning_leaf_technologies/hoopla/HooplaExportMain.java
@@ -181,7 +181,7 @@ public class HooplaExportMain {
 					if (singleWorkId == null) {
 						updatesRun = exporter2.exportHooplaData();
 					} else {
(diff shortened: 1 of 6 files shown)


mdnoble73 (changes requested):
Similar to my comments for OverDrive this can have race conditions as the indexer and user attempt to update the field so we can either store in the hoopla_export field and/or have another table that indicates titles that need to be reloaded.

LiYanjun19:
will make changes and submit to 26.10. Closing this PR.

PR #4793, opened 2026-08-31
DIS-2792: Hoopla Indexer Optimization

LiYanjun19:
Optimize Hoopla v1 and v2 export/indexing performance.

Current known work:
- Avoid loading all existing Hoopla titles during exporter construction; only scan existing title metadata when the exporter is actually extracting titles.  
- Add a stored rawResponseLength value to avoid recalculating UNCOMPRESSED_LENGTH(rawResponse) across millions of rows.

To test:
1. Apply the PR.
2. Run the Hoopla indexer as usual and verify that system resource usage is lower.


mdnoble73 (changes requested):
@LiYanjun19 can you resolve conflicts on this?

mdnoble73 (approved):

PR #4795, opened 2026-09-01
DIS-2878: Add support for Force Reindex to retrieve data from eContents

LiYanjun19:
Currently, Staff View > Force Reindex retrieves ILS records from their source system, but eContent records are generally reindexed using data already stored in Aspen. It does not make a vendor API request to retrieve the latest eContent metadata or availability first.

Add source update support for Force Reindex for OverDrive, Hoopla, CloudLibrary

Palace Project currently does not support single work extraction. 
Boundless (Axis 360) is not included in this work.

To Test:
1. Apply the PR.
2. Force reindex grouped works containing Hoopla, OverDrive, and CloudLibrary records.
3. Wait for the eContent records to refresh and reindex.
4. Review the Hoopla, OverDrive, and CloudLibrary logs to confirm that each Force Reindex request was processed.


LiYanjun19:
This PR depends on DIS-1202 and DIS-2784. I will rebase it after both are merged.

PR #4797, opened 2026-09-01
DIS-2875: Allow deleting Reading History entries containing apostrophes

poojary9991:
When trying to delete an entry from Reading history, clicking the “Delete” button does nothing if the title or author name has an apostrophe (e.g: The Winter’s tale). The confirmation popup does not show up and the page just jumps to the top.

Steps to Reproduce:
1. Go to the Account -> Reading History
2. Find an entry with an apostrophe in the title name
3. Click “Delete”
4. Popup does not show and page just jumps to the top
5. Apply the PR
6. Click “Delete” again and confirm the popup appears and the entry can be deleted successfully

### DIFF
--- a/code/web/interface/themes/responsive/MyAccount/readingHistoryEntry.tpl
+++ b/code/web/interface/themes/responsive/MyAccount/readingHistoryEntry.tpl
@@ -131,7 +131,7 @@
 
 			<div class="col-xs-12 col-md-3">
 				<div class="btn-group btn-group-vertical btn-block">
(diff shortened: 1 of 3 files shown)


mdnoble73 (commented):
@poojary9991 have you checked if you are now able to delete entries that have double quotes (") now ?

mdnoble73:
Did some more testing and looks like there are few/no records with a title containing a " so this should be fine.

PR #4804, opened 2026-09-01
DIS-2849 - Display libraries and locations in object list

mdnoble73:
- shows for Placards, System Messages, Basic Pages, Custom Pages, Web Resources, Quick Polls, Grapes Pages, Custom Web Resource Pages
- also Custom Form (only count)


NashvillePublicLibrary (approved):
simple, clean improvement for consortial libraries for these property displays.

PR #4809, opened 2026-09-02
DIS-2535 - re-enable Grouped Works V3

mdnoble73:


NashvillePublicLibrary (commented):
Turn v3 back on for testing! I'm looking forward to testing fun.

PR #4810, opened 2026-09-02
DIS-2815 Move Audience Indexing to Child Documents

mdnoble73:
- Index audience within child docs with grouped works v3
- When clearing a solr core (unusual) fully drop the core and rebuild to allow for changes in field definitions
- Optimize determining the correct audiences to show
- Sort scopes when determining scoping strings so they are in a known order


NashvillePublicLibrary (approved):
Moves audience to Record documents across all record indexers; Changes TreeSets to HashSets,

PR #4820, opened 2026-09-03
DIS-2608: Unit test portability

Jacobomara901:
@mdnoble73 This should not change functionality for your test system. This is just to allow the test suite to be run more universally on other systems. 
Stuff like:
 - Removing the hard coded windows paths
 - Making the distro/os detection dynamic
 - Making the bootstrap paths relative to the file, not where you ran it from
 - Allowing an optional env var to override the site name when the site you're running tests  on isn't "unit_tests.localhost"
 - Skipping the solr shutdown test if solr isn't running on localhost (will probably only be relevant for docker instances as solr runs in a different container)
 
To test:
 - Run the test suite as usual.
 
 The last 6 commits are me fixing the currently failing tests to ensure we start with a bare passing state.

### DIFF
--- a/code/web/services/EventRegistrationService.php
+++ b/code/web/services/EventRegistrationService.php
@@ -597,8 +597,7 @@ class EventRegistrationService {
 
 		require_once ROOT_DIR . '/CatalogFactory.php';
 		$catalog = CatalogFactory::getCatalogConnectionInstance(null, null);
-		$catalogSupportsNewUsers = $catalog != null && method_exists($catalog, 'findNewUser');
-		if ($catalogSupportsNewUsers) {
+		if ($catalog != null) {
 			$newUser = $catalog->findNewUser($barcode, '');
 			if ($newUser && !($newUser instanceof AspenError)) {
 				$newUser->getDisplayName();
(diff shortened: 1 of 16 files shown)


code/web/services/EventRegistrationService.php
```diff
@@ -597,7 +597,8 @@ public static function lookupUserByBarcode(string $barcode): array {
 
 		require_once ROOT_DIR . '/CatalogFactory.php';
 		$catalog = CatalogFactory::getCatalogConnectionInstance(null, null);
-		if (method_exists($catalog, 'findNewUser')) {
+		$catalogSupportsNewUsers = $catalog != null && method_exists($catalog, 'findNewUser');
```
gmcharlt: Nit: as PHPStan points out, the `method_exists()` check will always return true here. I think checking for a null `$catalog` is sufficient.

gmcharlt (changes requested):
Overall looks reasonable to me, but I've not yet put it through its paces. One nit found. 

Also noting that GitHub's display of the changes to `UserAccount.php` is less than helpful; the actual change is much less involved than it appears.

code/web/services/EventRegistrationService.php
Jacobomara901: Excellent point, I should have just removed the redundant check when I was fixing this. Will amend now

PR #4824, opened 2026-09-04
DIS-2893: Fix updateDatabase.php Smarty init order

lucasmontoya13:
Some database migration scripts (e.g. 26.09.00.php, via regenerateThemeCssForFontSize()) render theme CSS as part of the migration itself, which reads global $interface. updateDatabase.php only initialized $interface right before its own later, separate CSS-update call, after runPendingDatabaseUpdates() had already run. On a fresh database, this left $interface null while migrations executed, throwing a fatal error and hanging the entire backend startup. Fix moves the $interface initialization to before the migrations run.

Test plan

- Before: starting the backend against a fresh, empty database never completes; backend logs show PHP Fatal error: Uncaught Error: Call to a member function assign() on null in Theme.php:3171, and the container never reaches a ready state.
- After: same fresh-database startup completes; backend logs show migrations running to completion (or failing only on unrelated, pre-existing issues, not this crash), CSS updates succeed, and the backend reaches a ready state normally.

### DIFF
--- a/code/web/release_notes/26.09.00.MD
+++ b/code/web/release_notes/26.09.00.MD
@@ -132,6 +132,7 @@
 
 #### Docker
 - Remove a redundant background process check from the backend container's startup script. ([DIS-2867](https://aspen-discovery.atlassian.net/browse/DIS-2867)) (*LM*)
+- Fix a fatal error during database migrations on a fresh install, caused by Smarty theme variables being uninitialized before an update script that renders theme CSS. ([DIS-2893](https://aspen-discovery.atlassian.net/browse/DIS-2893)) (*LM*)
 
 #### EBSCO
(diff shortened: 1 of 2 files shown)


gmcharlt (changes requested):
Looks straightforward, but would "Docker" be better as the component? Also, it strikes me that this is a candidate for inclusion as a bugfix for 26.09, as the problem would be a blocker for new Docker setups.

lucasmontoya13:
Hi Galen! Yes, i should ask Mark (or a Jira Administrator) for adding Docker as a new component.
Also, agreed about adding it as a 26.09.00 hotfix

lucasmontoya13:
I can't change the branch I'm pushing to, so i'll close this PR and open a new one!

Thank you,Galen!

PR #4835, opened 2026-09-14
DIS-2871 Fix All Titles Search

mdnoble73:
Fix the "All Titles" Search being marked as an invalid search type

### DIFF
--- a/code/web/sys/SearchObject/AbstractGroupedWorkSearcher.php
+++ b/code/web/sys/SearchObject/AbstractGroupedWorkSearcher.php
@@ -1401,29 +1401,22 @@ abstract class SearchObject_AbstractGroupedWorkSearcher extends SearchObject_Sol
 	 * @return array
 	 */
 	public function getAllValidSearchIndexes() : array {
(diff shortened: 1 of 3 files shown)


code/web/sys/SearchObject/AbstractGroupedWorkSearcher.php
```diff
@@ -1401,29 +1401,22 @@ public function getAdvancedSearchIndexes() : array {
...
-				'Subject' => 'Subject',
-				'LocalCallNumber' => 'Call Number',
-				'ISN' => 'ISN',
-				'publisher' => 'publisher',
-				'year' => 'Year of Publication',
-				'toc' => 'Table of Contents',
-				'id' => 'Record Number',
-			];
-		}
+		return [
```
JonahCWilson, line 1404: If you return this, the return statement immediately below is unnecessary.

JonahCWilson (changes requested):

code/web/sys/SearchObject/AbstractGroupedWorkSearcher.php
mdnoble73: Yep, removed

PR #4838, opened 2026-09-15
DIS-2820: QA Fixes

catsoup11789:
Add options in Branded App Settings to restore legacy theme configuration and allowing admins to define a single theme for all locations


code/web/interface/themes/responsive/js/aspen/admin.js
```diff
@@ -3510,5 +3510,13 @@ AspenDiscovery.Admin = (function () {
...
 			});
 		},
+		toggleBrandedAppThemeOptions: function () {
+			const useIndividualThemes = $("#useIndividualThemes").prop("checked");
+			if (useIndividualThemes) {
+				$('#propertyRowoverallTheme').show();
+			} else {
+				$('#propertyRowoverallTheme').hide();
+			}
+		},
```
JonahCWilson, line 3520: Just a minor suggestion for reduced code complexity here.  Since we're doing one or the other no matter what, you can make one the default and just change it in the if condition.

```
const useIndividualThemes = $("#useIndividualThemes").prop("checked");
$('#propertyRowoverallTheme').hide();
if(useIndividualThemes){
    $('#propertyRowoverallTheme').show();
}
```

PR #4839, opened 2026-09-16
DIS-2926 - Internationalisation: 'Noon' and 'Midnight' must be translatable

Chloe070196:


code/web/sys/Utils/DateUtils.php
```diff
@@ -108,6 +108,29 @@ static function formatDateLocale($string, $dateStyle = 'medium', $timeStyle = 'n
...
+				'text' => 'Noon',
+				'isPublicFacing' => true,
+			]);
+		}
+		if ($isOnTheHour && $parsedTime['hour'] % 24 === 0) {
+			return translate([
+				'text' => 'Midnight',
+				'isPublicFacing' => true,
+			]);
+		}
```
JonahCWilson, line 129: A couple things here.  Apparently, `date_parse` will only return 0-23 for the hour, so the second half of the second if condition would always be true.  Since you're looking for midnight, you can check for `$parsedTime['hour'] === 0`.

A slightly shorter version, if you want it:

```
$hour = $parsedTime['hour'];
$isNoonOrMidnight = $hour === 0 || $hour === 12;

if($isOnTheHour && $isNoonOrMidnight){
    return translate([
        'text' => ($hour === 12) ? 'Noon' : 'Midnight',
        'isPublicFacing' => true
    ]);
}
```

JonahCWilson: Hmm, apparently I was wrong, 24:00 does give an hour 24, so it's not wrong to check for it, I just don't know if such a time string would ever happen.  But, always good to be defensive, so the `% 24 === 0` is better!

code/web/sys/LibraryLocation/Location.php
```diff
@@ -2340,6 +2340,7 @@ public function saveCloudLibraryScopes(): void {
 	}
 
 	public static function getLibraryHours($locationId, $timeToCheck): ?array {
+		require_once ROOT_DIR . '/sys/Utils/DateUtils.php';
```
JonahCWilson: Please doublecheck what I'm about to say, but with the new PSR-4 Autoloading, if you regenerate your files, `composer dump-autoload`, then you shouldn't need the `require_once` in any of the files in this PR.

code/web/sys/Utils/DateUtils.php
Chloe070196: Agreed. Went through the same sort of reflection when I extracted that out. Was not quite certain about the 24-specific check in the original - `%24` seemed safer, just in case we ever do get a 24 instead of 0 in there.

code/web/sys/LibraryLocation/Location.php
Chloe070196: I'll be honest - I didn't think that far on that one (adding those in systematically has become a bit of a reflex by now)

will give that a test and see if it can be removed safely

Chloe070196: As far as I could tell and test, it seemed happy without (no grumpy logs).

Re-doublechecking, here's where that went:

post [552732a](https://github.com/Aspen-Discovery/aspen-discovery/pull/4839/commits/552732a3693c1e1d92189f5e732a7ce46ae544b0), I temporarily dropped `"code/web/"` from `classmap` in `composer.json`, booted up a new dev box - got a lovely `{"success":false,"message":"Failed opening required 'DateUtils.php' (include_path='.:\/usr\/share\/php')"}` as expected instead of getting a library hours message. Undid that, rebooted, no error and a library hours message. Did the same to PSR-4 - no error either.

if I've wrapped my head around this right, yes, those `require_once` are redundant and  the classmap is doing its job.

-> adding [552732a](https://github.com/Aspen-Discovery/aspen-discovery/pull/4839/commits/552732a3693c1e1d92189f5e732a7ce46ae544b0) to remove `require_once` this would've otherwise introduced (separate commit just in case it needs dropping)

